PR #3, opened 2015-10-14
basic eslint support

dlockhart:


.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "brightspace",
+  "env": {
+    "node": true
```
dbatiste: If this becomes annoying, we could add it to the base config, or create a node-config that includes it.  Of course, if we end up with 15 flavors of predefined configs then it begins of defeat the purpose.  The [react config](https://github.com/Brightspace/eslint-config-brightspace/blob/master/react-config.js) sets things up for that case, but arguably is a little opinionated in including the envs that it does.  My hope/intent when creating that one was that it would cover 90% of react cases.

dlockhart: I'm cool with having to add this for node.js development.

PR #4, opened 2015-10-14
Switching from Gulp to Browserify and exposing the "d2l-intl" library

dlockhart:
@dbatiste: what do you think?


dbatiste:
Looks good.  It looks like lie is on bower too, but I don't think it makes any difference in this case which package manager is used.

PR #6, opened 2015-10-28
Initial version updates to pick up recent C12 typography.

dbatiste:
@dlockhart : these are initial version updates to pick up recent typography changes for C12.  Besides vui-typography, it also includes vui-input and jquery-vui-collapsible-section version updates since those components needed a couple small tweaks as a result of removing variables and changing the typography mixins.


dlockhart:
LGTM

PR #13, opened 2015-12-09
outputting classes for text-image-link control

dlockhart:
@dbatiste: thoughts? This control currently has a copy of our link styles (colour mostly) -- it should really reference them from here. If this works, I'll port this over to the c12 branch and adapt it accordingly. I'll also do something similar for html-block.


dlockhart:
Here's the CSS in the monolith that this is replacing:
http://search.dev.d2l/source/xref/LSOne/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/TextImageLink/TextImageLink.css

dbatiste:
Ya I agree.  I don't think focus-outline has been updated yet.

PR #27, opened 2016-01-04
Updated build pipeline to use clean-css and uglify-js over uglify

rylan:
The npm project Uglify on windows throws errors when trying to minify css. Updated project to use the vastly more popular clean-css and uglify-js projects for minifying js and css code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "brightspace-integration",
-  "version": "10.5.5",
+  "version": "10.5.6",
   "description": "Integration project for components consumed by Brightspace.",
   "private": true,
   "main": "src/index.js",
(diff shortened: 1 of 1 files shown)


dbatiste:
Looks good.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "brightspace-integration",
-  "version": "10.5.5",
+  "version": "10.5.6",
```
dbatiste: I think we want this version aligned with the LMS version it's going with.

dlockhart: The version isn't as important anymore... or not at all even since this isn't consumed in a semver way by anything. The important thing is the release tag.

dbatiste: Yeah - I sorta wish this wasn't aligned with LMS at all.  If we leave it be, hopefully it won't cause any confusion for anyone else looking at this.

rylan: It would be nice if the value was a semver though I don't see this kind of project having many breaking changes

PR #36, opened 2016-01-13
Add overriding styles for dialogs.

njostonehouse:
These styles are currently hard-coded into the LMS (referencing #353535). The plan is to later remove those styles from the LMS, which would fall back to these (identical) styles. This will then give us a good spot to update the styles for C12.

### DIFF
--- a/src/bsi.scss
+++ b/src/bsi.scss
@@ -8,3 +8,4 @@
 @import 'floating-container.scss';
 @import 'html-block.scss';
 @import 'text-image-link.scss';
+@import 'dialog.scss';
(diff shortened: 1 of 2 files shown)


dlockhart:
All the class names you're applying these styles to are in the legacy framework, so I'm not sure how I feel about now including this CSS on every page.

dbatiste:
Spit-balling here...

Eventually I think we'll want vui-dialog, so we could potentially create that now, and add those class names to legacy dialogs as needed for visual-redesign.  To isolate the current legacy styles that would cause trouble, they could be moved into a `VisualRedesignOff` file for legacy, rather than into a BSI file.   ?

njostonehouse:
I think vui-dialog is worth creating at some point, but we'd might as well wait until we have the design ready. It sounds to me like we probably want to block the Dialog defects for now, until we have a direction on dialogs.

PR #43, opened 2016-01-18
C12i12n grid

rylan:
Update to the MVC Grid, Legacy Grid, and Legacy Data-Grid to use vui-table styles.


package.json
```diff
@@ -18,8 +18,9 @@
     "preserve": "npm run build",
     "serve": "http-server dist",
     "test": "npm run build",
-    "uglify:css": "uglify -s ./dist/bsi.css -o ./dist/bsi.min.css -c",
-    "uglify:js": "uglify -s ./dist/bsi.js -o ./dist/bsi.min.js"
+    "uglify:css": "cleancss -s ./dist/bsi.css -o ./dist/bsi.min.css -c",
```
dlockhart: In the future, I'd recommend making these kinds of changes in an isolated pull request.

src/grid.scss
```diff
@@ -0,0 +1,290 @@
+@import '../bower_components/vui-table/table.scss';
+@import '../bower_components/vui-colors/colors.scss';
+@import '../bower_components/vui-input/input-checkbox.scss';
```
dlockhart: Why the inconsistent import paths? `../bower_components` vs `bower_components/`.

src/grid.scss
```diff
@@ -0,0 +1,290 @@
...
+		border-bottom-right-radius: px-to-base-rem(6);
+		[dir='rtl'] & {
+			border-bottom-right-radius: 0;
+			border-bottom-left-radius: px-to-base-rem(6);
+		}
+	}
+}
+
+@mixin _grid-row-selected($important:""){
+	background-color: $vui-color-celestine-light-1 unquote($important);
```
dlockhart: This will result in a trailing space before the semi-colon in the case where `$important` is excluded. Could you do something like this?

    @mixin blah($important: false) {
      background-color: if($important, $vui-color-celestine-light-1 !important, $vui-color-celestine-light-1);
    }

src/grid.scss
```diff
@@ -0,0 +1,290 @@
...
+	outline: 1px solid $vui-color-celestine;
+}
+
+/* MVC grid is d2l-grid
+ * Legacy data grid is .yui-dt */
+
+.d2l-grid,
+.yui-dt > table {
```
dlockhart: How does this approach compare to applying the `vui-table` (and other `vui-table-*`) class name to the elements in the LMS instead?

src/grid.scss
```diff
@@ -0,0 +1,290 @@
...
+.d2l-grid-loadmore-wrapper {
+	width: 20em;
+	margin: auto;
+	padding: px-to-base-rem(5);
+}
+
+.d2l-grid-cell-sel {
+	width: px-to-base-rem(13);
+	& input[type="checkbox"] {
+		@include vui-input-checkbox();
```
dlockhart: This seems like a case where `vui-input` could have just been applied to the checkbox's rendering code... `vui-input-checkbox()` generates a fair bit of CSS that's now going to be duplicated inside here.

rylan: I will look into appending the classes in the lms to remove the duplication.

src/grid.scss
rylan: typo, fixed

src/grid.scss
rylan: updated the different grids to render vui-input class on the checkboxes

src/grid.scss
rylan: updated.

package.json
rylan: of course

PR #75, opened 2016-02-18
Death to d2l-min-width

dlockhart:
Removing the min-width we apply to the page to allow for responsiveness.

@d2l-joelp: here you go!


dlockhart:
wrong branch

PR #76, opened 2016-02-18
removing min-width

dlockhart:
Removing the min-width we apply to the page to allow for responsiveness.

@d2l-joelp: here you go!


d2l-joelp:
Most excellent, thanks!

PR #80, opened 2016-02-24
Dialog

rylan:
Support for MVC dialogs to have the visual redesign look and feel 

Related PR https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/2236/overview

### DIFF
--- a/src/dialog.scss
+++ b/src/dialog.scss
@@ -52,7 +52,7 @@ $dialog-background-color: #fff;
 	border: 1px solid $vui-color-titanius;
 	border-radius: px-to-base-rem(8);
 	box-shadow: 0 2px 12px rgba(86,90,92,0.25);
-	padding: px-to-base-rem(10) px-to-base-rem(30);
+	padding: px-to-base-rem(30);
 
 	@media #{$dialog-media-break-small} {
 		border-radius: 0 !important;
--- a/src/dialog.scss
+++ b/src/dialog.scss
@@ -78,23 +78,19 @@ $dialog-close-icon: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABAAAAAQA
 .d2l-dialog-inner {
 	background-color: $dialog-background-color;
 	border: 1px solid $vui-color-titanius;
-	border-radius: px-to-base-rem(8);
+	border-radius: 0.4rem;
 	box-shadow: 0 2px 12px rgba(86,90,92,0.25);
-	padding: px-to-base-rem(30);
+	padding: 1.5rem;
 
 	@media #{$dialog-media-break-small} {
 		border-radius: 0 !important;
 		width: 100% !important;
 		height: 100% !important;
-
-		& > iframe {
-			width: 100% !important;
-		}
 	}
 }
 
 .d2l-dialog-footer .vui-button {
-	margin-top: px-to-base-rem(10);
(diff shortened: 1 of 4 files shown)


src/dialog.scss
```diff
@@ -0,0 +1,138 @@
...
+
+@mixin _dialog-animation_slide-from-top($state:"open"){
+
+	animation: {
+		duration: 0.5s;
+		timing-function: linear;
+		name: if($state=="open", dialog-slide-in-from-top-open, dialog-slide-in-from-top-close);
+
+	}
+	@if $state=="close" {
```
rylan: only need top value on close to prevent flickering between animation complete and node removal. On open it will animate to whatever top value the framework has set for the dialog

src/dialog.scss
```diff
@@ -0,0 +1,138 @@
+@import '../bower_components/vui-colors/colors.scss';
+@import '../bower_components/vui-typography/px-to-base-rem.scss';
+@import '../bower_components/vui-loading-spinner/loadingSpinner.scss';
+
+$dialog-media-break-small: "(max-device-width: 767px), screen and (max-width:767px)";
```
rylan: breakpoints might change

src/dialog.scss
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+.d2l-dialog-footer .vui-button {
+	margin-top: px-to-base-rem(10);
+}
+
+.d2l-dialog-footer .d2l-dialog-resize {
+	@media #{$dialog-media-break-small-iframe}{
+		display: none !important;
```
rylan: @media screen and (max-width:767px) is based on iframe width and not viewport width, have a support js change in the framework to remove the handle if the window shrinks enough to render alternate view of dialog for small devices.

src/dialog.scss
```diff
@@ -0,0 +1,137 @@
...
+		left: auto !important;
+		padding: 0 !important;
+		top: 0 !important;
+	}
+}
+
+.d2l-dialog-inner {
+	background-color: $dialog-background-color;
+	border: 1px solid $vui-color-titanius;
+	border-radius: px-to-base-rem(8);
```
dbatiste: If we change the base font-size (in vui-typography), do we still want this to be 8px?  Presumably if we scale the font down slightly, we'd scale other things down along with it, possibly including the border-radius.  I think in (most) other places we've chosen to just hard-code the `rem` value to that if font is scaled down, other things will be too.

src/dialog.scss
```diff
@@ -0,0 +1,137 @@
...
+	box-shadow: 0 2px 12px rgba(86,90,92,0.25);
+	padding: px-to-base-rem(30);
```
dbatiste: Same thing with this one (re calculating the `rem` value vs hard-coding it).

src/dialog.scss
```diff
@@ -0,0 +1,137 @@
...
+	min-height: calc(2rem + 2px);
+}
+
+.d2l-dialog-close {
+	border-top-right-radius: px-to-base-rem(6);
+	height: calc(2rem + 2px);
+	width: calc(2rem + 2px);
+
+	[dir='rtl'] & {
+		border-top-left-radius: px-to-base-rem(6);
```
dbatiste: Does `border-top-right-radius` need to be overridden?

PR #85, opened 2016-03-22
US68431: Add Daylight-specific shim styles.

njostonehouse:
@rylan @dlockhart @dbatiste @capajon adding shim styles to BSI.


dlockhart:
LGTM

PR #87, opened 2016-03-23
Add bsi.html (with placeholder).

dbatiste:
This PR adds `bsi.html` and `vulcanize` build step.  Eventually the `unresolved` JS will be replaced once we are including components.

@dlockhart : be awesome if you could take look.

### DIFF
--- a/src/bsi.html
+++ b/src/bsi.html
@@ -1,0 +1,3 @@
+<script>
+  document.body.removeAttribute('unresolved');
+</script>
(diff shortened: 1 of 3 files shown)


src/bsi.html
```diff
@@ -0,0 +1,3 @@
+<script>
+  document.body.removeAttribute('unresolved');
```
dbatiste: This is necessary until we include a component that uses Polymer (which will remove this for us).

dlockhart:
LGTM!

PR #89, opened 2016-03-24
pulling in new navigation prototype WC

dlockhart:
@dbatiste: seem reasonable?


src/bsi.html
```diff
@@ -1,3 +1,4 @@
+<link rel="import" href="../bower_components/brightspace-navigation/navigation.html">
```
dbatiste: You might get a merge conflict on this (since I recently updated).

PR #90, opened 2016-03-28
Use .scss from respective node_modules so that WC can pull from bower.

dbatiste:
This PR changes to pull a few VUI components from NPM and point to node_modules for their respective `.scss` files so that we can gradually switch over to WCs that will be pulled from Bower.  It also sets up `bsi-style` component where shared styles can be included.

@njostonehouse , @dlockhart , @capajon : FYI... let me know if you have any concerns.  I am hoping to switch over to the WC version of typography today.


src/floating-container.scss
```diff
@@ -1,5 +1,6 @@
-@import 'bower_components/vui-colors/colors.scss';
-@import 'bower_components/vui-typography/px-to-base-rem.scss';
+@import 'node_modules/vui-colors/colors.scss';
+@import 'node_modules/vui-typography/headings.scss';
```
dlockhart: Did you mean to include `headings.scss`?

dbatiste: Ah yes - I think this should have been included before.  It was working because other things were importing that file.

njostonehouse:
lgtm

PR #97, opened 2016-04-04
Adding vui-link web component

capajon:
@dbatiste 

Is there anything else that needs to be updated?


dbatiste:
I think that's it.  As mentioned, eventually we'll get to the point where the .scss can be removed, and also the reference to the branch containing the Sass.

PR #101, opened 2016-04-08
Change background to white

capajon:
Daylight bug bash story:
https://rally1.rallydev.com/#/15545167705d/detail/userstory/49673235406

I went through several pages to see if I needed to adjust any other special cases.  I didn't find any but please let me know if you are aware of any.

@dlockhart 

I'll probably wait to do a release/update the LMS BSI config file, so I'm not doing one per story - probably do it daily rather than per 'bash story'.


dlockhart:
Cool LGTM!

PR #102, opened 2016-04-08
adding first iteration of navigation web component

dlockhart:
@dbatiste look OK?

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "build:sass": "node-sass --output-style expanded ./src/bsi.scss ./dist/bsi.css",
     "postbuild:sass": "npm run autoprefix && npm run uglify:css",
     "build:icons": "cpy ./bower_components/vui-icons/images/**/*.* ./dist/images",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "build:sass": "node-sass --output-style expanded ./src/bsi.scss ./dist/bsi.css",
     "postbuild:sass": "npm run autoprefix && npm run uglify:css",
     "build:icons": "cpy ./bower_components/vui-icons/images/**/*.* ./dist/images",
-    "build:wc": "vulcanize ./src/bsi.html > ./dist/bsi.html",
+    "build:wc": "vulcanize --inline-scripts ./src/bsi.html > ./dist/bsi.html",
```
dlockhart: This was needed because `iron-ajax` includes a polyfill for promises that results in a 404 if you don't inline the script (or include it with BSI).

dbatiste: Looks good.

PR #103, opened 2016-04-08
Add note about trailing slash to readme

ryantmer:
Itty bitty PR. Omitting the trailing slash causes things to not work, which is less than ideal.


dlockhart:
Oops, good catch, thanks!

PR #106, opened 2016-04-11
US69524: Adjust spacing around the side Two Panel Selector search bar.

njostonehouse:


src/page-two-panel-selector.scss
```diff
@@ -44,11 +44,9 @@
 }
 
 .d2l-twopanelselector-side .d2l-page-search {
-	margin-left: 20px;
-	[dir='rtl'] & {
+	margin-bottom: 20px;
 		margin-left: 0;
-		margin-right: 20px;
-	}
+	padding: 0 30px;
```
dlockhart: Is the intent `30px` or `1.5rem`? (the search inputs internal padding is rem-based)

njostonehouse: It should be constant. This is the padding around the search field, to make sure it's the same width as the side panel below it (also will have 30px padding).

njostonehouse:
This PR is associated with LE PR 4461.

src/page-two-panel-selector.scss
```diff
@@ -44,11 +44,9 @@
 }
 
 .d2l-twopanelselector-side .d2l-page-search {
-	margin-left: 20px;
-	[dir='rtl'] & {
-		margin-left: 0;
-		margin-right: 20px;
-	}
+	margin-bottom: 20px;
+	margin-left: 0;
```
dbatiste: Is this overriding some left margin applied elsewhere?  If so, is there a similar RTL override needed?

dlockhart: Agree with Dave's comment. If it's not overriding something, then the rule can just be excluded (things get 0 margin by default).

njostonehouse: Yep, I missed having to override PageSearch.RTL.css. Thanks guys.

PR #109, opened 2016-04-11
Use web-component-shards to bundle WCs and resolve common dependencies

d2l-joelp:
- Use web-component-shards project to bundle WCs, and stop using vulcanize directly
- Moved all web components to the root of the project
- Changed relative path references to bower components to not have bower_components in the path
- Add my-courses-widget as a bower dependency
- Add shard endpoints for bsi.html and my-courses-widget.html
- Add instructions for how to reference a new WC.


package.json
```diff
@@ -13,7 +13,7 @@
     "build:sass": "node-sass --output-style expanded ./src/bsi.scss ./dist/bsi.css",
     "postbuild:sass": "npm run autoprefix && npm run uglify:css",
     "build:icons": "cpy ./bower_components/vui-icons/images/**/*.* ./dist/images",
-    "build:wc": "vulcanize --inline-scripts ./src/bsi.html > ./dist/bsi.html",
+    "build:wc": "web-component-shards --endpoints bsi.html my-courses-widget.html",
```
dlockhart: Can you confirm that scripts still get inlined? If it's working, there should be a promise polyfill that gets inserted by `iron-ajax`.

d2l-joelp: Yes, it looks like the scripts are getting inlined. All the code from `promise-polyfill` appears to be in shared.html

dlockhart:
I'm good if you want to merge this!

dbatiste:
Presumably it was necessary to move the WC `.html` files out to the root.  For my own understanding, what's the background reason for this (was definitely nice to have the source files separate from other things)?

dlockhart:
It seems like Polymer itself (and polyserve) tend to like it when you reference everything as though it were a sibling of yours in `bower_components` -- `../other-thing/thing.html`. Joel also noticed that by referencing stuff from the root, vulcanize attached the correct asset-path to each component... otherwise it had weird `bower_components` artifacts in there.

PR #114, opened 2016-04-14
Dbatiste/context menu enter key legacy fix

dbatiste:


dlockhart:
So I'm a little worried that we're still relying on `vui-button` so much. We're actually still importing all the CSS for it, despite using the WC version of it for actual buttons. The longer we wait to pull that out, the risker it's going to be.

dbatiste:
Yeah I was thinking about this - I think eventually (hopefully soon) we can use Polymer mixins to produce the styles.  We've been using `vui-button` in the LMS for button-menu styling, and in this case we're just using it to target these styles.  In such case we still need a class that can be applied, unless of course we change MVC and legacy to use the WC.

dbatiste:
I've removed the references to `vui-button`, and also `d2l-button-menu` (as using it in the LMS results in the MVC class initializer running in legacy-land).  There is still some Sass here to pull in a couple of things - much of this will be updated once the WC button-menu/context-menu can provide mixins.

PR #115, opened 2016-04-19
Us69647 html editor

capajon:
Will be updated with more details.

Related PR:
LE: https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/2428/overview

### DIFF
--- a/bsi-style.html
+++ b/bsi-style.html
@@ -1,7 +1,7 @@
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../vui-typography/typography.html">
 <link rel="import" href="../vui-link/link.html">
-<link rel="import" href="html-editor.html">
+<link rel="import" href="htmleditor.html">
 
 <link rel="import" href="button.html">
 <link rel="import" href="button-menu.html">
--- a/bsi-style.html
+++ b/bsi-style.html
@@ -17,6 +17,6 @@
 		<style include="bsi-collapse-pane"></style>
 		<style include="bsi-headings"></style>
 		<style include="bsi-homepages"></style>
-		<style include="bsi-html-editor"></style>
+		<style include="bsi-htmleditor"></style>
 	</template>
 </dom-module>
--- a/bsi-style.html
+++ b/bsi-style.html
@@ -1,6 +1,7 @@
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../vui-typography/typography.html">
 <link rel="import" href="../vui-link/link.html">
+<link rel="import" href="html-editor.html">
 
 <link rel="import" href="button.html">
 <link rel="import" href="button-menu.html">
@@ -16,5 +17,6 @@
 		<style include="bsi-collapse-pane"></style>
 		<style include="bsi-headings"></style>
 		<style include="bsi-homepages"></style>
+		<style include="bsi-html-editor"></style>
 	</template>
 </dom-module>
--- a/html-editor.html
+++ b/html-editor.html
@@ -5,33 +5,33 @@
 	<template>
 		<style>
 			:root {
-				--vui-htmleditor-height: 40px;
+				--vui-htmleditor-toolbar-height: 40px;
 				--vui-htmleditor-border-radius: 0.3rem;
 			}
+
+			/* Related to HtmlEditor.css: */
 
 			.d2l-htmleditor {
 				border-radius: var(--vui-htmleditor-border-radius);
 			}
 
-			.d2l-htmleditor-button {
-		    padding: calc((var(--vui-htmleditor-height) - 16px) / 2);
+			.d2l-htmleditor-header {
+				padding: 0.5rem 0.5rem 0 0.5rem;
 			}
 
-			.d2l-htmleditor-button:hover,
-			.d2l-htmleditor-button:focus {
-		    background-color: var(--vui-color-woolonardo);
+			.d2l-htmleditor-footer {
+				padding: 3px 0.7rem;
 			}
 
-			.d2l-htmleditor-color-button {
-		    padding: calc((var(--vui-htmleditor-height) - 16px - 4px) / 2);
+			.d2l-htmleditor-toolbar-item .d2l-select {
+				font-size: 0.7rem;
+				font-weight: normal;
 			}
 
-			.d2l-htmleditor-buttonmenuitem {
-				padding: 0.5rem;
-			}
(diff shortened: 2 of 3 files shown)


html-editor.html
```diff
@@ -0,0 +1,207 @@
...
+			.d2l-htmleditor-component-toggle:focus {
+				background-color: var(--vui-color-woolonardo);
+			}
+
+			.d2l-htmleditor-component-container {
+				line-height: normal;
+			}
+
+			.d2l-htmleditor-toolbar-item {
+		    min-height: var(--vui-htmleditor-toolbar-height);
```
dlockhart: Nit: missing tab

html-editor.html
```diff
@@ -0,0 +1,207 @@
...
+
+			/* Related to Group.css */
+
+			.d2l-htmleditor-group {
+				border-radius: var(--vui-htmleditor-border-radius);
+				margin-right: 0.3rem;
+			}
+
+			[dir="rtl"] .d2l-htmleditor-group {
+				margin-right: 0px;
```
dlockhart: For zero values, you can leave the unit off, e.g. `margin-right: 0;`

bsi-style.html
```diff
@@ -1,6 +1,7 @@
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../vui-typography/typography.html">
 <link rel="import" href="../vui-link/link.html">
+<link rel="import" href="html-editor.html">
```
dbatiste: Minor formatting thing here: I think the format `html-editor` vs `htmleditor` should match.  Below there are variables specified that use `--vui-htmleditor-*`.  I don't have strong feeling one way or another, other than to try and be consistent.

html-editor.html
```diff
@@ -0,0 +1,207 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../vui-colors/colors.html">
+
+<dom-module id="bsi-html-editor">
+	<template>
+		<style>
+			:root {
+				--vui-htmleditor-toolbar-height: 40px;
```
dbatiste: Is the toolbar height fixed at 40px?  What happens when the user selects a different font-size in Account Settings?

html-editor.html
```diff
@@ -0,0 +1,207 @@
...
+				border-radius: var(--vui-htmleditor-border-radius);
+				margin-right: 0.3rem;
+			}
+
+			[dir="rtl"] .d2l-htmleditor-group {
+				margin-right: 0px;
+				margin-left: 0.3rem;
+			}
+
+			.d2l-htmleditor-group .d2l-htmleditor-toolbar-item .d2l-select {
```
dbatiste: I suspect this was existing previously, but I am curious if `.d2l-htmleditor-group` is necessary on these selectors?

html-editor.html
```diff
@@ -0,0 +1,207 @@
...
+				border-bottom-right-radius: 0;
+				border-top-left-radius: var(--vui-htmleditor-border-radius);
+				border-bottom-left-radius: var(--vui-htmleditor-border-radius);
+			}
+
+			/* Related to Menu.css & MenuFlyout.css: */
+
+			.d2l-menuflyout.d2l-htmleditor-menuflyout .d2l-menuflyout-opener {
+				padding: calc((var(--vui-htmleditor-toolbar-height) - 16px) / 2) 0.35em;
+				border-left: 1px solid #CCC;
```
dbatiste: I think this color should be updated - not sure if you have that coming in a different PR.  I think the colors should be the same as what [vui-dropdown](https://github.com/Brightspace/valence-ui-dropdown/blob/master/menu.scss) provides.  If you need to add animation for opening menu, let's chat (I don't do what the same).

html-editor.html
```diff
@@ -0,0 +1,207 @@
...
+
+			[dir="rtl"] .d2l-htmleditor-color .d2l-htmleditor-color-select {
+				border-top-right-radius: 0;
+				border-bottom-right-radius: 0;
+				border-top-left-radius: var(--vui-htmleditor-border-radius);
+				border-bottom-left-radius: var(--vui-htmleditor-border-radius);
+			}
+
+			.d2l-htmleditor-color-button {
+				padding: calc((var(--vui-htmleditor-toolbar-height) - 16px - 2px) / 2);
```
dbatiste: This results in a "fixed" height since the variable is defined as `40px` - how does this look when the user changes their font-size in Account Settings?

bsi-style.html
capajon: I'll go with htmleditor since that matches the d2l class names established in the LE.

html-editor.html
capajon: Will investigate.  I did try changing my browser base font but now I get what the LE is doing! :)

html-editor.html
capajon: I'm going to add the border color to the list of outstanding items to address because I think it can be easily chunked out on its own and would impact the LE PR as well.  Also, there wasn't a push for the drop downs to match other menus exactly yet, but I can put that on the suggested things to work on next.

capajon:
Sorry I'm use to my previous team conventions and didn't expect so much review already. :)  But it is awesome, thanks.

I wanted to mention up front that there are a few straggling items I want to follow up on fixing - especially since I need Jeff's review and will most likely need to make some changes anyway.  I noted them in Rally:  https://rally1.rallydev.com/#/15545167705d/detail/userstory/53957995002

All the items will be addressed unless I commented otherwise.

capajon:
Should I pull the trigger on this and https://github.com/Brightspace/brightspace-integration/pull/116 ?  Obviously a related LE PR too but these are the first dominos.  I'm not aware of any outstanding items but if more review is needed that is obviously very valid and I'll wait. :)

bsi-style.html
```diff
@@ -16,5 +17,6 @@
 		<style include="bsi-collapse-pane"></style>
 		<style include="bsi-headings"></style>
 		<style include="bsi-homepages"></style>
+		<style include="bsi-html-editor"></style>
```
dbatiste: Rename to `bsi-htmleditor` for consistency.

dbatiste:
I'd be onboard with updating (since it improves things greatly), however as mentioned earlier we need to be careful of risk given the TH demo next week.

bsi-style.html
capajon: Ughhh sorry missed the module name :(  Updating.

PR #116, opened 2016-04-19
Us69647 html editor daylight off

capajon:
Will be updated with more details.

Related PR: https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/2428/overview

### DIFF
--- a/src/bsi.scss
+++ b/src/bsi.scss
@@ -11,3 +11,4 @@
 @import 'page-two-column.scss';
 @import 'page-two-panel-selector.scss';
 @import 'text-image-link.scss';
+@import 'html-editor.scss';
--- a/src/html-editor.scss
+++ b/src/html-editor.scss
@@ -1,0 +1,62 @@
+.d2l-htmleditor {
+	border-radius: 3px;
+}
+
+.d2l-htmleditor-header {
+	padding: 3px;
+}
+
+.d2l-htmleditor-toolbar-item {
+	min-height: 2.2em;
+	border-left: 1px solid #CCC;
+}
+
+.d2l-htmleditor-toolbar-item:first-child {
+	border-left: 1px solid transparent;
+}
+
+.d2l-htmleditor-toolbar-item .d2l-select {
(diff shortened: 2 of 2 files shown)


capajon:
This one is ready for review.  Does anyone see any issues? I tried to group by the original css files to make it a little easier to compare to the LE changes (I hope!).

dlockhart:
I'm guessing the end state of this involves CSS split across BSI (master and Daylight) and ILP? A slightly different approach would be to just move everything out of ILP into the two BSI branches. Not sure which one I prefer for components like this that don't have a VUI version.

The advantage to not spreading them across BSI/ILP is that the CSS is in a single place and people in the future don't accidentally ignore one of them when making modifications.

capajon:
I was wondering about that afterwards actually.  I think I prefer what you are describing, keeping it all in one location.  I don't like that you have to look across two repos to see the 'whole picture'.  Feels hard to discover.

But counter-points:

First, for the CSS that is not changing (and there is a lot of it): I don't love duplicating it in the daylight and non-daylight branches of BSI.  It might be okay though, since the non-daylight CSS is unlikely to change now and if it somehow does it isn't too crazy to update it twice.  

Second, we would want to structure the CSS into subfiles (roughly matching the LEs CSS files or something similar).  Or at least something to keep things organized.  Dumping all the CSS into one file seems less than ideal.  I'm not sure what the best approach would be for web components, to organize that much CSS?  I can think of ways, just not sure what would be the nicest in the end.  I do miss SCSS.  There should be a SASS pre processor for web components. :)  Seems doable...

Anyway, I'd prefer to address this (if we want to) when I go back to finish cleaning up the little items in the editor that are outstanding (after Jeff's review).  Or maybe even later on depending.

dlockhart:
All good points. Yeah easy to split stuff across WCs, but I'd say until we turn the editor into a WC (very long-term project), we will continue tol use SCSS in BSI -- it's a completely valid tech for what we're using it for.

PR #117, opened 2016-04-22
US69524: Allow vui-typography to be set solid by removing padding, ma…

njostonehouse:
…rgin and line-height leading.


njostonehouse:
@dbatiste Not sure what we'd like as the name of this class, or even if it should be integrated into typography proper. I think this is a good option to provide, especially for headings.

PR #118, opened 2016-04-22
picking up latest navigation component with page loading indicator

dlockhart:


dlockhart:
Not going to merge this just yet since it depends on a corresponding change to ILP-Core and I think I'm going to wait until mid-next week to do that.

PR #128, opened 2016-04-27
Adding missing icons for dialog from switch to BSI from LE (relative …

capajon:
…path will not work)

@dbatiste   Any concerns?  I guess that ultimately the user will not have to download the x icon a couple more times (sort of) in their use of the lms but I can live with that considering how small it is. :)


dbatiste:
Looks good.  Spoke with Sarah about resize in Daylight - she's still working on an icon for that... hopefully today or tomorrow, otherwise I will just put the old icon in until we get it.

PR #131, opened 2016-04-28
US68737: Include a bottom margin of 60px on page headers.

njostonehouse:


src/page.scss
```diff
@@ -48,3 +48,8 @@
 .d2l-page-main-padding {
 	padding: 10px 1.5rem;
 }
+
+.d2l-page-header {
+	margin-bottom: 60px;
```
dbatiste: Should this actually be 3rem?  Also wondering if this is just patching override on headings (i.e. fix on the fix)?

njostonehouse: This is increasing the margin beneath a header as part of Daylight's spacing-out. It shouldn't be related to font headings.

njostonehouse:
I'm going to break this PR down into smaller changes. Too much stuff going on here that I haven't fully qualified.

PR #144, opened 2016-05-03
Nstonehouse/us71024/transplant styles daylight cr

njostonehouse:


htmlblock.html
```diff
@@ -0,0 +1,129 @@
...
+				font-family: Monospace;
+				font-size: 13px;
+				margin: 13px 0;
+			}
+
+			.vui-typography .d2l-htmlblock p {
+				margin: 0.5em 0 1em 0;
+			}
+
+			[dir='rtl'] .vui-typography .d2l-htmlblock {
```
dbatiste: Is this working?  Note that `[dir='rtl']` and `vui-typography` are almost always on the same element.

htmlblock.html
```diff
@@ -0,0 +1,129 @@
...
+				line-height: normal;
+				text-align: left;
+				overflow-x: auto;
+				overflow-y: hidden;
+			}
+
+			.vui-typography .d2l-htmlblock ul,
+			.vui-typography .d2l-htmlblock ol {
+				list-style-position: outside;
+				margin: 1em 0;
```
dbatiste: Are more changes coming for this? Thinking about the use if `em` values here.

htmlblock.html
```diff
@@ -0,0 +1,129 @@
...
+			.vui-typography .d2l-htmlblock ul[type="disc"] {
+				list-style-type: disc;
+			}
+
+			.vui-typography .d2l-htmlblock ul ul,
+			.vui-typography .d2l-htmlblock ul ol,
+			.vui-typography .d2l-htmlblock ol ul,
+			.vui-typography .d2l-htmlblock ol ol {
+				margin-top: 0;
+				margin-bottom:0;
```
dbatiste: Nitpick: should be space after `:`.

dbatiste:
My understanding is that this is simply a transplant - more changes coming eventually to fix the styles to follow typography, yes?

njostonehouse:
Yes, it's just a transplant, although I don't have any plans currently to adjust the styles, I agree it needs to be done.

htmlblock.html
njostonehouse: That's an excellent point. I hadn't checked. Will test.

htmlblock.html
njostonehouse: More changes should come, yes. Nothing in progress, but definitely a good point.

dbatiste: Might be worthwhile creating a story to track (if one doesn't already exist).

njostonehouse: Yup, I put @joshassad on it.

PR #155, opened 2016-05-10
Us70869 page padding on smaller viewports

capajon:
This is a result of work on the announcements page to make it look better on mobile.  We found the padding to be a bit excessive on smaller screens.

I'm using EMs here which I personally like for media queries but I could also change it to px if that's what is preferred.

Jeff and I discussed this and would like to further improve it in the not too distant future, to work with our overall grid design - but this is still helpful now.

@dlockhart

### DIFF
--- a/src/page.scss
+++ b/src/page.scss
@@ -50,7 +50,7 @@
 
 	@media screen and (max-width: 25em)  {
     padding-left: 0.75rem;
-		padding-right: 0.75rem;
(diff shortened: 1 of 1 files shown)


dlockhart:
Yeah I think we're going to want to fix this really soon (I may look at it this sprint if I have time) to make it completely fluid, aligned with the grid system and with the padding inside the nav. LG for now though.

PR #167, opened 2016-06-03
not including all the VUI icons CSS classnames

dlockhart:
@dbatiste, @njostonehouse, @capajon 

With the new Daylight icons, the size of `bsi.min.css` has gotten out of control:
- before Daylight icons: 137Kb
- with Daylight icons: 335Kb
- with no icons as all (after this PR): 92Kb

As you can see, the main problem is that we're including the CSS to use each icon as a background image by class-name. The monolith only uses those class names in a couple places, and even then has a way to fall back. So I'd like to remove this altogether with Daylight.

Corresponding PR in ILP:
https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/2660/overview

A few UMD-based FRAs [are referencing these class names](http://search.dev.d2l/source/search?q=%22vui-icon%22&defs=&refs=&path=&hist=&type=&project=LSOne), and I've reached out to them to try and get them to switch over to the mixins. In the meantime, I've manually embedded the ones they need.

### DIFF
--- a/src/vui.scss
+++ b/src/vui.scss
@@ -1,5 +1,5 @@
 @import '../bower_components/vui-focus/focus.css.scss';
-@import '../bower_components/d2l-icons/icons.css.scss';
+@import 'icons.scss';
 @import '../bower_components/jquery-vui-accordion/accordion.css.scss';
 @import '../node_modules/vui-breadcrumbs/breadcrumbs.css.scss';
 @import '../node_modules/vui-button/button.css.scss';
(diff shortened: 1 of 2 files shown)


src/vui.scss
```diff
@@ -1,5 +1,4 @@
 @import '../bower_components/vui-focus/focus.css.scss';
-@import '../bower_components/d2l-icons/icons.css.scss';
```
omsmith: Is there a dependency that can be removed now?

dlockhart: Unfortunately, no, because we still copy all the icons into the `/dist/images` directory and deploy those to the CDN. That's why all this CSS class business was so redundant... our users ended up downloading 2 copies of all the images eventually anyway.

PR #172, opened 2016-06-08
Adjusting dialog height in daylight

capajon:
@dlockhart Hoping this small change makes the BSI release today.

This is for https://rally1.rallydev.com/#/25052610463d/detail/defect/54447030760

It does help there.  However, I am worried that I am breaking other dialog stuff by removing it.  I am trying various dialogs out though, including on mobile, and it seems okay so far.  So I don't know.  Dialogs are a  bit fragile atm.


dlockhart:
Hmm, yeah hard to tell what that's going to do without a lot of testing. But clearly it's broken as it is now... and we're protected behind the flag, so worth a shot!

capajon:
*crosses fingers*

PR #203, opened 2016-06-28
US72043: Add iron-icon as a dependency for the LMS to be able to use …

njostonehouse:
…the provided icon sets.


dlockhart:
LGTM, thanks!

PR #209, opened 2016-07-08
upgrading to vui-loading-spinner 2.0.0

awikkerink:
@dlockhart @mdedys 

The ASV needs the new vui-loading-spinner in this rally story: https://rally1.rallydev.com/#/42651186647d/detail/userstory/50630921204

@mdedys is it ready? @dlockhart is a name change necessary?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -23,7 +23,6 @@
     "vui-focus": "~0.7.1",
     "vui-input": "~1.6.0",
     "vui-list": "~1.1.0",
-    "vui-loading-spinner": "~1.0.0",
     "vui-table": "~0.1.0",
     "vui-validation": "~1.1.0"
   },
--- a/bsi.html
+++ b/bsi.html
@@ -5,3 +5,4 @@
 <link rel="import" href="../d2l-icons/d2l-icons.html">
 <link rel="import" href="../d2l-navigation/navigation.html">
 <link rel="import" href="../iron-icon/iron-icon.html">
+<link rel="import" href="../vui-loading-spinner/loading-spinner.html">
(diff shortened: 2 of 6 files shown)


mdedys:
Yes that web component is ready. I have tested it and works. My only concern is that in your story it uses an orange color to it instead of the blue that I made it with. If it requires that then it would require a change to the library to support the two different colors.

awikkerink:
Maybe just add a color attribute?

awikkerink:
The color still needs to be defined. I assumed there was no difference. I'll go find out what it should be. I would think consistency is better?

ryantmer:
@awikkerink https://github.com/Brightspace/d2l-colors-ui is a thing! :) Has all the Daylight colours in it, if that's what ye need.

awikkerink:
The spinner uses svg I think, and so css won't work with it? (Unless the svg is inlined)

mdedys:
@ryantmer it is not as simple since the the outer circle is a png for the spinner...

ryantmer:
Ah, damn. Tbh only read the comments here, didn't look into it at all :stuck_out_tongue_winking_eye:

dbatiste:
Dave L. is out of office today so I can chime on on this.

@mdedys , @awikkerink :  Yeah I think we want to update the name of the component and the repo.  I've seen a few different spinner designs over the past couple of months, and of course I've seen the original one that you implemented quite a while ago.  Do you know if the design is all settled on that?  

I think there are a few minor updates that should be made in that repo - I will try to summarize them and pass them along.  

Related to the color and the svg: for the web component, I'd be inclined to use `iron-icon` and point it at the `svg` files rather than encoding in CSS.  It will inline the `svg`, and as you mentioned then the color can be controlled through CSS (fill would need to be excluded from the `svg`) and the predefined color variables can be used too.  This is the approach we're taking with new icons.  Further to that we can provide a var/mixin hook for overriding/theming, but that can be added at any time down the road.

awikkerink:
Closing to update to v3.0.0 instead

PR #211, opened 2016-07-12
US74104: Combine custom D2L link styles into one scss file. Add style…

njostonehouse:
…s for "flat" links.


src/links.scss
```diff
@@ -0,0 +1,55 @@
...
+}
+
+.d2l-clickable {
+	cursor: pointer;
+}
+
+.d2l-link.d2l-link-flat:not(:hover):not(:focus) {
+	color: inherit;
+}
+
```
dlockhart: I'm assuming all the CSS above here is coming from the monolith?

dlockhart:
LGTM

src/links.scss
njostonehouse: first: http://search.dev.d2l/source/xref/LSOne/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/Link/Link.css
second: http://search.dev.d2l/source/xref/LSOne/lp/framework/web/D2L.LP.Web.UI/Desktop/Controls/Link/Clickable.css
third: new

dlockhart: Excellent, so it'll move into a `DaylightOff` file at some point when this version of BSI is picked up. Thanks!

src/links.scss
```diff
@@ -0,0 +1,55 @@
...
+.d2l-link, .d2l-link:visited {
+	display: inline-block;
+}
+
+.d2l-clickable {
+	cursor: pointer;
+}
+
+.d2l-link.d2l-link-flat:not(:hover):not(:focus) {
```
dbatiste: My understanding is that the "flat" links should inherit the color from its context, and should have the same hover/focus color as normal links.  Is that correct?

Could this be done without the double `:not` pseudo-classes?  For example:
```
.d2l-link.d2l-flat {
   color: inherit;
}
```
Or does that interfere with the hover and focus colors too?  If so, a more explicit approach would be to provide corresponding `:hover` and `:focus` selectors that explicitly set the color.

The reason I ask is that, I think `:not` wasn't too good performance-wise in the past.  I tend to think (right or wrong) of it like doing a NOT predicate in a db query.  Browsers are much more performant now and they do a bunch of processing with the selectors, but I still wonder about cases like this.

src/links.scss
njostonehouse: Yep, right. It'll be part c of the series of PRs I'm opening up in the Monolith.

src/links.scss
njostonehouse: In my testing that selector also affected `:hover` and `:focus`, yes. When I'd initially implemented this in the LMS I didn't have access to the link mixins, so I used the nots. I might be able to more easily get the "standard" hover and focus style colours here.

PR #218, opened 2016-07-14
Update d2l-my-courses to v0.8.1

ryantmer:
This version isn't available quite yet, but Travis is being incredibly slow, so CI is many minutes to start anyway... by the time it does, this version should be up.


ryantmer:
Aight, tis up. Travis seems to be speeding up a bit, yay!

PR #228, opened 2016-07-29
US72882: Update icons dependency to include a few more icons.

njostonehouse:
Updating dependency on icons repo to automatically pull in minor version bumps: Typically adding new icons. Also resolve the icons repo to the latest version to avoid version conflicts with dependencies.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -3,7 +3,7 @@
   "dependencies": {
     "d2l-button": "~3.0.0",
     "d2l-colors": "~2.0.0",
-    "d2l-icons": "~2.1.0",
+    "d2l-icons": "^2.5.0",
     "d2l-image-action": "~0.4.3",
     "d2l-link": "~3.0.0",
     "d2l-loading-spinner": "~4.0.0",
@@ -27,5 +27,8 @@
     "vui-table": "~0.1.0",
     "vui-validation": "~1.1.0"
   },
-  "private": true
+  "private": true,
+  "resolutions": {
+    "d2l-icons": "^2.5.0"
+  }
 }
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,7 @@
     "d2l-link": "~3.0.0",
     "d2l-loading-spinner": "~4.0.0",
     "d2l-my-courses": "https://github.com/Brightspace/d2l-my-courses-ui.git#0.8.4",
-    "d2l-navigation": "https://github.com/Brightspace/d2l-navigation-ui.git#0.7.3",
+    "d2l-navigation": "https://github.com/Brightspace/d2l-navigation-ui.git#0.7.4",
     "d2l-offscreen": "~2.1.0",
     "d2l-typography": "~4.0.1",
     "iron-icon": "~1.0.8",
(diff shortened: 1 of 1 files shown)


njostonehouse:
@capajon this should grab our latest icons changes.

bower.json
```diff
@@ -27,5 +27,8 @@
     "vui-table": "~0.1.0",
     "vui-validation": "~1.1.0"
   },
-  "private": true
+  "private": true,
+  "resolutions": {
+    "d2l-icons": "^2.5.0"
```
dbatiste: What happens (in the build) if there a dependency conflict and one of the dependencies needs a newer version (ex. v2.6)?

njostonehouse: From my testing, it seems that the `^` will allow the resolved version to be "bumped" up to a version that matches the lowest version allowed by the dependencies (without raising the major version number). 

Examples: 
Resolution set to `^2.5.0` with a dependency set to `~2.5.0` and `~2.1.0` will result in using `^2.5.0`.
Resolution set to `^2.0.0`, with a dependency set to `~2.5.0` and another one set to `~2.1.0` will result in using `2.1.0`. 
Resolution set to `^1.0.0`, with a dependency set to `~2.5.0` will result in a build error: Unsuitable resolution declared for d2l-icons: ^1.0.0`

We're comfortable with this minor version bump in the BSI context for icons, since non-breaking changes to that repo should be smoothly incorporated into the dependent projects, and BSI should be the source of truth for the set of icons to be using in the LMS.

njostonehouse: @dlockhart @dbatiste @capajon Still discussing this approach for dependency resolution. Interested in continuing this discussion on Tuesday (Aug 2) before merging.

omsmith: Uhg, just use carats. Those examples are so awful!

ryantmer: Just came across this very conflict in #229 in fact. We're at `^2.4.0`, so `^2.5.0` is a-okay by me :+1:

_Resolution set to `^2.0.0` with a dependency set to `~2.5.0` and another one set to `~2.1.0` will result in using `2.1.0`_
I'm confused by this one - why would it use the lower version of the dependencies, since the dependency that requires `~2.5.0` could be depending on features that got introduced in 2.2.0-2.5.0?

njostonehouse: I would've expected example 2 to resolve to 2.5.0 too, but I think the idea is that the resolution (`^2.0.0`) is okay with either one, so it adheres to the "closest." I also could've screwed up in testing.

@omsmith I would definitely prefer all the repos pulling in the icons repo use carats. Using `resolution` was my attempt to "encourage" that from here, rather than having to monitor the dependencies individually.

ryantmer: Ah, could be. Bower + version resolution = a bad day, it seems :stuck_out_tongue_winking_eye: Also #humblebrag - [carets all day err day](https://github.com/Brightspace/d2l-my-courses-ui/blob/master/bower.json)

njostonehouse:
I believe these PRs will bring the troublesome dependencies up to a compatible version for BSI, which would mean not having to use `resolutions`.

https://github.com/Brightspace/d2l-navigation-ui/pull/145
https://github.com/Brightspace/d2l-dropdown-ui/pull/64
https://github.com/Brightspace/d2l-hierarchical-view-ui/pull/11

PR #248, opened 2016-08-25
DE22615 - only hide overflow when the dialog is fullscreen

vsumner:


dlockhart:
Looks great @vsumner thanks! I'll merge this now. If you're ready to fix this in Brightspace, I can create a release of BSI and then you can update the config file in Brightspace to point at it.

PR #250, opened 2016-08-30
line height fix

dlockhart:
FYI @njostonehouse, @dbatiste, @capajon 

Using `normal` line-height instead of `100%` to avoid clipping of extenders (e.g. "y" and "g") in certain browsers. In general I think we should always be using `normal` when the desired effect is for the text to be "as short as possible". I'm going to adjust a few places in the LMS as well that are doing this.

Reference: http://stackoverflow.com/questions/25470084/minimum-line-height-to-ensure-text-is-not-cut-off

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
-node_js: stable
+node_js: "5"
 sudo: false
 after_success:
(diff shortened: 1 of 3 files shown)


njostonehouse:
LGTM

PR #260, opened 2016-09-13
Two panel responsive collapsing columns pass one

capajon:
@dlockhart This is related to the "useResponsiveCollapseLayout" changes in the LE that we were holding off putting in 10.6.6.

### DIFF
--- a/src/page-two-panel-selector.scss
+++ b/src/page-two-panel-selector.scss
@@ -58,3 +58,19 @@
 .d2l-twopanelselector-padding {
 	padding-top: 30px;
 }
+
+@media(max-width: 1000px) {
+	.d2l-twopanelselector-wrapper.d2l-repsonsive-collapse-layout {
+		.d2l-box-layout {
+			display: block;
+		}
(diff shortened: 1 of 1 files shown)


src/page-two-panel-selector.scss
```diff
@@ -58,3 +58,19 @@
 .d2l-twopanelselector-padding {
 	padding-top: 30px;
 }
+
+@media(max-width: 1000px) {
+	.d2l-twopanelselector-wrapper.d2l-repsonsive-collapse-layout {
+		.d2l-box-layout {
```
omsmith: All of these should likely be `> ...`

capajon: Right didn't notice that - seems critical for box-layout.  I don't think it should matter on the others but I can put it there too for safety. Will update.

omsmith:
Can't comment on the correctness, but looks safer

dlockhart:
Tried this out, seems to work well! LGTM

PR #261, opened 2016-09-14
fix selections for responsive twopaneselector

omsmith:
@dlockhart 
@capajon

### DIFF
--- a/src/page-two-panel-selector.scss
+++ b/src/page-two-panel-selector.scss
@@ -61,16 +61,13 @@
 
 @media(max-width: 1000px) {
 	.d2l-twopanelselector-wrapper.d2l-repsonsive-collapse-layout {
-		> .d2l-box-layout {
+		.d2l-twopanelselector-side {
 			display: block;
-			> .d2l-twopanelselector-side {
-				display: block;
-				max-width: none;
-				width: auto;
-			}
-			> .d2l-twopanelselector-main {
-				display: block;
-			}
+			max-width: none;
+			width: auto;
+		}
+		.d2l-twopanelselector-main {
+			display: block;
 		}
 	}
 }


dlockhart:
So this was working in content before because there are several ways this can get laid out?

omsmith:
This is in user progress, but I highly doubt this part of the structure varies from page to page. Are you sure  content ever worked after the direct descendent changes went in?

dlockhart:
Yeah, content is different without that title bar at the top:

screen shot 2016-09-14 at 4 27 53 pm

omsmith:
Aha! funfun

dlockhart:
Indeed! But yeah, this looks like a good fix.

omsmith:
I mean, this is obviously referring to different `d2l-box-layout`s, but it works swimmingly for me in user progress. I'm not sure the `display:block` change is even necessary for that.

omsmith:
Updated to not change up `> .d2l-box-layout`. Seems to still be okay to me in both user progress in content (though I obviously don't know exactly what content is meant to look like).

omsmith:
@capajon Did you consider `.d2l-twopanelselector-side-bg`, which appears to be a thing in content?

capajon:
@omsmith  Yeah I think that was a sibling that I couldn't easily target.... so technically it is spanning the whole page I think, but it effectively didn't matter because the right panel covered it up when collapsed.  That reminds me to look at it again though as I wasn't 100% satisfied with it.

omsmith:
Yeah, a short main panel definitely doesn't cover it :)

Might have to set a body-level class, or avoid adding the page bg in the first place

capajon:
Yeah the answer wasn't obvious so went ahead with it as-is knowing we would iterate on it (wanted to get something in there for now as the short content + small width & long enough viewport height is rare).
Thanks for cleaning up this CSS further btw. :)

PR #262, opened 2016-09-16
Adjust two panel mobile styling further

capajon:
@omsmith @njostonehouse 

There will be a corresponding change in the LE to add .d2l-repsonsive-collapse-layout to .d2l-twopanelselector-side-bg (but this can still go in without it).

Although with a 100vh on the main panel the bg wouldn't actually show anymore, but this feels more comprehensive.

### DIFF
--- a/src/page-two-panel-selector.scss
+++ b/src/page-two-panel-selector.scss
@@ -68,6 +68,10 @@
 		}
 		.d2l-twopanelselector-main {
 			display: block;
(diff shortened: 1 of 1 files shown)


capajon:
@omsmith If the 100vh on main doesn't work for user progress let us know - we can make it more specific to content.  Actually, maybe it should be, I might adjust this.

PR #264, opened 2016-09-16
using d2l-icon WC for image-actions and image-links

dlockhart:
@Surekh This switches BSI over to pull in the new version of `d2l-icons` that exposes the new `<d2l-icon>` web component. There'll be a corresponding PR into the monolith that will:

1. Switch the image-link control over to render `<d2l-icon>` instead of `<iron-icon>`. Then it will get the automatic RTL mirroring goodness.
2. Update to the version of BSI that this PR will be merged into.


bower.json
```diff
@@ -3,16 +3,15 @@
   "dependencies": {
     "d2l-button": "^3.0.4",
     "d2l-colors": "^2.1.0",
-    "d2l-icons": "^2.5.0",
-    "d2l-image-action": "~0.4.3",
+    "d2l-icons": "^2.8.0",
+    "d2l-image-action": "^0.5.0",
```
dlockhart: Also grabbing an upgraded version of image-actions (print, settings, etc.) that makes use of the new <d2l-icon> element.

src/icons.scss
```diff
@@ -1,14 +0,0 @@
-.d2l-iron-icon-18 {
```
dlockhart: This file isn't needed anymore, since the size will get set automatically by `<d2l-icon>`.

src/image-link.scss
```diff
@@ -1,11 +1,10 @@
...
-.d2l-imagelink iron-icon {
-	color: $d2l-color-tungsten;
+.d2l-imagelink d2l-icon {
 	transition: color 0.3s ease;
 }
 
-.d2l-imagelink:hover iron-icon,
-.d2l-imagelink:focus iron-icon {
+.d2l-imagelink:hover:not(.d2l-imagelink-disabled) d2l-icon,
+.d2l-imagelink:focus:not(.d2l-imagelink-disabled) d2l-icon {
```
dlockhart: These 2 lines actually fix a bug where the color was changing on-hover on disabled image-links.

PR #275, opened 2016-09-29
Changed shim opacity to 0.7 as per Nicole Hayne's request.

ian-wallace:


ian-wallace:
Thanks @dlockhart , 
I assume you have to merge the change because I don't have permissions on the brightspace repo.
Ian

PR #279, opened 2016-10-06
Fix legacy dialogs on mobile

capajon:
@njostonehouse @dlockhart 

The issue this is solving is legacy dialogs opening with a height of 1px on the content.  This happened because of some new responsive rules that set heights to 100%.  The JS tries to dynamically set the height based on a few (confusing) things - and that CSS messes it up.

The fix ended up being fairly simple, just using flexbox basically.  I'll note that I tried a lot of other things first in the LE, mostly with how the javascript worked, but it is really confusing and never quite worked well in all situations.  Flexbox seems to take care of things so I'm happy enough with it.  I don't think dialogs are any better/worse overall with this change (but mostly work now so that's good).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,6 +18,7 @@
     "jquery-vui-more-less": "^1.2.0",
     "jquery-vui-scrollspy": "~0.2.0",
     "polymer": "^1.5.0",
+    "sass-flex-mixin": "^1.0.3",
     "vui-breadcrumbs": "^2.1.0",
     "vui-dropdown": "~0.7.1",
     "vui-field": "^1.2.0",
--- a/src/dialog.scss
+++ b/src/dialog.scss
@@ -72,19 +72,6 @@ $dialog-resize: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.
 	width: 400px;
 }
 
-.ddial_o,
-.d2l-dialog {
-	@media #{$dialog-media-break-small} {
-		width: 100% !important;
-		height: 100% !important;
-		left: auto !important;
-		padding: 0 !important;
-		top: 0 !important;
-		position: fixed !important;
-	}
-}
-
-
 .d2l-dialog-document-body {
 	background-color: transparent;
 	height: 100%;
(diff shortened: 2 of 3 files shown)


bower.json
```diff
@@ -18,6 +18,7 @@
     "jquery-vui-more-less": "^1.2.0",
     "jquery-vui-scrollspy": "~0.2.0",
     "polymer": "^1.5.0",
+    "sass-flex-mixin": "^1.0.3",
```
capajon: I wanted to include all the flexbox prefixes and this seemed very good for that, but I never noticed any browser issues without them.  But you never know, maybe some BB user out there will thank us for it (in their heart).

src/dialog.scss
```diff
@@ -304,3 +272,53 @@ $dialog-resize: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.
 	}
 
 }
+
+@media #{$dialog-media-break-small} {
+
+	.ddial_o,
+	.d2l-dialog {
+		width: 100% !important;
```
capajon: An angel lost its wings everytime we included an !important like this but I'm not trying to solve dialogs here.

src/dialog.scss
```diff
@@ -304,3 +272,53 @@ $dialog-resize: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.
 	}
 
 }
+
+@media #{$dialog-media-break-small} {
```
capajon: I pulled most of the rules related to this breakpoint together, rather than per element which is more typical in scss.  I did it because these are very much a "bolt on" (hacky-ish) solution to making dialogs look nice on mobile, and so it felt nicer to be able to see them all together in one spot.

ryantmer:
@capajon We (by which I mean @apalaniuk) made the [`d2l-simple-overlay`](https://github.com/Brightspace/d2l-my-courses-ui/blob/master/d2l-simple-overlay.html) that is dialog-esque; in fact, it closely mimics the mobile dialog as shown on http://design.d2l/components/dialogs/. It would have to be broken out and changed to work as a normal dialog on mobile, but some of the work is there.

(Saying this, I realize the majority of the work would be actually using it within the LMS, to which I say... eek.)

bower.json
dlockhart: The prefixes should actually be getting added automatically by our post-build autoprefixer:
https://github.com/Brightspace/brightspace-integration/blob/master/package.json#L14

Not sure how good they are though.

capajon: Cool!  I didn't know at all!  

I might still keep this because it is so robust and seems to handle the IE differences you mentioned in another comment.

PR #297, opened 2016-10-27
Dbatiste/instrument

dbatiste:
@dlockhart : these changes add a couple of measurements, and logs the results to console (temp).  We can adjust as needed.  I expect the logging to console part to be temporary until we have a place to store metrics.

* update to latest d2l-navigation (to get `d2l.navigation` measurement)
* add `d2l.page.display` measurement (first meaningful paint)
* add `WebComponentsReady` measurement
* add `domInteractive` measurement
* log `domContentLoaded` duration
* log `load` duration

We can differentiate between before & after handlders for `domContentLoaded` and `load`.  Currently I have them aligned with the chrome timeline (blue and red lines respectively).

I don't really like all the measurements that go from 0 -> x on the timeline.  I wish simple markers could be shown on the timeline.  Anyway, maybe we tweak when we know better what's useful.

### DIFF
--- a/src/page-loading/page-loading.js
+++ b/src/page-loading/page-loading.js
@@ -10,8 +10,10 @@
 		}
 
 		var timing = performance.timing;
+
+		D2L.Performance.measure('domInteractive', 'navigationStart', 'domInteractive');
+
 		var measures = {
-			'domInteractive': timing.domInteractive - timing.navigationStart,
 			'domContentLoaded': timing.domContentLoadedEventEnd - timing.navigationStart,
 			'load': timing.loadEventStart - timing.navigationStart
 		};
--- a/src/page-loading/page-loading.js
+++ b/src/page-loading/page-loading.js
@@ -23,7 +23,7 @@
 			measures[custom[i].name] = Math.floor(custom[i].duration);
 		}
 
-		console.log('measures', measures);
+		D2L.Performance.timing = measures;
 
 	}
 
--- a/src/page-loading/page-loading.js
+++ b/src/page-loading/page-loading.js
@@ -1,34 +1,73 @@
 (function() {
 	'use strict';
 
-	var pageReady, navReady = false;
+	var pageReady, pageLoaded, navReady = false;
+
+	function logMeasures() {
+
+		if (!pageReady || !navReady || !pageLoaded) {
+			return;
+		}
+
+		var timing = performance.timing;
+		var measures = {
+			'domInteractive': timing.domInteractive - timing.navigationStart,
+			'domContentLoaded': timing.domContentLoadedEventEnd - timing.navigationStart,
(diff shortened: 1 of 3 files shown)


src/page-loading/page-loading.js
```diff
@@ -1,34 +1,75 @@
...
+		if (!pageReady || !navReady || !pageLoaded) {
+			return;
+		}
+
+		var timing = performance.timing;
+
+		D2L.Performance.measure('domInteractive', 'navigationStart', 'domInteractive');
+
+		var measures = {
+			'domContentLoaded': timing.domContentLoadedEventEnd - timing.navigationStart,
```
dbatiste: Did not make explicit measurements for timeline for these because they are already on the timeline as blue and red lines.  We want them logged though.

src/page-loading/page-loading.js
```diff
@@ -1,34 +1,75 @@
...
+
+		console.log('measures', measures);
+
+	}
 
 	function check() {
 		if (pageReady && navReady) {
+			D2L.Performance.measure('d2l.page.display');
 			document.body.classList.remove('d2l-page-loading');
+			logMeasures();
```
dlockhart: Is calling that on page load not enough? Seems like it'll get called twice but only the one from on-load will actually do anything.

src/page-loading/page-loading.js
```diff
@@ -1,34 +1,75 @@
...
+			'domContentLoaded': timing.domContentLoadedEventEnd - timing.navigationStart,
+			'load': timing.loadEventStart - timing.navigationStart
+		};
+
+		var custom = D2L.Performance.getEntriesByType('measure');
+		for(var i=0; i<custom.length; i++) {
+			measures[custom[i].name] = Math.floor(custom[i].duration);
+		}
+
+		console.log('measures', measures);
```
dlockhart: How will we remember to get rid of this by next week before the release? Another option might be to store it somewhere and then we can console.log it manually if need be.

src/page-loading/page-loading.js
dbatiste: There is no guarantee that `load` will fire before or after the (`pageReady && navReady`) criteria is met (ex. since nav is asynchronous).

src/page-loading/page-loading.js
```diff
@@ -1,34 +1,75 @@
...
+			'domContentLoaded': timing.domContentLoadedEventEnd - timing.navigationStart,
+			'load': timing.loadEventStart - timing.navigationStart
+		};
+
+		var custom = D2L.Performance.getEntriesByType('measure');
+		for(var i=0; i<custom.length; i++) {
+			measures[custom[i].name] = Math.floor(custom[i].duration);
+		}
+
+		D2L.Performance.timing = measures;
```
dbatiste: Added the timing here rather than writing to console.

PR #301, opened 2016-11-01
Dbatiste/update d2l performance

dbatiste:
@dlockhart : this updates to latest `d2l-performance` which includes the name check.  It also updates the measurement names as follows:

d2l.navigation
d2l.page.display
d2l.page.domContentLoaded
d2l.page.domInteractive
d2l.page.load
d2l.page.webComponentsReady

Let me know of you think we should change any of these.


dlockhart:
Looks good!

src/page-loading/page-loading.js
```diff
@@ -11,11 +11,11 @@
 
 		var timing = performance.timing;
 
-		D2L.Performance.measure('domInteractive', 'navigationStart', 'domInteractive');
+		D2L.Performance.measure('d2l.page.domInteractive', 'navigationStart', 'domInteractive');
 
 		var measures = {
-			'domContentLoaded': timing.domContentLoadedEventEnd - timing.navigationStart,
-			'load': timing.loadEventStart - timing.navigationStart
+			'd2l.page.domContentLoaded': timing.domContentLoadedEventEnd - timing.navigationStart,
```
dbatiste: It is worth pointing out that this is the point in time immediately after `domContentLoaded` event handlers have run.  This appears to be consistent with what Chrome displays on the timeline.  If desired, we can capture the time before handlers are run, since that value is also available.

src/page-loading/page-loading.js
```diff
@@ -11,11 +11,11 @@
...
+			'd2l.page.load': timing.loadEventStart - timing.navigationStart
```
dbatiste: Similarly, this is using the point in time immediately before `load` handlers are run, and appears consistent with what Chrome shows in the timeline.  We have the option to capture the time after the handlers are run if desired.

PR #302, opened 2016-11-01
resovled posthtml issue while run BSI build script

edwardg8:


capajon:
LGTM, works in Travis too.  Good work. :)
For anyone else viewing, this was primarily an issue in Windows.

PR #304, opened 2016-11-02
no longer waiting for the "nav ready" event to fade the page in

dlockhart:
@dbatiste look OK? Basically only waiting for the WC ready event before fading in the page. I still want to do a lot of work to navigation so that it progressively loads while still showing _something_, but that's separate.


dbatiste:
Looks good.

PR #305, opened 2016-11-02
Move the performance include to bsi.js.

dbatiste:
@dlockhart : this is the change to load d2l-performance from bsi.js.


dlockhart:
@dbatiste LGTM

PR #311, opened 2016-11-04
firing an event when perf measures are ready so we can do things with them

dlockhart:
@dbatiste @omsmith The main use case for this right now is so that we can grab this data when it's ready and send it off to AppDynamics.

### DIFF
--- a/src/page-loading/page-loading.js
+++ b/src/page-loading/page-loading.js
@@ -19,8 +19,15 @@
 		};
 
 		var custom = D2L.Performance.getEntriesByType('measure');
-		for(var i=0; i<custom.length; i++) {
+		for (var i = 0; i < custom.length; i++) {
 			measures[custom[i].name] = Math.floor(custom[i].duration);
+		}
+
+		for (var m in measures) {
+			document.dispatchEvent(new CustomEvent('D2LPerformanceMeasure', {
+				bubbles: true,
+				detail: { name: m, value: measures[m] }
+			}));
 		}
 
 		D2L.Performance.timing = measures;


omsmith:
LGTM

I've noticed sometimes I don't get all of the measures. I figured I was grabbing them too early, but this suggests they should all be ready by the time it's set...

Sometimes navigation isn't there. I've only seen webComponentsReady on the object once. (IE11). I dunno

dlockhart:
Yeah that makes some amount of sense. WebComponentsReady actually only fires in browsers other than Chrome, since it'll have native support and won't fire the event.

Navigation should always be there as long as you wait long enough for it. Dave and I chatted about changing this slightly so that this code isn't really needed at all and instead each timing event just added to the performance timing object itself. So then this part would change to instead of firing a single event, it would fire one for each bit of information. In fact... I think I'll do that part right now so I don't need to update the monolith later.

awikkerink:
"WebComponentsReady actually only fires in browsers other than Chrome, since it'll have native support and won't fire the event." Actually, it does fire in Chrome, as long as the polyfills are included.

dlockhart:
@awikkerink Yep, but we're no longer including the polyfills if we detect they're not needed, which is the case in Chrome. Yay performance bump!

dlockhart:
Alright, I've updated this to fire an event for each measurement we receive, which will future-proof the tracking code in the monolith and allow it to continue to receive performance events (and send them on to AD) throughout the lifetime of each page.

dbatiste:
@omsmith : re not seeing the navigation measurement - the navigation component is currently asynchronous as a result of the HM Ajax call, so I've observed on some pages the `load` event firing before the nav component is complete.  Assuming you wait long enough, I would expect the nav measurement to be present.

PR #320, opened 2016-11-17
Reduce d2l-table action offset from 15px to 10px when in a dialog

awikkerink:


awikkerink:
@dlockhart @dbatiste is this correct?

dbatiste:
I think so.

PR #328, opened 2016-11-28
US79943: Update BSI with heavier, smaller root font styles.

njostonehouse:


njostonehouse:
(needs typography 5.2 to be published)

PR #330, opened 2016-11-29
Removing some background colors to allow for new nav shadow

capajon:


capajon:
This will likely change a bit still as I look at eliminating some divs in the LE.

PR #340, opened 2016-12-02
Page loading tweaks

dlockhart:
Simplifying how the page-loading code works:
- since navigation is now not a dependency on fading the page in, removing all references to it
- renaming the `domContentLoaded` measure to `domContentLoadedHandlers` to more accurately reflect its purpose
- logging the performance measures in the `onload` callback, since waiting for the `pageReady` isn't necessary
- no longer measuring the `webComponentsReady` event since it always corresponds to the `page.display` event anyway, and in cases where the polyfill wasn't required (Chrome), it wasn't getting logged at all.

@dbatiste thoughts?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,5 +6,8 @@
   },
   "rules": {
       "strict": 2
+  },
+  "globals": {
+    "D2L": true
   }
 }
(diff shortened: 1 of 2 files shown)


dbatiste:
Looks good!  Much simpler!

PR #350, opened 2016-12-07
Update to latest d2l-colors.

dbatiste:
@awikkerink : this is the update to BSI for the various version updates related to d2l-colors. I've updated the version for tables here as well.


awikkerink:
I also updated d2l-button-group. There are some more substantial changes in it, but it did test it in different browsers, so it should be good to go.

dbatiste:
Ok, sounds good.  I will update to the latest version of that too.

PR #358, opened 2016-12-12
US80190: Add a "d2l-heading-strong" class to modify headings to have …

njostonehouse:
…a stronger font weight.

### DIFF
--- a/headings.html
+++ b/headings.html
@@ -47,13 +47,9 @@
 				padding-bottom: 0;
 			}
 
-			.d2l-typography .d2l-heading-strong {
+			.d2l-typography .d2l-heading-1.d2l-heading.strong,
+			.d2l-typography .d2l-heading-2.d2l-heading-strong {
 				font-weight: 700;
-			}
-
-			.d2l-typography .d2l-heading-3.d2l-heading-strong,
-			.d2l-typography .d2l-heading-4.d2l-heading-strong {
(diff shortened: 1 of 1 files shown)


njostonehouse:
@dbatiste Does this seem reasonable? LinkHeadings used to have a bold and a not-bold version (awkwardly named "level2" and "level3") that got lost when those "levels" were changed to match heading levels for Daylight. This provides a way to apply a heavier weight to the headings without changing the rest of the style.

dbatiste:
I suppose if that's what we need.  Looking at the 900 weight, it seems like a subtle difference.  Hopefully it's different enough that it doesn't look _mis-matchy._

Though... just as we remove light (one less network request), we add an extra bold.  :(

njostonehouse:
Currently, nothing uses the 900 weight, it just seemed weird to let some heading values get bolded and not others. Would you prefer I remove it for now, and only add it in when there's a use case for it?

dbatiste:
It's currently not included in the [font assets](https://github.com/Brightspace/font-generator/blob/master/src/config.json), so 900 isn't currently downloaded.  I'd rather not include it unless necessary.  Does the new typography design call for it?

njostonehouse:
No, it doesn't. It was something I included for completion. I'm removing it, and some of the other unused code to be more succinct.

PR #379, opened 2017-01-06
remove page-loading fade-in

dlockhart:
@dbatiste look OK? This CSS along with the `d2l-page-loading` class is no longer needed now that we're not waiting for the navigation to load before fading the page in. Instead, we can rely on the `unresolved` attribute on the body, which Polymer will remove for us and supply the opacity CSS.

I'll make a PR in platformTools to remove the class names and update BSI.

### DIFF
--- a/src/bsi.scss
+++ b/src/bsi.scss
@@ -10,7 +10,6 @@
 @import 'list.scss';
 @import 'navigation/index.scss';
 @import 'page.scss';
-@import 'page-loading/page-loading.scss';
 @import 'page-two-column.scss';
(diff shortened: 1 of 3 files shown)


omsmith:
is d2l.page.display going to make very little sense now, or should the timing still be equivalent?

dlockhart:
It'll fire at the same time as `WebComponentsReady`, which is when Polymer will fade the page in, so it should be the same as previously. Originally we were _also_ keeping everything hidden except the loading bar while navigation made its API calls, but that was removed in November but the then-redundant CSS wasn't.

PR #387, opened 2017-01-11
Nav s mobile menu

capajon:
First iteration for mobile wrapper placeholder


src/navigation/header.scss
```diff
@@ -24,3 +26,43 @@
 	flex: 0 0 auto;
 	height: 100%;
 }
+
+.d2l-navigation-header-open-button-wrapper {
+	height: 100%;
```
dlockhart: Not sure if it matters, but the WC version of this has `display:inline-block` and since you're using a `<div>`, it'll be `display:block`.

src/navigation/header.scss
```diff
@@ -1,5 +1,7 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
 
+$mobile-menu-bp: 992px;
```
dlockhart: Might be good at some point to move these breakpoints into `common.scss` so all the files can reference them.

src/navigation/header.scss
capajon: seems okay so far but I'll double check

PR #399, opened 2017-01-16
reduce title size to 25px

edwardg8:
It seems that we don't have to add another css class for it since the reduced title is applied whenever a title is present at header-logo-area.

### DIFF
--- a/src/navigation/header.scss
+++ b/src/navigation/header.scss
@@ -58,7 +58,7 @@
 
 .d2l-navigation-s-header-logo-area > .d2l-navigation-s-link {
 	display: block;
-	font-size: 25px;
+	font-size: 1.25rem;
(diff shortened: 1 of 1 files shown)


src/navigation/header.scss
```diff
@@ -58,7 +58,7 @@
 
 .d2l-navigation-s-header-logo-area > .d2l-navigation-s-link {
 	display: block;
-	font-size: 1.5rem;
+	font-size: 25px;
```
dlockhart: I think you want to keep this a root-em unit. The root font size is `20px`, so a value of `0.8rem` would give you `25px`. This way if the user increases or decreases their font size in account settings, the 25px will go up or down slightly.

dlockhart: Sorry, `1.25rem` to get to 25 :)

PR #401, opened 2017-01-17
Add direct reference to new d2l-siren-parser

ryantmer:
No other deps changed since our last release for 10.6.10, so only this one needs to be updated.


awikkerink:
Did you mean to close this?

ryantmer:
Yarr - see discussion in #daylight-dev. Not needed, did it the way Owen suggested instead :)

PR #402, opened 2017-01-18
css to have large and small logo size

edwardg8:

PR #404, opened 2017-01-18
Us81010 nav s mobile menu

capajon:


src/navigation/mobile-menu.scss
```diff
@@ -168,3 +171,55 @@
...
+	-webkit-flex-grow: 1;
+	-ms-flex-grow: 1;
+	flex-grow: 1;
+
+	.d2l-navigation-s-link {
+		font-size: 1rem;
+	}
+
+	.d2l-navigation-s-logo {
+		img {
```
dlockhart: I think this is already in some of the work that @edwardg8 did in [another pull request](https://github.com/Brightspace/brightspace-integration/pull/402). Probably best to co-ordinate those 2...

src/navigation/common.scss
```diff
@@ -3,6 +3,9 @@
 $d2l-navigation-header-height: 90px;
 $d2l-navigation-header-height-mobile: 72px;
 $d2l-navigation-mobile-menu-bp: 992px;
+$d2l-navigation-mobile-menu-title-bp: 660px;
+$d2l-navigation-mobile-menu-mid-bp: 615px;
+$d2l-navigation-mobile-menu-small-bp: 380px;
 $d2l-navigation-header-height: 90px;
 $d2l-navigation-header-height-mobile: 72px;
```
dlockhart: Just noticed these last 2 variables are defined twice, probably worth cleaning up while we're in here.

PR #408, opened 2017-01-20
Change when logo and title are shown for various breakpoints and scen…

capajon:
…arios

### DIFF
--- a/src/navigation/header.scss
+++ b/src/navigation/header.scss
@@ -56,13 +56,13 @@
 			display: none;
 		}
 		.d2l-navigation-s-home-icon,
-		&.d2l-navigation-s-title-shown .d2l-navigation-s-home-divider,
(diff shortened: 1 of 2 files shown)


src/navigation/header.scss
```diff
@@ -52,9 +52,21 @@
 	padding: 0 7px;
 
 	.d2l-navigation-s-header-container & {
-		@media(max-width: $d2l-navigation-mobile-menu-bp) {
+		@media(max-width: $d2l-navigation-mobile-menu-title-bp) {
 			display: none;
 		}
+		.d2l-navigation-s-home-icon,
+		&.d2l-navigation-s-title-shown .d2l-navigation-s-home-divider,
+		&.d2l-navigation-s-title-shown .d2l-navigation-s-logo {
```
dlockhart: As mentioned in the other PR, I think you can use the existing `.d2l-navigation-s-has-title` class that's already on the root element.

PR #409, opened 2017-01-20
"jsonp" imports for IE should be getting gzipped.

omsmith:
give them a .js extension instead which makes more sense anyway

https://github.com/Brightspace/webcomponentsjs/pull/1


omsmith:
I regretted setting the extension as "jsonp" pretty quick, but ignored it.

PR #411, opened 2017-01-20
fixes #400: using html-minifier instead of posthtml to minify WC files

### LINKED ISSUE #400: posthtml is not working (discovered in Linux)
In posthtml-cli, the error was not being caught. node v7 realized this by showing an uncaught rejected promise. Here is the error from posthtml-cli:
```
TypeError: Cannot read property 'minifier' of undefined
    at /home/awikkerink/dev/brightspace-integration/node_modules/deepmerge/index.js:50:55
    at Array.forEach (native)
    at mergeObject (/home/awikkerink/dev/brightspace-integration/node_modules/deepmerge/index.js:49:25)
    at deepmerge (/home/awikkerink/dev/brightspace-integration/node_modules/deepmerge/index.js:67:16)
    at /home/awikkerink/dev/brightspace-integration/node_modules/post-load-plugins/node_modules/post-config/lib/modules.js:2:1375
    at Array.forEach (native)
    at /home/awikkerink/dev/brightspace-integration/node_modules/post-load-plugins/node_modules/post-config/lib/modules.js:2:934
    at Array.forEach (native)
    at _class.configExtends (/home/awikkerink/dev/brightspace-integration/node_modules/post-load-plugins/node_modules/post-config/lib/modules.js:2:418)
    at exports.default (/home/awikkerink/dev/brightspace-integration/node_modules/post-load-plugins/node_modules/post-config/lib/index.js:1:624)
```

It appears to be looking for `html` and `posthtml` properties from package.json.

dlockhart:
`posthtml` (and the CLI and its dependencies) have one problem after another. Even without the config file, I couldn't get them to behave with the `posthtml-minifier` plugin.

So... easier to just call directly into the code that the posthtml-minifier plugin was using directly! Which is what this does. Cuts out the middle-man and all that.

The minifier doesn't like having non-HTML things in the `--input-dir`, so this required that I dump the results of `web-component-shards` into `tmp-wc` and then minify from there into `dist`.


html-minifier.conf
```diff
@@ -0,0 +1,6 @@
+{
+	"collapseWhitespace": true,
+	"minifyCSS": true,
+	"minifyJS": true,
+	"removeComments": true
```
awikkerink: `ignoreCustomComments`?

dlockhart: Good catch -- fixed and confirmed that `@license` comments are still present.

dlockhart:
CI failed due to some changes @ryantmer is making to app-localize... I think once his other PR is merged this should work again.

ryantmer:
Yep, sorry about that - merged now, I re-ran them should be good now.

ryantmer:
Oh, I think this will have to be rebased due to the changes to bower.json, my bad :(

dlockhart:
Yeah no worries, on it.

PR #413, opened 2017-01-24
Move page-loading from bsi.js to bsi.html to ensure WebComponents are…

awikkerink:
… loaded first

### DIFF
--- a/bsi.html
+++ b/bsi.html
@@ -1,3 +1,5 @@
+<script src="src/page-loading/page-loading.js"></script>
+
 <link rel="import" href="bsi-style.html">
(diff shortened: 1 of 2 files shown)


awikkerink:
I'm just going to try to reproduce the issue on my local instance and make sure everything is working with the new BSI.

dlockhart:
Cool sounds good. I find it happens very often if you just stay on the same page and hit ENTER in the address bar -- anything that will indicate to the browser to use as much of the cache as possible.

awikkerink:
Verified with Fiddler and Autoresponder on Firefox

PR #416, opened 2017-01-25
(perf) Replace >:last-child,>:first-child,>:only-child

awikkerink:
~35% improvement for every style calculation

On the user_list_view.d2l page with 100 users displayed and 3 grades +1 final grade
Before: 19.97s
After: 18.81s

### DIFF
--- a/src/deprecated/table.scss
+++ b/src/deprecated/table.scss
@@ -47,7 +47,8 @@ $vui-table-row-background-color-selected: $d2l-color-celestine-light-1;
 		@include vui-table-cell();
 	}
 
-	& tr > :first-child {
+	& tr > th:first-child,
+	& tr > td:first-child {
 		border-left: $vui-table-border;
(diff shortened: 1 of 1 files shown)


src/deprecated/table.scss
```diff
@@ -171,7 +218,7 @@ $vui-table-row-background-color-selected: $d2l-color-celestine-light-1;
 	border-top: $vui-table-border;
 	border-right: $vui-table-border;
 	display: table-cell;
-	font-weight: 400;
+	font-weight: 300;
```
dbatiste: Note: the light (300) font-weight is no longer included.

dlockhart: Yeah good catch -- @awikkerink this should be 400 as 300 is no longer used (or even available). The browser is going to simulate a 300 weight which will likely look terrible.

PR #417, opened 2017-01-25
set min/max width to twopanelselector left and main

edwardg8:


src/page-two-panel-selector.scss
```diff
@@ -15,6 +15,15 @@ $d2l-twopanelselector-width: 320px;
 	width: $d2l-twopanelselector-width;
 }
 
+.daylight .d2l-twopanelselector-side {
+	min-width: $d2l-twopanelselector-width;
+}
+
+.daylight .d2l-twopanelselector-main {
+	max-width: 826px;
```
edwardg8: I couldn't find a way to set max-width without hard coding pixel value. I think it's because css styles (table), so setting percentage will end up being undefined width.

src/page-two-panel-selector.scss
```diff
@@ -15,6 +15,15 @@ $d2l-twopanelselector-width: 320px;
 	width: $d2l-twopanelselector-width;
 }
 
+.daylight .d2l-twopanelselector-side {
```
dlockhart: You don't need `.daylight` in front of these -- the master branch of BSI is for Daylight always.

dlockhart: Which means this rule can be combined with the one above it.

src/page-two-panel-selector.scss
dlockhart: Ditto for not needing `.daylight` here. How are you calculating this `826px` value?

src/page-two-panel-selector.scss
edwardg8: that's nice!!

omsmith: As an informational note, unnecessary selector depth will also slow down the page unnecessarily. Since CSS selectors are processed right-to-left, in this case it's forcing the browser to search up the dom to the `body` (basically the worst case) in order to find `.daylight` before it can apply the style.

src/page-two-panel-selector.scss
```diff
@@ -12,9 +12,14 @@ $d2l-twopanelselector-width: 320px;
 
 .d2l-twopanelselector-side {
 	max-width: $d2l-twopanelselector-width;
+	min-width: $d2l-twopanelselector-width;
 	width: $d2l-twopanelselector-width;
 }
 
+.d2l-twopanelselector-main {
+	max-width: 826px;
```
dlockhart: My earlier comment got lost, but where is this `826px` value coming from? How did you calculate it?

src/page-two-panel-selector.scss
edwardg8: Interesting, that's good to know!! So the broswer is looking for something very specific first, then more and more general?

src/page-two-panel-selector.scss
edwardg8: So basically by inspecting div elements and I used page ruler to measure pixels.

edwardg8: let me try with more accurate value by subtracting from 1230px, d2l-max-width.

edwardg8: Earlier I excluded padding which seems unnecessary. The rendered content pane width is always 909px if max-width is less than it (weird) so I didn't notice. It's enough to get it by 1230 (parent)-321(left).

PR #419, opened 2017-01-25
De23947 responsive dividers nav

capajon:
I also changed the breakpoint variable names a bit.  Probably should have done that separately after, sorry.


dlockhart:
@capajon @edwardg8 probably good to get this PR in with Edward's, along with the corresponding changes to the LMS and then do a BSI release?

capajon:
Thanks!

PR #423, opened 2017-01-26
De23098 logo focus

edwardg8:


src/navigation/logo.scss
```diff
@@ -19,6 +23,16 @@
 	max-width: $d2l-navigation-small-logo-width;
 }
 
+.d2l-navigation-s-logo:hover .d2l-navigation-s-button-highlight-bar,
+.d2l-navigation-s-logo-link:focus .d2l-navigation-s-button-highlight-bar {
```
edwardg8: we may have to change the class name of d2l-navigation-s-button-highlight-bar and place it elsewhere since it can be applied to not only buttons.

src/navigation/logo.scss
```diff
@@ -19,6 +23,16 @@
...
+	display: block;
+}
+
+.d2l-navigation-s-logo:hover .d2l-navigation-s-logo-link,
+.d2l-navigation-s-logo-link:focus {
+	box-shadow: 0 0 3px 0 rgba( $d2l-color-celestine, 0.4 );
+}
```
capajon: In SCSS you can nest selectors, so this could become:
.d2l-navigation-s-logo:hover,
.d2l-navigation-s-logo-link:focus {
  .d2l-navigation-s-button-highlight-bar {
    display: block
  }
  .d2l-navigation-s-logo-link {
    box-shadow: 0 0 3px 0 rgba( $d2l-color-celestine, 0.4 );
  }
}

Or something close to that (see my other comment/question)

src/navigation/logo.scss
```diff
@@ -19,6 +23,16 @@
 	max-width: $d2l-navigation-small-logo-width;
 }
 
+.d2l-navigation-s-logo:hover .d2l-navigation-s-button-highlight-bar,
+.d2l-navigation-s-logo-link:focus .d2l-navigation-s-button-highlight-bar {
+	display: block;
+}
+
+.d2l-navigation-s-logo:hover .d2l-navigation-s-logo-link,
+.d2l-navigation-s-logo-link:focus {
```
capajon: Just want to confirm this shouldn't be ".d2l-navigation-s-logo-link:focus .d2l-navigation-s-logo-link"

dlockhart: The first selector `.d2l-navigation-s-logo:hover .d2l-navigation-s-logo-link` doesn't seem right to me... why would we want the box-shadow to show up when the user hovers on anything? I think you just want the second selector.

src/navigation/logo.scss
```diff
@@ -1,7 +1,11 @@
 @import 'common.scss';
+@import '../../bower_components/d2l-colors/d2l-colors.scss';
 
 .d2l-navigation-s-logo {
-	display: inline-block;
+	align-items: center;
+	display: flex;
+	height: 100%;
+	position: relative;
```
dlockhart: Is the `align-items` and `display:flex` needed? Or could it work with just the `height:100%` and `position:relative`? I'd be surprised if the other two are required, since there isn't anything inside the logo that should require a flex layout.

src/navigation/logo.scss
edwardg8: Looks a lot cleaner with nest selectors

src/navigation/logo.scss
edwardg8: Yea, we don't want  ".d2l-navigation-s-logo:focus .d2l-navigation-s-logo-link" because d2l-navigation-s-logo won't get focus when you tab it.

src/navigation/logo.scss
dlockhart: Chatted offline about this, conclusion: yes they are needed

PR #430, opened 2017-01-27
use box-shadow instead of background color for button focus style

edwardg8:
Changed button's focus style from background color to shadow-box.

See DE23228 for the details.

### DIFF
--- a/button-menu.html
+++ b/button-menu.html
@@ -53,6 +53,11 @@
 				cursor: default;
 			}
 
+			.bsi-button-menu.bsi-button-menu-primary {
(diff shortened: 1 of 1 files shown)


dlockhart:
As we discussed, I think you'll want to start with the `d2l-button` component and create a mixin there to apply the focus style. You should then be able to use that mixin from here.

dlockhart:
@edwardg8 I'm going to close this since it's pretty stale and I don't think it's going to work with the new designs anyway. Can re-open something similar when/if that stuff is ready.

edwardg8:
Makes sense, thank you

PR #431, opened 2017-01-27
Remove Redundant table.css.scss

awikkerink:
* Saves ~6% per style recalc
* Galen says things haven't changed on Grades, Classlist, Discussions, or Org Unit Editor (in Chrome)


omsmith:
Out of curiosity, is that 6% since your last improvement, or 6% since before you started improving things (assumedly the first)?

awikkerink:
6% since master. So without my DataGrid changes

awikkerink:
It works out to be 23-28ms on grades list for 100 rows, and 3 grade items including final grades

PR #432, opened 2017-01-30
Remove expensive selector for IE11/Edge

awikkerink:
@dlockhart I'm not sure what the status of the icons are. How are they used in customizations? Do I need to add a `.d2l-icon-custom` selector as well? (For buttons)

### DIFF
--- a/src/deprecated/icons.scss
+++ b/src/deprecated/icons.scss
@@ -1,11 +1,10 @@
 @import '../../bower_components/d2l-icons/icons.scss';
 
-/* avoids colliding with the "d2l-icon" class that gets put on the
-  <d2l-icon> web component when shadow DOM isn't used */
-[class*=" d2l-icon-"]:not(d2l-icon),
-[class^="d2l-icon-"]:not(d2l-icon),
-[class*=" vui-icon-"],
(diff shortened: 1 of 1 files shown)


dlockhart:
Yes, you'll need `d2l-icon-custom` to catch [these cases](http://search.dev.d2l/source/search?q=%22d2l-icon-%22&defs=&refs=&path=&hist=&type=&project=Lms).

And `vui-icon-custom` for [these ones](http://search.dev.d2l/source/search?q=%22vui-icon-%22&defs=&refs=&path=&hist=&type=&project=Lms).

I think though you can get away without `vui-icon-delete`, `.vui-icon-edit`, `.vui-icon-file-pdf-xlarge`, `.vui-icon-remove`, `.vui-icon-share` and their definitions in `deprecated` though... I can't find any references to them anymore in code search (unless you did?).

awikkerink:
I haven't found any references. I just kept them in because they could be around.

awikkerink:
@dlockhart I added `d2l-icon-custom`, but not `vui-icon-custom` because `vui-icon-custom` is just used for daylight off.

PR #435, opened 2017-01-31
De23300 content viewer

edwardg8:


dlockhart:
Where is this going to be used in the LMS? Does it need to go in BSI or could it live with the content viewer or in our core CSS?

edwardg8:
I'm not sure what is the best practice here. I see the content-viewer pull-out uses links, the css for the links are from BSI (links.scss), so maybe it makes more sense in BSI?

edwardg8:
http://caniuse.com/#feat=word-break

PR #436, opened 2017-01-31
bump d2l-more-less

omsmith:


omsmith:
Adds `blur-color` support

PR #460, opened 2017-02-07
Extend problematic selector

awikkerink:


awikkerink:
https://rally1.rallydev.com/#/15545167705d/detail/defect/79884800536

The bottom border of a focused menu-button was being overwritten by the "last-item" selector. This solution forces SASS to put the selector after, and put higher precedence on the selector because multiple classes are specified.

awikkerink:
@njostonehouse I added you as a reviewer because you were in the original revision of the defect

PR #461, opened 2017-02-08
DE23802 allow separator target to be absolutely positioned within con…

capajon:
…tent left pane

Will have corresponding LE change, but safe to merge now.


capajon:
Won't be proceeding with this user story

PR #462, opened 2017-02-08
Add d2l-telemetry

awikkerink:


awikkerink:
@dlockhart @omsmith What do you think about merging this and #463 today? It won't actually do anything until the Monolith PR is merged, but then it's in.

PR #471, opened 2017-02-13
Set display:none for mobile nav menu in closed state

dbatiste:
This change is being made to prevent tabbing to mobile flyout children when in the closed state.

### DIFF
--- a/src/navigation/mobile-menu.scss
+++ b/src/navigation/mobile-menu.scss
@@ -1,6 +1,10 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
 @import 'common.scss';
 
+.d2l-navigation-s-mobile-menu[data-state="closed"] {
+	display: none;
+}
+
 .d2l-navigation-s-mobile-menu-content {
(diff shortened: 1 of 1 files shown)


src/navigation/mobile-menu.scss
```diff
@@ -1,6 +1,10 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
 @import 'common.scss';
 
+.d2l-navigation-s-mobile-menu[data-state="closed"] {
+	display: none;
+}
```
dlockhart: Good catch. Could `display: none` and all the different state transition styles applied to `d2l-navibation-s-mobile-menu-content` on lines ~30 be applied to the outer container now instead as well?

dbatiste: Hmmm maybe... I will explore that.  :)

dbatiste: The mask is in the outer container, so I _think_ we still want the transition on the inner container.  At a minimum though, I was able to remove these `display` properties from the inner/content container.

PR #482, opened 2017-02-15
Fix d2l.page.load event not being measured in IE11

awikkerink:
I just noticed telemetry wasn't being reported for IE11 with daylight on. This fixes it.

### DIFF
--- a/bsi.html
+++ b/bsi.html
@@ -1,4 +1,4 @@
-<script src="src/page-loading/page-loading.js"></script>
+<script src="src/page-loading/web-components-ready.js"></script>
 
 <link rel="import" href="bsi-style.html">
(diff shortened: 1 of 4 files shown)


dlockhart:
So just a timing issue based on the ordering of the event registration?

awikkerink:
What was happening is that the HTMLImports shim was loading stuff after the `load` event

joshassad:
Captured in Rally as [DE24297](https://rally1.rallydev.com/#/15545167705d/detail/defect/94899995820?fdp=true)

awikkerink:
`d2l.page.display` was already being reported to Google Analytics, so this isn't required for 10.6.11

PR #484, opened 2017-02-16
using the sass-based typography instead of web components

dlockhart:
@omsmith @dbatiste @njostonehouse I don't intend to merge this right away, just looking for feeback. There's a [corresponding PR](https://github.com/BrightspaceUI/typography/pull/83) in typography, which is why this is going to fail CI.

Problem: our fonts are loading in way too late in the rendering pipeline, in all browsers. The browser doesn't even know to fetch `fonts.css` from the CDN until after the web component code has initialized and run, then it still waits to download the `WOFF` files until it determines they're needed.

This PR consumes new Sass mixins defined in `d2l-typography`, similar to how things used to work pre-Daylight. It then overrides a few of them based on what was in `typography.html`.


src/navigation/main.scss
```diff
@@ -15,6 +15,7 @@
 	-webkit-flex-wrap: nowrap;
 	-ms-flex-wrap: nowrap;
 	flex-wrap: nowrap;
+	transition: opacity ease-in 0.2s;
```
dlockhart: Things render so quickly now without the page fade-in, the navigation links chunking in after they've calculated whether the "more..." menu is needed is very jarring. I may look at ways to do this calculation faster, but in the meantime this makes it more palatable.

dlockhart:
@omsmith @dbatiste @njostonehouse This is ready for prime-time now -- hope to merge today. Any last comments? I'm just performing some quick manual before/after performance tests, although I don't expect too much of a difference. The main improvement will be earlier font availability.

dlockhart:
Some really rough local machine performance comparisons

screen shot 2017-03-01 at 4 17 00 pm

The Chrome numbers are _slightly_ optimisitic because I added an additional 200ms to the "before" numbers (for Chrome only) beyond "d2l.page.display" to account for the fact that it no longer has the 200ms fade-in.

PR #489, opened 2017-02-18
smaller bsi.js bundle

omsmith:
Minified file is smaller by about 7K

* don't use the pre-minified lie bundle
* remove unnecessary wrapper around page-loading
* bundle scripts with rollup, less boilerplate than the browerify bundle

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "autoprefix": "postcss -c postcss.config.json",
     "prebuild": "rimraf dist && mkdir dist",
     "build": "npm run build:js && npm run build:sass && npm run build:icons && npm run build:wc",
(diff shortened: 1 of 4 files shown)


omsmith:
Haven't actually loaded it up in the LMS yet.

awikkerink:
I did a quick check of the bundle, and it appears the promises are being excluded. Do we actually need to include lie?

omsmith:
Yes it's needed for IE11. Looking into some other problems with the bundle as well

awikkerink:
It looks like the `function Promise { ... }` is not being included because it is being exported, and the `import 'blah'` only handles global stuff, not exported stuff.

omsmith:
Have it mostly fixed locally. Still saves ~5K. I'll push later once I figure out the final ponderings

PR #492, opened 2017-02-22
Add bower_components for OnPrem release

ryantmer:
This ensures that any further patches/changes that need to be made to the OnPrem (10.6.11) BSI release will use the same bower package versions that were initially deployed.

This also removes the bower install step from Travis - it seems that bower sort of just... ignores the bower_components folder, and fetches everything anew. It's entirely strange, and if that is indeed the case, entirely dumb, but this is one solution to that problem.


package.json
```diff
@@ -16,7 +16,6 @@
     "prebuild:wc": "rimraf tmp && rimraf tmp-wc",
     "build:wc": "web-component-shards -d tmp-wc/ --endpoints bsi.html d2l-my-courses.html d2l-navigation.html d2l-image-banner-overlay.html",
     "postbuild:wc": "npm run uglify:html && npm run jsonpify-imports",
-    "postinstall": "./node_modules/.bin/bower install",
```
ryantmer: The only actual change in here, aside from adding `bower_components`.

This was deemed necessary as `bower install` was just re-fetching everything rather than using the existing `bower_components` folder being committed here. This seems strongly related to behavior [like this](https://github.com/bower/bower/issues/2379), which was a similar issue that was closed as "won't fix" on bower.

ryantmer:
[v10.6.11-daylight.14 package versions](https://travis-ci.org/Brightspace/brightspace-integration/builds/200238179#L558). This was used to create 
[BSI-bower-install.txt](https://github.com/Brightspace/brightspace-integration/files/794735/BSI-bower-install.txt), which was what I used to install what is being committed here. In doing so, I found two "conflicts":

- Stickyfill is being addressed by both a commit hash, and by `#master` - right now, they point to the same thing, but either way, it was installed by the commit hash here
- In order to avoid a conflict in d2l-polymer-behaviors, I had to install the following packages in the given order:
  1. simple-overlay#0.0.2
  2. d2l-polymer-behaviors#0.0.4
  3. d2l-my-courses-ui#0.30.1

package.json
dlockhart: Yeah looking at the actual diff makes my browser dizzy, so looks good based on this one change!

PR #493, opened 2017-02-23
Fix jsonp match

ryantmer:
Was only ever matching the first file, meaning we were only getting bsi.html.js, not the rest of them. Introduced in https://github.com/Brightspace/brightspace-integration/commit/d16d58cbd21dee24c645c24829b079035fee6117 to fix this exact problem, seemingly.


omsmith:
Yeah, meh about Windows. Everything should work fine on WSL I think...

PR #495, opened 2017-02-23
Remove package.json version

ryantmer:
This just like... isn't true, right? And not needed? (not a big deal, at worst it's just slightly confusing)


dlockhart:
Agreed, not needed!

PR #500, opened 2017-02-24
Where did sliding tile behavior go?

ryantmer:
No idea why these didn't show up... absolutely none. But hey, they're added now.


omsmith:
Might help if you remove `bower_components` from `.gitignore`, they'll show up as new files then.

ryantmer:
Yeah, I `git add --force bower_components` to get all the others in, so no idea how these didn't get added... I'll remove bower_components from it in another PR though, good idea!

PR #508, opened 2017-02-28
Fix jsonpify-imports script to work on all envs

CodeBaboon:
This works (for me, at least) on Windows and WSL, would love for someone to give it a try on Linux and if it works there too lets merge it in. I'd love to get me some bsi.html.js in my dist folder again :)


omsmith:
Travis logs are having issues right now, plus we don't really output what we've built - maybe that's a thing we want? Dunno

Verified in bash on ubuntu using node 4 (which WSL would be equivalent to).

dlockhart:
Yeah I've been thinking about ways we could add in some verification steps for all the assets (not just jsonp guys) to avoid something like what happened Monday

PR #513, opened 2017-03-01
new timings

dlockhart:
Adding some new user timings:
- "server": captures time spent on the server processing the request (fetchStart -> responseStart)
- "download": time spent downloading the response (responseStart -> responseEnd)
- "visible": very similar to "display" (and may replace it) except it excludes the time spent on the server and downloading (responseEnd -> webComponentReady). This helps us isolate client-side performance improvements from variation in server time / download time.

In testing, I found that generally: display = server + download + visible. That to me indicates that there isn't much overlap between those events -- i.e. that the browser isn't doing a lot of processing of the document until it's fully downloaded. That may change for large documents, but it's what I was seeing.

I'm going to adjust the code in the LE that sends these along to Google Analytics such that it includes the new "visible" timing. Eventually we may phase out "display" although it's kind of handy that it includes all 3.

### DIFF
--- a/src/page-loading/page-loading.js
+++ b/src/page-loading/page-loading.js
@@ -6,6 +6,8 @@
 		D2L.Performance.measure('d2l.page.domInteractive', 'fetchStart', 'domInteractive');
 		D2L.Performance.measure('d2l.page.domContentLoadedHandlers', 'domContentLoadedEventStart', 'domContentLoadedEventEnd');
 		D2L.Performance.measure('d2l.page.load', 'fetchStart', 'loadEventStart');
+		D2L.Performance.measure('d2l.page.server', 'requestStart', 'responseStart');
+		D2L.Performance.measure('d2l.page.download', 'responseStart', 'responseEnd');
 	});
 
 })();
--- a/src/page-loading/web-components-ready.js
+++ b/src/page-loading/web-components-ready.js
@@ -8,6 +8,7 @@
 			return;
 		}
 		pageReady = true;
+		D2L.Performance.measure('d2l.page.visible', 'responseEnd');
 		D2L.Performance.measure('d2l.page.display', 'fetchStart');
 	}
 


src/page-loading/page-loading.js
```diff
@@ -6,6 +6,8 @@
 		D2L.Performance.measure('d2l.page.domInteractive', 'fetchStart', 'domInteractive');
 		D2L.Performance.measure('d2l.page.domContentLoadedHandlers', 'domContentLoadedEventStart', 'domContentLoadedEventEnd');
 		D2L.Performance.measure('d2l.page.load', 'fetchStart', 'loadEventStart');
+		D2L.Performance.measure('d2l.page.server', 'fetchStart', 'responseStart');
```
omsmith: I wonder if _requestStart_ would be better? I dunno

dlockhart: requestStart: "the time immediately before the user agent starts requesting the current document from the server, or from relevant application caches or from local resources"

fetchStart: "the time immediately before the user agent starts checking any relevant application caches. Otherwise, it must return the time when the user agent starts fetching the resource"

It's not super clear to me what the difference is. From [here](http://stackoverflow.com/questions/31977140/what-is-an-accurate-measure-for-time-to-first-byte), is seems like `fetchStart` may include TCP and SSL negotiation.

So if that's true... then yeah, "server" time is probably best measured without those negotiations via "requestStart". Should we then use "requestStart" for all the places "fetchStart" is used though?

dbatiste: I've found this Mozilla doc to be helpful (in order): this is the order of the timings: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming

Looks like between `fetchStart` and `requestStart` there is domain look-up, acquiring connection, and possibly some secure connection time.

dbatiste: Is the difference really really small?

dlockhart: On my local machine, `requestStart` happens `3ms` later than `fetchStart` -- but there's not a lot of negotiation going on there. On Spark (SSL), the difference is `13ms`, which is much more significant.

Since TCP/SSL negotiation doesn't have anything to do with "time spent on the server", I'm going to go with Owen's suggestion and use `requestStart` for "server" time. I'll leave the other places that use `fetchStart` ("domInteractive", "load", "display") since they're more about the total time from one point to another.

PR #514, opened 2017-03-02
update to image-banner-overlay lang terms

cgalvind2l:


bower.json
```diff
@@ -13,7 +13,7 @@
     "d2l-menu": "^0.3.1",
     "d2l-more-less": "^2.1.7",
     "d2l-my-courses": "https://github.com/Brightspace/d2l-my-courses-ui.git#0.31.2",
-    "d2l-image-banner-overlay": "https://github.com/Brightspace/d2l-image-banner-overlay.git#0.0.4",
+    "d2l-image-banner-overlay": "https://github.com/Brightspace/d2l-image-banner-overlay.git#v10.6.12",
```
jstefaniuk-d2l: I don't think you need the `v` here, but if it works it works

cgalvind2l: I wasn't really sure because it's a branch rather than a tag.

PR #515, opened 2017-03-03
vendor prefixes aren’t needed as they’re added automatically by autoprefixer

dlockhart:
Confirmed this results in the same output CSS by doing a side-by-side diff.


omsmith:
👍

PR #518, opened 2017-03-03
making the “d2l.page.visible” timing visible on the page if asked for it

dlockhart:
@dbatiste, @omsmith What do you guys think? This would let us not need to keep opening up the console to get at the timing variable values when doing manual performance tests... which will be especially handy in Edge since the console leaks like a sieve.

How it works: you just need to put this in the URL bar and visit any page `&timingdebug=1`. That'll store in session storage that you want to see the "d2l.page.visible" timing variable output to the page for every page visit. To stop showing it, either add `&timingdebug=0` or remove it manually from session storage.

It looks like this, just kind of floating there:

screen shot 2017-03-03 at 6 32 37 pm

### DIFF
--- a/src/page-loading/timing-debug.js
+++ b/src/page-loading/timing-debug.js
@@ -1,0 +1,44 @@
+(function() {
+	'use strict';
+
+	window.addEventListener('D2LPerformanceMeasure', function addTiming(e) {
+
+		if (e.detail.value.name !== 'd2l.page.visible') {
+			return;
+		}
+
+		window.removeEventListener( 'D2LPerformanceMeasure', addTiming );
+
+		var res = /(\?|&)timingdebug=(1|0)/gi.exec(location.search);
+		if ( res !== null && res.length === 3 ) {
+			var timingVal = (res[2] === '0') ? false : true;
+			try {
+				if (timingVal) {
+					window.sessionStorage.setItem('TimingDebug', '1');
+				} else {
+					window.sessionStorage.removeItem('TimingDebug');
+				}
+			} catch (e) {}
+		}
+
+		var timingDebug = false;
+		try {
+			timingDebug = (window.sessionStorage.getItem('TimingDebug') !== null);
+		} catch (e) {}
+
+		if (!timingDebug) {
+			return;
+		}
+
+		requestAnimationFrame(function() {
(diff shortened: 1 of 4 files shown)


src/page-loading/timing-debug.js
```diff
@@ -0,0 +1,42 @@
...
+		var timingDebug = false;
+		try {
+			timingDebug = (window.sessionStorage.getItem('TimingDebug') !== null);
+		} catch (e) {}
+
+		if (!timingDebug) {
+			return;
+		}
+
+		requestAnimationFrame(function() {
```
dlockhart: @omsmith Should I try to use fast-dom here?

src/page-loading/timing-debug.js
```diff
@@ -0,0 +1,42 @@
+(function() {
+	'use strict';
+
+	window.addEventListener('D2LPerformanceMeasure', function(e) {
+
+		if (e.detail.value.name !== 'd2l.page.visible') {
+			return;
+		}
+
```
omsmith: `window.removeEventListener`

src/page-loading/timing-debug.js
omsmith: Seems reasonable - we'll be bringing it in anyway. Currently in an import though...

dbatiste:
Great idea Dave.  This will make it easier to see these timings on mobile devices too.

src/page-loading/timing-debug.js
dbatiste: With scroll-spy, I was thinking of updating so that it will work with fast-dom if present, otherwise it will just work vanilla.

dlockhart: Owen made fast-dom an import, so hopefully you can import it in scroll-spy and I can import it here (and more/less imports it)... and together it'll just get imported once.

dbatiste: Yeah, I was more just not feeling great about introducing a hard dependency on it, within potentially a lot of things.  :(

dlockhart: Fixed

dlockhart: I'll do a separate pull request to add fast-dom since it's going to require some re-jigging of more/less possibly...

PR #520, opened 2017-03-06
fixes #519: explcitly importing fastdom so it can be used in the mono…

### LINKED ISSUE #519: import fast-dom
Currently more/less is importing the fast-dom library, but we'd like to use it in more components and inside Brightspace itself. Let's import it into BSI. Also, update the debug-timing stuff to use it.

dlockhart:
…lith


omsmith:
Used in the monolith, or use in bsi?

dlockhart:
Both? Technically it was able to be used before in the monolith as soon as you added more/less, but this way the usage is explicit. If we kill more/less one day, all of a sudden fastdom would be mysteriously gone.

omsmith:
Just not sure bringing in globals via a WC import is the best way to consume them in the monolith

omsmith:
(like, it's usable as long as you try to use it after WCR, yeah?)

dlockhart:
Ah ok... so you'd rather have BSI proxy with a facade or something? I can do that...

omsmith:
or include it via the .js somehow or something. this seems fine for chrome, but anything without native imports the timing is a little weirder no?

dlockhart:
I've made some major changes based on @omsmith's comments. The problem with just importing it into `bsi.html` is that for browsers who don't natively support web components, it won't be available until after the WCR event. If instead I put it into the JavaScript bundle (`bsi.js`), it won't be available to web components since that JS file is included much lower.

Instead, I've wrapped the web-component version so web components can just use the import, but everyone else can go through `D2L.FastDom`, which will call directly into `fastdom` if it's available, or wait for the WCR event and then call into.

src/d2l-fastdom.js
```diff
@@ -0,0 +1,47 @@
...
+var idMap = {};
+
+function helper(method, cb) {
+	if (window['fastdom']) {
+		return window['fastdom'][method](cb);
+	}
+	index++;
+	var fastDomId = 'd2l_fastdom_' + index;
+	idMap[fastDomId] = 'pending';
+	window.addEventListener('WebComponentsReady', function wcr() {
```
omsmith: would it make sense to have a single listener, which just dumps a queue of actions instead of one per action?

dlockhart: Good idea, will add.

dlockhart: @omsmith alright, adjusting things to just use the one event listener and a queue

src/d2l-fastdom.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function helper(method, cb) {
+	if (window['fastdom']) {
+		return window['fastdom'][method](cb);
+	}
+	index++;
+	var fastDomId = 'd2l_fastdom_' + index;
+	idMap[fastDomId] = {method: method, id: 'pending', cb: cb};
+	if (!listenerAdded) {
```
omsmith: This technically shouldn't be needed. Registering the same listener for the same even is a noop, but this might be faster...?

src/d2l-fastdom.js
```diff
@@ -0,0 +1,67 @@
...
+
+var index = 0;
+var idMap = {};
+var listenerAdded = false;
+
+function wcr() {
+
+	window.removeEventListener('WebComponentsReady', wcr);
+	if (!window['fastdom']) {
+		return;
```
omsmith: This part kinda sucks

dlockhart: Haha yeah, good call. I'll just do a rAF.

src/d2l-fastdom.js
dlockhart: I like the simplicity of registering each time and having the duplicates be discarded -- will change, thanks.

src/d2l-fastdom.js
```diff
@@ -0,0 +1,67 @@
...
+		var item = idMap[id];
+		if (item.id === 'pending') {
+			item.id = window['fastdom'][item.method](item.cb);
+		}
+	}
+
+}
+
+function helper(method, cb) {
+	if (window['fastdom']) {
```
omsmith: what happens after wcr, but fastdom doesn't exist (given we're checking/expecting that elsewhere)?

PR #524, opened 2017-03-07
Update autoprofix config to reflect current browser support.

dbatiste:
I _think_ this is what we want.

To give an idea of the raw size change:
```
bsi.css          111KB -> 97KB (-12.6%)
bsi.min.css      89KB -> 79KB (-11.2%)
```

### DIFF
--- a/postcss.config.json
+++ b/postcss.config.json
@@ -3,6 +3,6 @@
 	"input": "./dist/bsi.css",
 	"dir": "./dist/",
 	"autoprefixer": {
-		"browsers": "Last 1 versions, Firefox ESR, IE >= 11"
+		"browsers": "Last 3 versions, Firefox ESR, IE >= 11, Safari >= 9"
 	}
 }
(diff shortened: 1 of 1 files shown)


postcss.config.json
```diff
@@ -3,6 +3,6 @@
 	"input": "./dist/bsi.css",
 	"dir": "./dist/",
 	"autoprefixer": {
-		"browsers": "Last 4 versions, Firefox ESR, IE >= 9, Safari >= 6"
+		"browsers": "Last 1 versions, Firefox ESR, IE >= 11"
```
dlockhart: The only worry would be Slightly older versions of Safari (9 maybe?). But the savings are significant.

dbatiste: Yeah.  So, in terms of size, we can get most of the benefit by just going `Last 3 versions, Firefox ESR, IE >= 11, Safari >= 9`...

```
bsi.css          111KB -> 99KB
bsi.min.css      89KB -> 81KB
```

That would be a safer change, and get us most of the benefit.

dbatiste: Of course, compression makes these differences less significant, but it gives an idea.

dlockhart: For sure. I'd feel more comfortable with that.

PR #525, opened 2017-03-07
moving HTML block styles into sass

dlockhart:
Now that all the typography styles as SASS, this htmlblock "web-component" isn't needed anymore. I'll be doing the same with a few others in subsequent PRs (e.g. `htmleditor.html`).

Likely some minor performance gains achieved here due to the removal of a style include.

@njostonehouse FYI since I think you were original involved in creating these.


dlockhart:
I'm actually not sure that the `d2l-typography` class prefix is needed anymore... it may have been needed because of how the style include got applied, but shouldn't anymore with straight CSS.

src/html-block.scss
```diff
@@ -0,0 +1,130 @@
...
+	strong,
+	b *,
+	strong * {
+		font-weight: bold;
+	}
+
+	h1 {
+		font-size: 2em;
+		line-height: 37px;
+		margin: 21.43px 0;
```
dbatiste: I know these existed previously.  The partial pixels is pretty weird.

dlockhart: Yeah I know, they're pretty weird... probably worth looking at these again at some point.

PR #527, opened 2017-03-07
moving html editor styles to sass and merging with monolith styles

dlockhart:
@dbatiste, @njostonehouse Similar (but more complex) to the HTML block stuff. Goes along with [this PR](https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/4985/overview) in Brightspace, this moves the styles into Sass and merges them with what was already defined in Brightspace.

It's crazy complicated to follow every rule, especially with the addition of the monolith CSS, so don't worry too much about that.


dlockhart:
I've updated this (and the PR in Brightspace) to also include the editor components (button, color-button, groups and menu) that were getting Daylight-styled.

PR #538, opened 2017-03-20
remove my-courses imports

dlockhart:
Removing alert, course-image, image-selector, organization-hm-behavior and simple-overlay from BSI’s import since my-courses already includes them.

This ends up reducing `shared.html` by ~70KB and increasing `d2l-my-courses.html` by the same amount.


omsmith:
These are used by course banner.

@CodeBaboon

CodeBaboon:
Yeah, it does, good catch Owen. I think other teams have been planning to make use of alerts as well (thought maybe not in the context of BSI, I'm unsure).

omsmith:
Though is that just "d2l-image-banner-overlay"? I would've expected it to all stay in shared.html if that's the case?

CodeBaboon:
Hmm, so looking closer, seems we never got to actually implementing the image selection (which is where some of those repos were intended to be re-used). I'm not sure if there are still plans to go down that path or if the feature is stagnant at this point. `Alert` and the organization-hm-behavior' are definitely used in the banner, though. But the rest could maybe get placed back into my-courses for now if you want.

dlockhart:
It's OK that they're used in `d2l-image-banner-overlay`, this won't break that but it just means it'll end up in `shared.html` if it's used in 2 or more places instead of `bsi.html`. Things should only be listed in `bsi.html` if the monolith is expecting them to exist. I can't find any usage of any of these web-components in the monolith itself. If that changes in the future, we can add them back.

dlockhart:
Each web-component that we include in `bsi.html` (or `shared.html` for that matter) costs us in page start-up performance, and it's not insignificant in the slower browsers. So if things can be moved to their own imports (like `d2l-my-courses.html` or `d2l-image-banner-overlay.html` have done), converted to not be a web component or removed entirely, that an important gain.

dlockhart:
The things that moved from shared -> my-courses were: course-image, image-selector and simple-overlay. That's where the 70KB swap came from. organization-hm-behavior and d2l-alert were both used in 2 places (d2l-my-courses and d2l-image-banner-overlay) so they stayed in `shared.html` -- but they still shouldn't have been listed in `bsi.html` because the monolith doesn't use them.

CodeBaboon:
Thanks for the awesome explanation, @dlockhart !

PR #544, opened 2017-03-21
switching WC build to use polymer-build instead of web-component-shards

dlockhart:
@omsmith @ryantmer @njostonehouse @capajon @edwardg8

This is a pretty major change in how BSI's web components get built. I've switched it from the seemingly poorly supported [web component shards](https://github.com/PolymerLabs/web-component-shards) to [polymer-build](https://github.com/Polymer/polymer-build), which is part of Polymer's new CLI.

This has a few advantages:
* Being part of the CLI, it should be better maintained
* web-component shards required you to be stuck on node version 5.x or else it blew up, which [we logged in December](https://github.com/PolymerLabs/web-component-shards/issues/20).
* Any errors during the build with web-component-shards wouldn't result in a non-zero error code, also [something we logged](https://github.com/PolymerLabs/web-component-shards/issues/16)
* This could set us up to use some of the other bits of Polymer CLI, like linting, minifying, different bundles, etc.
* The JSON-based config should make it easier to break out more shards (polymer-build calls them fragments) in the future -> better maintainability

I've diff'd the pre-minified outputs of the new and old ways and things look pretty good. The ordering of the shared assets is slightly different, but the file sizes are within 100 bytes of each other. Also tested this across all our major browsers.


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: node_js
-node_js: "5"
+node_js: stable
```
dlockhart: Party time

bower.json
```diff
@@ -12,7 +12,7 @@
     "d2l-loading-spinner": "^5.0.3",
     "d2l-menu": "^0.3.5",
     "d2l-more-less": "^2.2.0",
-    "d2l-my-courses": "https://github.com/Brightspace/d2l-my-courses-ui.git#1.2.12",
+    "d2l-my-courses": "https://github.com/Brightspace/d2l-my-courses-ui.git#1.2.13",
```
dlockhart: polymer-build must have slightly better linting, as it caught an error in my-courses which I fixed in version 1.2.13.

polymer-build.js
```diff
@@ -0,0 +1,18 @@
+const vfs = require('vinyl-fs');
+const mergeStream = require('merge-stream');
+const PolymerProject = require('polymer-build').PolymerProject;
+
+const project = new PolymerProject({
+	"entrypoint": "web-components/index.html",
```
omsmith: heh, looked at this before - hadn't considered giving it an empty file :P

package.json
```diff
@@ -13,8 +13,8 @@
     "postbuild:sass": "npm run autoprefix && npm run uglify:css",
     "build:icons": "cpy --parents --cwd=bower_components/d2l-icons \"images/**/*.*\" ../../dist",
     "prebuild:wc": "rimraf tmp && rimraf tmp-wc",
-    "build:wc": "web-component-shards -d tmp-wc/ --endpoints bsi.html d2l-my-courses.html d2l-navigation.html d2l-image-banner-overlay.html",
-    "postbuild:wc": "npm run uglify:html && npm run jsonpify-imports",
+    "build:wc": "node polymer-build.js",
+    "postbuild:wc": "rimraf tmp-wc/bower_components && rimraf tmp-wc/js && rimraf tmp-wc/web-components && npm run uglify:html",
```
dlockhart: Unfortunately `polymer-build` copies `bower_components` and any other directories you reference (like `js` and `web-components`) into the output directory. They're not needed, so I'm deleting them here to prevent them from making it to the next stage of the build.

polymer-build.js
dlockhart: Yeah... a little sad to do it.

dlockhart:
Thanks @CodeBaboon!

PR #552, opened 2017-04-19
Adding max left and right page padding

capajon:
We noticed in email, after setting it to full width, that the page padding gets a little too generous at very wide viewports.  I believe this should be a pretty safe change.

### DIFF
--- a/sass/page-two-panel-selector.scss
+++ b/sass/page-two-panel-selector.scss
@@ -1,4 +1,5 @@
 @import '../bower_components/d2l-colors/d2l-colors.scss';
+@import 'common.scss';
 
 $d2l-twopanelselector-width: 320px;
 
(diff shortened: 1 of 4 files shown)


dlockhart:
This is duplicated in the navigation code here (with a responsive rule down below):
https://github.com/Brightspace/brightspace-integration/blob/master/sass/navigation/common.scss#L18

capajon:
Okay I'll make a mixin for it.

We don't seem to have a common scss file between navigation and page level scss.  Do you prefer I create a new common file, or in this case define the new mixin in page.scss and import page.scss into navigation/common.scss?

dlockhart:
I don't really have a preference, up to you!

capajon:
@dlockhart do you mind sanity checking the refactor? I'll do the BSI release after.  Thanks!

sass/page-two-panel-selector.scss
```diff
@@ -1,4 +1,5 @@
 @import '../bower_components/d2l-colors/d2l-colors.scss';
+@import 'common.scss';
```
dlockhart: Could the usage here have just been replaced with a call to the mixin? Or did you not want the max-breakpoint in this case?

capajon: Yeah exactly, and actually I only want the relative % to kick in at < 1000px (which is when two panel selector starts stacking and not being in a column layout).  There were ways to do that with a mixin of course but I thought it would be confusing later on... so I just went with the shared vars at least.

dlockhart: Cool makes sense!

PR #553, opened 2017-04-20
reintroducing d2l-link class

dlockhart:
I'll be switching the LMS over to use the "d2l-link" CSS class instead of the web component, as the performance gains are too significant right now to ignore.

This change pulls in the latest version of `d2l-link` -- which has Sass mixins now -- and applies the mixins to the `.d2l-link` CSS class and `.vui-link` class (for backwards compatibility with old FRAs). It also applies it inside HTML blocks for user-authored content.


dlockhart:
@dbatiste I'm going to merge this so that I can create the LP PR and get CI going, but if you see any problems I can still fix them.

PR #559, opened 2017-04-25
moving button-menu and button-split CSS from the WC into Sass

dlockhart:
@dbatiste FYI. This is just a performance improvement we can make now that button's CSS isn't in a web component anymore. As we discussed, I'm still interested in converting button-menu to straight-up use button's style, but that will be a future enhancement.


sass/button-menu.scss
```diff
@@ -1,7 +1,75 @@
...
+@import 'common.scss';
+
 .bsi-button-menu > span {
 	@include vui-dropdown-button-menu-content;
 }
 
 .bsi-button-menu.bsi-button-menu-primary > span {
 	@include vui-dropdown-button-menu-primary-content;
 }
+
```
dlockhart: All of this CSS is just moved mostly as-is from `button-menu.html`

sass/button-split.scss
```diff
@@ -0,0 +1,39 @@
+@import '../bower_components/d2l-colors/d2l-colors.scss';
```
dlockhart: This is all moved as-is from`button-split.html`

PR #560, opened 2017-04-25
moving collapse-pane CSS from the WC into Sass

dlockhart:


web-components/collapse-pane.html
```diff
@@ -1,26 +0,0 @@
...
-				text-decoration: none;
-			}
-
-			.d2l-collapsepane-action:hover > .d2l-heading,
-			.d2l-collapsepane-action:focus > .d2l-heading {
-				color: var(--d2l-color-celestuba);
-				text-decoration: underline;
-			}
-
-			.d2l-typography .d2l-collapsepane-header .d2l-heading {
```
dlockhart: This rule didn't make it over to the Sass, because this CSS is already present in the monolith.

PR #566, opened 2017-05-09
Force the latest d2l-authify

CodeBaboon:
1.0.1 breaks in Safari < 10.1 and IE, so lets make sure we get 1.0.2 at a minimum


dlockhart:
It'll grab it automatically, but worth upgrading d2l-image-banner-overlay to `1.0.2` as well?

CodeBaboon:
I think that's what this does

dlockhart:
Haha wow, sorry. I got this mixed up with another diff I was looking at. Carry on...

CodeBaboon:
1.0.2 is busted, too. Closing.

PR #568, opened 2017-05-09
Move to latest banner overlay release

CodeBaboon:
1.0.1 breaks in Safari < 10.1 and IE, and 1.0.2 breaks in everything so lets make sure we get 1.0.3 at a minimum


jstefaniuk-d2l:
Tested!

PR #585, opened 2017-06-20
Fix for dialogs scrolling out of view when inputs are used on ios

capajon:
Will add comments to related LMS PR.


capajon:
I'd like to go ahead and merge this so I can get a BSI release going (any outstanding work for the story that I am aware of should be in the LMS and not end up changing the styles).

PR #591, opened 2017-06-28
Moving homepages style from https://github.com/Brightspace/homepages-ui

capajon:
Moving styles from https://github.com/Brightspace/homepages-ui
(created a PR to updated README at least: https://github.com/Brightspace/homepages-ui/pull/1)

I toyed around with not using vui-grid-system (which, maybe:
- Just copy output CSS directly, or
- Include config for susy directly in this repo

But I wanted to keep things mostly the same for step 1.  I am also not sure where our grid system stands.  We could look to clean this up further as part of the hygiene sprint?


sass/homepages/homepages.scss
```diff
@@ -0,0 +1,62 @@
...
+		#{$className} {
+			width: span($i);
+		}
+	}
+}
+
+$columns: susy-get(columns);
+$col-prefix: 'homepage-col-';
+$col-classes: gen-col-list($columns, $col-prefix);
+$gutter-width: 5.1%;
```
capajon: This line is the only change that was ultimately required.  The design calls for 60px at the full width breakpoint.  I thought about updated vui-grid-system instead but I noticed it has been downloaded a bit in the past month, and I'm not sure we would want this design change to apply anywhere outside of homepages.  Considering I think this should be further cleaned up I just went with the simplest approach for now.

PR #596, opened 2017-07-06
added css of text-based nav editMenu

edwardg8:


edwardg8:
Will push LMS changes later, had to decline the PR due to rebase

sass/navigation/main.scss
```diff
@@ -85,3 +85,24 @@
 		display: none;
 	}
 }
+
+.d2l-navigation-s-gutters {
+	position: relative;
+}
+
+.d2l-navigation-ib-edit-menu {
```
dbatiste: I think we should rename these classes if they are going to apply in both the normal and image-based scenarios.  I think the class names should follow the others here.

edwardg8: i'll name it `d2l-navigation-s-edit-menu` then

PR #597, opened 2017-07-06
US88109 Ensure close button is always visible and hover is active on …

mdulat:
…entire area of button.

### DIFF
--- a/sass/dialog.scss
+++ b/sass/dialog.scss
@@ -248,7 +248,7 @@ $dialog-resize: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.
 	top: 0;
 	height: calc(1.3rem + 2px);
 	width: calc(1.3rem + 2px);
-	z-index: 9999;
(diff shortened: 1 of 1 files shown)


dbatiste:
Hey Margaret, I'm unable to locate `US88109` to get more info.  Any chance you can provide some background on what was happening here, and why the z-index value is needed, and whether a smaller z-index value can be used?

mdulat:
Sure. In the QuickLinks picker, the close (x) button is hidden behind other header elements. See screenshot attached. For types that don’t include the header elements, the close X is visible, but only clickable in the small area on the top and right of the element (the part that is visible in screenshot), and hover styling only appears in this small area as well.

It could probably be a smaller z-index value, but I’m not sure what value would be safe to use. Maybe 1 would suffice.

Margaret

dbatiste:
Ok ya - I would suggest using `z-index: 1;` if possible.

mdulat:
I guess I don't have permissions to merge. Can you do it please?

PR #598, opened 2017-07-07
Dbatiste/image based nav mobile link groups

dbatiste:


sass/dialog.scss
```diff
@@ -22,25 +22,26 @@ $dialog-resize: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.
 		top: -1000px;
 	}
 
-	@keyframes dialog-slide-in-from-top-open {
```
dbatiste: It was necessary to move the keyframe definitions outside of the mixin in order to keep them outside of the media-queries.  This is necessary because in IE, the keyframe animations will not run if the keyframes are not defined outside the media-query.  Dialogs rely on `animationend` to remove classes, causing problems dismissing dialogs in mobile view in IE.

sass/dialog.scss
```diff
@@ -298,6 +305,10 @@ html.d2l-dialog-open, html.d2l-dialog-open body {
 		position: fixed !important;
 	}
 
+	.d2l-dialog-inline {
+		max-width: 100%;
```
dbatiste: The `width: 100% !important;` was not enough to override the `max-width: 60vw;` in Safari.

sass/navigation/mobile-image-based.scss
```diff
@@ -12,3 +12,8 @@
 	border: 1px solid $d2l-color-gypsum;
 	line-height: 0;
 }
+
+.d2l-navigation-mobile-ib-item-group-content {
+	justify-content: space-between;
+	max-width: 100%;
```
dbatiste: The desktop inline dialog content limits the width so that the user will see 3 tiles across... same as what they see in the rendered opener icon in the tray.  In mobile view, we just want to make best use of the space available (i.e. show as many tiles across as possible).

PR #599, opened 2017-07-10
Lock bower components versions

ISkomorokh:
> Dave Lockhart [5:22 PM] 
One other thing to note — for the branch of BSI, `bower_components` is checked in, which is nice because it’ll let you see what changes happen to any dependency. What I’d also suggest before making any changes is to lock the version of every dependency in `bower.json` by removing the `^` in front. That way they won’t auto-upgrade beyond what version is explicitly set there.

>[5:23] 
Then you can make explicit choices based on what defects you’re trying to fix about when to actually upgrade one of those dependencies. `^1.0.0` means it’ll auto-upgrade to `1.10` or `1.99999` all the way up to but not including `2.0`.


ISkomorokh:
@dlockhart please review

ISkomorokh:
Ok, so I did that. Unfortunately reality differs from ideal theoretical scenario :).
When I do `bower install/update` I get this:
image

Anyway, changes in `bower_components` are minor (most of the time it is version spec in bower files). Please, see my latest commit to see the difference.

ISkomorokh:
Also, for some reason, I don't see [these changes](https://github.com/Brightspace/brightspace-integration/commit/b642c07d6e7b8c1b9d9bc013526bf236fa35aa55) in our branch.

bower.json
```diff
@@ -2,38 +2,38 @@
...
+    "d2l-colors": "2.2.3",
+    "d2l-dropdown": "5.0.2",
+    "d2l-icons": "3.0.0",
+    "d2l-image-action": "0.7.0",
+    "d2l-link": "3.3.1",
+    "d2l-loading-spinner": "5.0.3",
+    "d2l-menu": "0.3.2",
+    "d2l-more-less": "2.1.7",
     "d2l-my-courses": "https://github.com/Brightspace/d2l-my-courses-ui.git#0.31.2",
-    "d2l-image-banner-overlay": "https://github.com/Brightspace/d2l-image-banner-overlay.git#v10.6.12",
```
ISkomorokh: I'm not sure this is correct

dlockhart: Yeah that's interesting eh... `10.6.12` isn't a valid release for it. I think you're right though, `0.0.4` seems about right for around February, and that matches what's in the checked in `bower_components`. @CodeBaboon does this look OK? We're trying to determine what version of `image-banner-overlay` is actually checked in here for the on-prem release.

dlockhart:
Haha yeah, so as those version conflicts are going to be because the dependencies themselves are still using `^` versioning. I'd suggest choosing the version of each of them that's checked in here (e.g. colors version `2.2.3`) and then persisting it to `bower.json` with `!` so that others won't see it again.

Luckily, Travis CI won't have an issue since it just uses everything from `bower_components` -- it no longer does a `bower install`.

dlockhart:
As for the missing commits -- good catch. I bet what happened was that this branch was created right around there (Feb 9 or so), which means any commits from Feb 10 onward we should look at to see if they belong in here. But we can start doing that after this PR merges.

bower.json
CodeBaboon: I was away on vacation and missed this, but `v10.6.12` is [a branch](https://github.com/Brightspace/d2l-image-banner-overlay/tree/v10.6.12) not a release. I don't believe `0.0.4` will work properly with the on-prem LMS codebase, at least at the time it was originally branched into 10.6.12. This change concerns me.

CodeBaboon: If nothing else it will be missing the translations added here: https://github.com/Brightspace/d2l-image-banner-overlay/commit/332ed0c5d3569abaf4a28d97508f9e208fceae97

CodeBaboon: @dlockhart  It appears that `332ed0c5d3569abaf4a28d97508f9e208fceae97` is the only missing commit but still will be an issue: https://github.com/Brightspace/d2l-image-banner-overlay/compare/0.0.4...v10.6.12

dlockhart: @ISkomorokh @CodeBaboon so this PR got closed in [favour of this one](https://github.com/Brightspace/brightspace-integration/pull/600/files), but that still had the change from `v10.6.12` branch (I see that now, thanks) to `0.0.4`.

Looks like we should be either switching back to pointing at the 10.6.12 branch, or picking a version higher than 0.0.4 that will work with the on-prem release and has the translations. I'd like to avoid creating on-prem versions of _every_ shared component if we can... that will get so out of control so quickly.

ISkomorokh: @CodeBaboon Thanks. 
> If nothing else it will be missing the translations added here: Brightspace/d2l-image-banner-overlay@332ed0c
Are you sure about that? I've just checked and those translations are present in `\bower_components\d2l-image-banner-overlay\src\localize-behavior.html`

CodeBaboon: @ISkomorokh That's interesting, I'm not sure how that would happen if `0.0.4` does not include that commit. Is there a chance that the `\bower_components\` folder you are looking in was not cleared out or pulled from a cached copy of the `v10.6.12` branch or something?

ISkomorokh: Well, that folder was checked in during branching of 10.6.12 (onprem). It wasn't changed since than.

CodeBaboon: Ah that explains it. So the reference in `bower.json` is still technically incorrect but it isn't a deployed issue so long as the files checked in under `/bower_components/` remain at the original `v10.6.12` point.

PR #600, opened 2017-07-10
Switched from shards to polymer

ISkomorokh:


ISkomorokh:
@dlockhart please review.

NOTE: not sure how to verify it properly
image

dlockhart:
To verify, follow the instructions in the README for [running a local version of BSI](https://github.com/Brightspace/brightspace-integration#running-locally) against a dev instance of Brightspace. I can walk you through it offline if you run into trouble.

ISkomorokh:
Ok, I've tried to verify the result against my dev env and see a very strange result. With daylight off the Login page looks like this 
image

dlockhart:
Hmm, lots of things look wrong with that login page.

ISkomorokh:
In 10.6.12 `bsi.css` is downloaded from https://s.brightspace.com/lib/bsi/10.6.11-2/ and it is totally differ from the one that is produced by the build. Any idea what could be done to solve this?

njostonehouse:
With Daylight off, the value in `_config/Infrastructure/D2L.LP.Web.UI.Html.Bsi.config.json` at `baseLocation` will be used for BSI. I believe that's the version referenced in 10.6.12. When Daylight is enabled, you'll get baseLocation-c12i12n (built from the master branch). If you're wanting to update pre-Daylight BSI, you'll need to build from the https://github.com/Brightspace/brightspace-integration/tree/daylight-off branch. Does that help?

ISkomorokh:
@njostonehouse thanks a lot! This explains everything. Anyway, I need to check that. I'll be back ... :)

bower_components/d2l-page-load-progress/d2l-page-load-progress.html
```diff
@@ -4,12 +4,12 @@
 <dom-module id="d2l-page-load-progress">
 	<style>
 		:host {
-			background-color: var(--d2l-color-woolonardo);
+			background-color: var(--d2l-color-sylvite);
```
dlockhart: This change (and the one below) worries me a little. Woolonardo was renamed to Sylvite and Pressicus to Corundum in colors version `2.4.0`, but version `2.2.3` is included here. So I think those variables won't exist yet.

I think `d2l-page-load-progress` needs to stay at version `1.2.2` (before the color change)... it's used by "navigation", which although an older version is included here, I think is causing the newer page-load-progress to be included. Maybe include a resolution entry for `d2l-page-load-progress` to `1.2.2` in the `bower.json` to fix?

ISkomorokh: Well, this is what `bower update` prints out:
image

ISkomorokh:
Now everything builds fine and I see no errors when working with my local instance. The only thing that bothers me is the fact that there are lots of diffs between http://s.brightspace.com/lib/bsi/10.6.11-2/bsi.css and local `dist\bsi.css`. I was expecting minor amount of diffs :(

This PR contains all the changes of #599 . Please, review. If everything is fine, I'll merge this one and will decline #599 
It would be great if someone from D2L side could check this branch out and verify that we are on the right track.

bower_components/d2l-page-load-progress/d2l-page-load-progress.html
ISkomorokh: I've even tried to lock versions once again from the scratch using:

```json
 "resolutions": {
    "polymer": "1.7.1",
    "d2l-colors": "2.2.3",
    "Stickyfill": "d1fbf3c87b",
    "page-load-progress": "1.2.2",
    "app-localize-behavior": "0.10.1",
    "d2l-image-action": "0.7.0",
    "d2l-typography": "5.2.4",
    "d2l-polymer-behaviors": "0.0.4"
  }
```
But anyway I'm getting the same changes in this file :( 
I'm afraid that this lock (PR) might introduce unexpected issues.

njostonehouse:
I would expect the differences between what's built from this branch and what's built at v10.6.11-2 to be vastly different. I don't think it's anything to worry about. The two bsi.css packages are mutually exclusive, and represent styles that are deliberately distinct (Daylight On and Daylight Off). The only issue would be if you were intending to port a change to **pre**-Daylight that required an update to **its** BSI, then this change would not accomplish that. But this branch doesn't have a way to deal with pre-Daylight code, so it's not a concern in this context.

ISkomorokh:
@njostonehouse ohhhh... that makes sense. I should have compared against https://s.brightspace.com/lib/bsi/10.6.12-daylight.5/bsi.css :) The same mistake as with `baseLocation` vs `baseLocation-c12i12n`
Now I see only 3 minor differences! This is great!

PR #601, opened 2017-07-11
US88463-reposition of editnavbar

edwardg8:


sass/navigation/main.scss
```diff
@@ -11,6 +12,7 @@
 	display: flex;
 	height: calc(1rem + 40px);
 	flex-wrap: nowrap;
+	margin-right: 45px;
```
dlockhart: What's this for? Could you provide some reasoning behind the change or some before/after screenshots?

edwardg8: We want to re-position the edit navbar menu for text-based nav (details refer to `US88463` - has screenshot of the changes). Earlier we have the edit menu above more menu, now we want to move it to the right side of more menu. The 45px is reserving the space for this edit navbar menu (35px width of menu +10px width space between the last link and the edit menu

edwardg8: Before:
image

After:
image

dlockhart: Hmm. So for people who don't have permission to see the menu (all students), there'll be an empty space to the right. I'd recommend putting a classname somewhere when it renders to indicate that the menu is going to be present, and then only apply the margin when that class is there.

edwardg8: Thank you for catching that, make the change right now.

edwardg8: I made changes from LMS, linked the BSI changes here to LMS PR

sass/navigation/main.scss
```diff
@@ -23,6 +24,10 @@
 	}
 }
 
+.d2l-navigation-s-main-wrapper[has-edit-menu] {
+	margin-right: 45px;
```
dlockhart: What about RTL?

edwardg8: Wow, the eidtmenu for both image-based or text-based, there isn't RTL considered. Thanks for pointing it out. I'm going to test and address all RTL problems here

PR #602, opened 2017-07-11
US87960 Adjusting navbar breakpoints and related spacing

capajon:
There are a few fixes besides just adjusting breakpoints.  Also, I changed how margins work on most items so that separators (the triple dot vertical spacers) can be removed but spacing between icons can be maintained.  

Also, there are a couple of changes I'd like to make to the rendering to remove a couple of items I don't think are necessary now (example a couple of div.d2l-navigation-s-gutter).  I am noting that for the hygiene sprint.  It would be difficult to do without safely in a coordinated LMS/BSI commit anyway, so this is a good first step regardless.

### DIFF
--- a/sass/navigation/course-menu.scss
+++ b/sass/navigation/course-menu.scss
@@ -1,9 +1,14 @@
+@import 'common.scss';
+
 .d2l-navigation-s-course-menu {
 	height: 100%;
-}
-
-.d2l-navigation-s-course-menu .d2l-navigation-s-button-highlight {
-	margin: 0 5px;
+	margin-right: $d2l-navigation-margin-regular;
+	@media (max-width: $d2l-navigation-bp-mobile-menu) {
+		display: none;
+	}
+	@media (max-width: $d2l-navigation-bp-separators) {
+		margin-right: $d2l-navigation-margin-compact;
+	}
 }
 
 .d2l-navigation-s-course-menu d2l-dropdown {
@@ -14,9 +19,6 @@
 	line-height: 1.5rem;
 }
 
-@media (max-width: 554px) {
-	.d2l-navigation-s-course-menu,
-	.d2l-navigation-s-divider.d2l-navigation-s-course-menu-divider {
-		display: none;
-	}
+.d2l-navigation-s-divider.d2l-navigation-s-course-menu-divider {
+	margin-left: -16px;
 }
--- a/sass/navigation/divider.scss
+++ b/sass/navigation/divider.scss
@@ -1,11 +1,15 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
+@import 'common.scss';
 
 .d2l-navigation-s-divider {
 	display: inline-block;
 	flex: 0 0 auto;
-	margin: 0 9px;
+	margin: 0 9px 0 -21px;
 	d2l-icon {
 		color: $d2l-color-mica;
+	}
+	@media (max-width: $d2l-navigation-bp-separators) {
+		display: none;
 	}
 }
 .d2l-navigation-s-divider-no-margin {
--- a/sass/navigation/header.scss
+++ b/sass/navigation/header.scss
@@ -38,14 +38,9 @@
 	padding: 0 7px;
 
 	.d2l-navigation-s-header-container & {
-		&.d2l-navigation-s-header-no-home-icon .d2l-navigation-s-logo-divider {
-			@media (min-width: $d2l-navigation-bp-mobile-menu) {
-				display: none;
-			}
-		}
 		& .d2l-navigation-s-home-icon,
-		.d2l-navigation-s-has-title & .d2l-navigation-s-logo-divider,
-		.d2l-navigation-s-has-title & .d2l-navigation-s-logo {
+		& .d2l-navigation-s-logo-divider,
+		& .d2l-navigation-s-logo {
 			@media (max-width: $d2l-navigation-bp-mobile-menu) {
 				display: none;
 			}
--- a/sass/navigation/header.scss
+++ b/sass/navigation/header.scss
@@ -1,4 +1,5 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
+@import 'common.scss';
 
 .d2l-navigation-s-header {
 	border-bottom: 1px solid $d2l-color-gypsum;
@@ -21,6 +22,7 @@
 	display: inline-block;
 	flex: 0 1 auto;
 	height: 100%;
+	margin-right: $d2l-navigation-margin-compact;
 	@media (min-width: $d2l-navigation-bp-mobile-menu) {
 		display: none;
 	}
@@ -36,9 +38,6 @@
 	padding: 0 7px;
 
 	.d2l-navigation-s-header-container & {
-		@media (max-width: $d2l-navigation-bp-title) {
-			display: none;
-		}
 		&.d2l-navigation-s-header-no-home-icon .d2l-navigation-s-logo-divider {
 			@media (min-width: $d2l-navigation-bp-mobile-menu) {
 				display: none;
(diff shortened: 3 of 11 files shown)


sass/navigation/course-menu.scss
```diff
@@ -14,9 +19,6 @@
 	line-height: 1.5rem;
 }
 
-@media (max-width: 554px) {
-	.d2l-navigation-s-course-menu,
-	.d2l-navigation-s-divider.d2l-navigation-s-course-menu-divider {
-		display: none;
-	}
+.d2l-navigation-s-divider.d2l-navigation-s-course-menu-divider {
+	margin-left: -16px;
```
capajon: I don't love using a negative margin here but it keeps the design exactly as it was before and the margins are much easier to work with as separators are hidden.  Still, would like to revisit this in hygiene sprint or something like that.

dlockhart: Sorry, I don't quite get what this is trying to accomplish? Previously it looks like at a certain breakpoint, it was hiding the course menu and its divider. Now it never gets hidden, but its divider is moved over a bit?

dlockhart: I see now from the rule down below that all dividers are hidden at a particular breakpoint. Still not sure what the `-16px` is for though -- could you post a screenshot?

sass/navigation/divider.scss
```diff
@@ -1,12 +1,16 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
+@import 'common.scss';
 
 .d2l-navigation-s-divider {
 	display: inline-block;
 	flex: 0 0 auto;
-	margin: 0 9px;
+	margin: 0 9px 0 -21px;
```
dlockhart: Applying this `-21px` left-margin manually to a QA site completely hides the divider. Not sure if that was the intention?

sass/navigation/header.scss
```diff
@@ -21,6 +22,7 @@
 	display: inline-block;
 	flex: 0 1 auto;
 	height: 100%;
+	margin-right: $d2l-navigation-margin-compact;
```
dlockhart: Any right or left-specific margins/padding will need to be inverted for RTL.

sass/navigation/header.scss
```diff
@@ -36,24 +38,20 @@
...
-		&.d2l-navigation-s-header-no-home-icon .d2l-navigation-s-logo-divider {
-			@media (min-width: $d2l-navigation-bp-mobile-menu) {
-				display: none;
-			}
-		}
 		& .d2l-navigation-s-home-icon,
-		.d2l-navigation-s-has-title & .d2l-navigation-s-logo-divider,
-		.d2l-navigation-s-has-title & .d2l-navigation-s-logo {
+		& .d2l-navigation-s-logo-divider,
+		& .d2l-navigation-s-logo {
```
dlockhart: Some logic is changing here I guess? Always hiding the logo and its divider, even where there's no title?

sass/navigation/logo.scss
```diff
@@ -4,8 +4,15 @@
 .d2l-navigation-s-logo {
 	align-items: center;
 	display: inline-flex;
+	flex-shrink: 0;
+	-webkit-flex-shrink: 0;
+	-moz-flex-shrink: 0;
```
dlockhart: According to [AutoPrifixer](https://autoprefixer.github.io/), you don't need `-webkit-` or `-moz` anymore, but might need `-ms` for IE11. Also, I've read that you should use the full 'flex' instead of `flex-shrink` to be really explicit about the other 2 values.

sass/navigation/logo.scss
```diff
@@ -4,8 +4,15 @@
...
 	height: 100%;
+	margin-right: $d2l-navigation-margin-regular;
```
dlockhart: Needs to be inverted for RTL.

sass/navigation/mobile-menu.scss
```diff
@@ -182,7 +183,23 @@
 	}
 
 	.d2l-navigation-s-link {
-		font-size: 1rem;
+		font-size: 0.9rem;
```
dlockhart: I _think_ that just removing `font-size` will have the same effect? I believe the default font size is `0.9rem`.

sass/navigation/mobile-menu.scss
```diff
@@ -182,7 +183,23 @@
...
+		margin-right: $d2l-navigation-margin-regular;
```
dlockhart: Invert for RTL.

sass/navigation/notifications.scss
```diff
@@ -7,13 +9,12 @@
...
 	display: inline-block;
 	height: 100%;
-	margin: 0 -15px;
 }
 
 .d2l-navigation-s-notification {
 	display: inline-block;
 	height: 100%;
-	margin: 0 15px;
+	margin-right: $d2l-navigation-margin-regular;
```
dlockhart: Invert for RTL.

sass/navigation/mobile-menu.scss
```diff
@@ -192,22 +209,20 @@
...
 		}
 	}
-	@media (max-width: $d2l-navigation-bp-mobile-menu-small) {
-		.d2l-navigation-s-logo-divider {
-			d2l-icon {
-				display: none;
-			}
+
+	.d2l-navigation-s-logo {
+		margin-right: $d2l-navigation-margin-regular;
```
dlockhart: Invert for RTL.

sass/navigation/mobile-menu.scss
```diff
@@ -192,22 +209,20 @@
...
-			}
-		}
-
-		.d2l-navigation-s-link {
-			display: none;
+	.d2l-navigation-s-home-icon {
+		flex-shrink: 0;
+		-webkit-flex-shrink: 0;
+		-moz-flex-shrink: 0;
+		margin-right: $d2l-navigation-margin-regular;
```
dlockhart: Invert for RTL.

sass/navigation/mobile-menu.scss
```diff
@@ -182,7 +183,23 @@
...
-		font-size: 1rem;
+		font-size: 0.9rem;
+		margin-right: $d2l-navigation-margin-regular;
+		@media (max-width: $d2l-navigation-bp-mobile-menu-title) {
+			display: none;
+		}
+	}
+
+	.d2l-navigation-s-divider {
+			display: inline-block;
```
dlockhart: I think there should already be CSS elsewhere that sets `display: inline-block`.

dlockhart:
Quite a few cases where RTL needs to be applied to right/left margin/padding. When I originally built nav, I was really careful to purposefully _not_ use only right or left margins and instead use equal margins on each side of things -- that was so that the CSS wouldn't need special RTL rules to invert things. Maybe there's a way to apply your changes in a similar way?

sass/navigation/divider.scss
capajon: This is like the -16px above.  Basically a 30px margin is added to the right of most nav items now.  That way when the separators are removed the spacing is as designed now.  However, the sepators themselves add too much spacing and these negative margins adjust things so the spacing overall is exactly as designed before, when the separators are shown.

sass/navigation/logo.scss
capajon: Hmmm, it did not work in my Firefox without the prefix.  Flex-shrink is all that I needed to change to fix something specifically with IE and firefox, but I can do the full flex if better.

sass/navigation/header.scss
capajon: Let me confirm again this is as designed.

sass/navigation/mobile-menu.scss
capajon: Maybe.  But I think we might want to keep it explicit as the design kind of "counts" on this font size for the header spacing to work out.  It isn't so much that lining up with the default size is important.

sass/navigation/mobile-menu.scss
capajon: Yeah I have to look at all the RTLs. Ughh, not sure how that didn't occur to me.

capajon:
The RTL consideration is too severe.  Starting over. :)

PR #605, opened 2017-07-12
New breakpoint design for navigation

capajon:
Second go at this, now with less margin changes and other issues, I hope. :)

I still used margin-right and some RTL styles on a few elements because the combination of them and sharing of styles between the mobile menu and main nav is a bit tricky and I wanted to go with something I was more confident in.  But most margins are like what they were before.  I wish the mobile menu and nav styles were separated more and only used vars and mixins to share sometimes.

I tested in all the combinations of nav settings I am aware of (more or less!) and RTL and things seem.. okay so far.  I will check again tomorrow.

I also tried to break the branding flag less than I did before, but I am changing lots of stuff that effects both on/off for it.  So I assume lots is broken before.  Actually I just tried it and indeed it is in rough shape (logo doesn't show when it needs to, etc).  It would be way too difficult to try and address that fully.  I think that in retrospect we should have copy+pasted css and wrapped the old stuff in some "branding-off" class if possible.  Oh well.  Next time maybe.

### DIFF
--- a/sass/navigation/admin-menu.scss
+++ b/sass/navigation/admin-menu.scss
@@ -17,7 +17,7 @@
 	line-height: 1.5rem;
 }
 
-@media (max-width: 992px) {
+@media (max-width: $d2l-navigation-bp-mobile-menu) {
 	.d2l-navigation-s-admin-menu {
 		display: none;
 	}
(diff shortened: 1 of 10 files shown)


dlockhart:
Yeah I wouldn't worry about the case where the flag is OFF, at this point it's not going to get disabled, only removed in time in the (hopefully short) future.

PR #616, opened 2017-07-17
DE26572-add focus style for inline dialog X button

edwardg8:


sass/dialog.scss
```diff
@@ -285,6 +285,12 @@ $dialog-resize: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.
 		background-image: $dialog-close-hover;
 	}
 
+	&:focus {
+		border-color: $d2l-color-celestine;
+		border-width: 2px;
```
dbatiste: I notice that the width of the border is different in the focus state.  This typically causes "bumping" of the element position, and can be seen in this case.  It's subtle since it's only one pixel, but it's present nonetheless.  Basically we don't want the `img` moving when we change the hover/focus state.  

There are a variety of ways this effect can be avoided.  The most straight-forward is usually to simply change the `border-color` while keeping the `border-width` constant (ex. transparent to celestine).  This case is slightly different because there are three states: normal (no visual border), hover (1px border), and focus (2px border).  There are a couple of ways one might approach this... applying `box-sizing: border-box;` may do the trick because it will tell the browser that when measuring (ex. for layout), include the padding and border in the element dimensions.  Since this element appears to specify explicit `height` and `width`, the element won't change size, and therefore won't bump position.

PR #619, opened 2017-07-20
Ported metrics for Google Analytics to daylight version of 10.6.12

Vitalii-Misechko:


Vitalii-Misechko:
I've successfully verified that 'd2l.page.visible' message is sent to Google Analytics. 
Does anyone know how to verify 'd2l.page.server' or 'd2l.page.download' event/messages?

Vitalii-Misechko:
If the fact that these metrics (`d2l.page.server`, `d2l.page.download`) exist in JS files is enough then I've verified this port completely

awikkerink:
This will also need to be updated: http://search.dev.d2l/source/xref/Lms/core/lp/framework/web/D2L.LP.Web/UI/TelemetryAnalytics/TelemetryAnalyticsPageElementLoader.cs#94

The TelemetryAnalytics is only used for testing. @CodeBaboon is this still a thing?

CodeBaboon:
@awikkerink We make use of `telemetryEndpoint` within things like my-courses, and we use D2L.Telemetry within parent portal and other repos. I don't know anything really about this TelemetryAnalytics file within the LMS though, in particular.

CodeBaboon:
Oh and I should add that when we use D2L.Telemetry inside our repos we initialize it ourselves, I believe, not relying on what appears to be going on in that link above.

awikkerink:
@edwardg8 thoughts?

edwardg8:
I'm not too sure about Google Analytics metrics has anything to do with telemetry metrics (i.e. do they have the same metrics, how are they wired up etc. ).  If we want to update http://search.dev.d2l/source/xref/Lms/core/lp/framework/web/D2L.LP.Web/UI/TelemetryAnalytics/TelemetryAnalyticsPageElementLoader.cs#94, just make sure that the new `listOfRequiredMetrics` is subset of `"D2L.Telemetry.sendPerformance(D2L.Performance.getEntriesByType( 'measure' )`, otherwise it will never send to telemetry. And `listOfRequiredMetrics` is basically the list of metrics that we want to send altogether with 1 post request, ideally all metrics we want to send.

Vitalii-Misechko:
Thank you, guys, for details.

I've just double checked the OnPrem branch and didn't found any `Telemetry` code. 
Hence, I think that  these two metrics `d2l.page.server`, `d2l.page.download` are just fine here and I should ignore them because they are used in `Telemetry` and do not cause any issues.

How does this sound?

awikkerink:
"because they `are` used in Telemetry" Did you mean `aren't`?

Vitalii-Misechko:
I meant that `d2l.page.server` and `d2l.page.download` are part of `Telemetry`. Due to the fact that `Telemetry` in LMS is not ported to OnPrem branch, these metrics are not used by anyone.

dlockhart:
So "server" and "download" won't be sent to GA, but should be visible if you just print out `D2L.Performance.timing` to the console:
https://docs.dev.d2l/index.php/Client-Side_Performance

Vitalii-Misechko:
yep, I see them

PR #620, opened 2017-07-21
DE23791-give fixed height when logo is svg

edwardg8:


sass/navigation/logo.scss
```diff
@@ -18,6 +18,10 @@
 	max-width: 260px;
 }
 
+.d2l-navigation-s-svg-logo img {
```
dbatiste: Something to be aware of (I know the current approach was already being taken): while browsers have improved greatly in terms of CSS selector matching performance, I'd typically still try to avoid generic (i.e. common tags) right-most selector phrases still applies (aside, I think this is still especially true with adjacent selectors).  There are a bunch of clever enhancements that have been made, but the more elements the right-most phrase (`img`) matches, the more checks the browser will need to do.

edwardg8: so it works better when we apply a class to `img` in this case (i.e. `svg-img` class on `img')?

dlockhart: It doesn't work any "better", it's just "faster". When the browser is trying to find the element to apply a particular CSS rule to, it starts right and then goes left. So in your case, it's going to find "all `img` elements in the entire page", and then for each one it'll look for an ancestor (walk the entire DOM tree up) to find one with the class "d2l-navigation-s-svg-logo". Then it applies the styles.

Like Dave said, they're getting good at optimizing this, but older browsers still suck at it. You can imagine a way to make it easier on the browser.

dlockhart: https://developers.google.com/web/fundamentals/performance/rendering/reduce-the-scope-and-complexity-of-style-calculations

PR #630, opened 2017-08-02
Add vanilla scroll-spy as separate bsi asset.

dbatiste:


dbatiste:
In order to resolve the bower conflict encountered in my PR build, it looks like i need to add the following resolution for `webcomponentsjs`:

```
  "resolutions": {
    "webcomponentsjs": "^0.7"
  }
```

I think this is because of the range of Polymer versions specified in for [d2l-scroll-spy](https://github.com/BrightspaceUI/scroll-spy/blob/master/bower.json#L15). 

@dlockhart : do you know if this is what we're expecting to do for these hybrid components?

dlockhart:
Hmm, yes -- this is the first hybrid element we've pulled into BSI, and normally it's an all-or-nothing approach that needs to be taken if you want to avoid forced resolutions like this.

dlockhart:
The other option would be to do a v1 release of scroll-spy that isn't hybrid, use that in BSI, then make it hybrid in v2.

PR #621, opened 2017-07-25
remove d2l-navigation webcomponent

omsmith:
See platformTools#1675

"Tangentially related to US88653 and US88649"

PR #631, opened 2017-08-18
removing reference to the "d2l-navigation" web component from on-prem

dlockhart:
This web component stopped being used in early January as it was replaced by a feature-flagged server-side rendered version instead. On-prem is not using it, and this code was removed the the `master` version of BSI [last month](https://github.com/Brightspace/brightspace-integration/pull/621).

The reason I'd like to pull it out here is that I'd like to physically remove the repository. I'll start by renaming it, but if that doesn't break anything I'm going to [delete the repo entirely](https://github.com/Brightspace/d2l-navigation-ui). If I do that though, this reference in the on-prem version will blow up next time you do a build, hence this pull request.


bower.json
```diff
@@ -14,12 +14,13 @@
...
     "d2l-my-courses": "https://github.com/Brightspace/d2l-my-courses-ui.git#0.31.2",
     "d2l-image-banner-overlay": "https://github.com/Brightspace/d2l-image-banner-overlay.git#0.0.4",
-    "d2l-navigation": "https://github.com/Brightspace/d2l-navigation-ui.git#0.15.1",
     "d2l-offscreen": "2.2.5",
     "d2l-performance": "0.0.4",
     "d2l-simple-overlay": "https://github.com/Brightspace/simple-overlay.git#0.0.4",
     "d2l-table": "0.6.3",
     "d2l-typography": "5.2.4",
+    "iron-icon": "PolymerElements/iron-icon#^1.0.0",
+    "iron-iconset-svg": "PolymerElements/iron-iconset-svg#^1.0.0",
```
dlockhart: These were explicitly included by `d2l-navigation`, but now that it's gone they need to be included in BSI.

bsi.html
```diff
@@ -10,6 +10,7 @@
 <link rel="import" href="bower_components/d2l-dropdown/d2l-dropdown-content.html">
 <link rel="import" href="bower_components/d2l-dropdown/d2l-dropdown-menu.html">
 <link rel="import" href="bower_components/d2l-menu/d2l-menu.html">
+<link rel="import" href="bower_components/d2l-menu/d2l-menu-item-link.html">
```
dlockhart: This was also explicitly imported by the navigation component, but the monolith needs it.

bower.json
omsmith: I think you mean implicitly

dlockhart: Well, they were _explicitly_ imported by the navigation component (it has the entries in bower.json and the imports), but _implicitly_ by BSI? Haha I dunno.

omsmith: Yes, they should have always been _explicitly_ included by BSI, but instead were _implicitly_ brought in, due to navigation's _explicit_ dependency on them.

PR #633, opened 2017-08-24
Add reference to activity-exemptions

RyanSweetD2L:
- Use d2l-my-courses-ui as a pattern for integrating into bsi

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "brightspace-integration",
   "dependencies": {
-    "d2l-activity-exemptions": "https://github.com/Brightspace/activity-exemptions.git#0.0.2",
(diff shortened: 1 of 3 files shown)


RyanSweetD2L:
What needs to happen to allow BSI to pull from the `activity-exemptions` github repo?  I would like to get Travis working...

dlockhart:
@RyanSweetD2L I'm guessing it's failing because it's a private repo. Does it need to be?

Also, what's the plan with consuming this in the LMS? I'm not 100% sold on the approach we've taken with my-courses (and course-image-banner)... I'm currently worried about how it's going to scale longer term. So just want to make sure we don't add to the problem.

RyanSweetD2L:
@dlockhart This is currently behind a feature flag in the LMS, but will be made generally available sometime later this year.  Do we have any mechanism for referencing a private repo from BSI?  Right now we are using the my-courses approach because we haven't found another good pattern (aside from FRA).

PR #635, opened 2017-08-30
DE27430-update-icons

edwardg8:


edwardg8:
I just updated the Award svg again with trophy instead of ribbons, it seems like Sarah and Sandra are not aligned with what icons to use for awards and grades, so very likely later we'll have to update those again.

PR #638, opened 2017-08-31
Nav remove branding flag

dlockhart:
I'm in the process of removing the feature flag around the branding work we did in January. In doing so, the following styles are now no longer needed since they only applied when the flag was in an OFF state.

This change will go alongside changes in the monolith the remove the corresponding rendering code.


sass/navigation/index.scss
```diff
@@ -4,7 +4,6 @@
 @import 'course-menu.scss';
 @import 'divider.scss';
 @import 'header.scss';
-@import 'header-course.scss'; // todo: REMOVE once le updated to not use old -course classname in server side nav
```
dlockhart: I can't find a reference to that `d2l-navigation-s-header-course` class name anywhere, so this can be removed.

sass/navigation/mobile-menu.scss
```diff
@@ -73,39 +73,11 @@
 	font-weight: 700;
 }
 
-.d2l-navigation-s-mobile-menu-org-header {
```
dlockhart: This class [is only used](http://search.dev.d2l/source/xref/LSOne/platformTools/navbars/D2L.PlatformTools.Navbars/Web/Desktop/Rendering/Views/Daylight/MobileMenu.Renderer.cs#205) when "HasBranding" if `false`.

sass/navigation/mobile-menu.scss
```diff
@@ -73,39 +73,11 @@
...
-		height: 100%;
-		min-width: 0;
-	}
-	button {
-		flex: none;
-		height: 100%;
-	}
-}
-
-.d2l-navigation-s-mobile-menu-course-header,
```
dlockhart: Same [for this one](http://search.dev.d2l/source/xref/LSOne/platformTools/navbars/D2L.PlatformTools.Navbars/Web/Desktop/Rendering/Views/Daylight/MobileMenu.Renderer.cs#238), only when HasBranding is `false`.

PR #642, opened 2017-09-06
Update to latest d2l-polymer-behaviors with swipe gesture.

dbatiste:


bower.json
```diff
@@ -14,9 +14,10 @@
     "d2l-menu": "^0.3.7",
     "d2l-more-less": "^2.2.2",
     "d2l-my-courses": "https://github.com/Brightspace/d2l-my-courses-ui.git#1.7.1",
-    "d2l-image-banner-overlay": "https://github.com/Brightspace/d2l-image-banner-overlay.git#^1.1.0",
+    "d2l-image-banner-overlay": "https://github.com/Brightspace/d2l-image-banner-overlay.git#^1.1.1",
```
dbatiste: This change eliminates the unnecessary d2l-polymer-behaviors dependency from d2l-image-banner-overlay.

bower.json
```diff
@@ -40,6 +41,7 @@
   },
   "private": true,
   "resolutions": {
+    "d2l-polymer-behaviors": "^1.1.0",
```
dbatiste: A couple of components that depend on `d2l-polymer-behaviors` have been hybridized, but we're not ready to update them here, hence the resolution here.

PR #647, opened 2017-09-26
Use activity exemptions 0.0.5 on master

jwong-d2l:


jwong-d2l:
this is so we can start testing exemptions of non-grade items

PR #649, opened 2017-09-27
Replace activity exemptions with d2l-simple-overlay

RyanSweetD2L:


jbydeley-d2l:
:shipit:

dbatiste:
How is this app integrated into the LMS?  Is it rendered in the main LMS document?  I am guessing there is some code elsewhere that is going to import the app and append it to the overlay, is that correct?

RyanSweetD2L:
@dbatiste That is correct.  The app is integrated into the LMS outside of BSI.

PR #656, opened 2017-10-03
Use d2l-fetch from bower, to take advantage of version resolution

ryantmer:
Under this new method, `.use`ing the various middlewares is done by the consumer of the element, in this case BSI. This means we can get the same versions for everything, and the same middlewares.

TODO:

- [x] Merge and release https://github.com/Brightspace/d2l-my-courses-ui/pull/391
- [x] Merge and release https://github.com/Brightspace/d2l-image-banner-overlay/pull/34
- [x] Merge and release https://github.com/Brightspace/d2l-fetch/pull/10
- [x] Merge and release https://github.com/Brightspace/d2l-fetch-auth/pull/7
- [x] Merge and release https://github.com/Brightspace/d2l-fetch-dedupe/pull/14
- [x] Merge and release https://github.com/Brightspace/d2l-fetch-simple-cache/pull/5
- [x] Update bower.json here to point to all these fancy new versions


d2l-image-banner-overlay.html
```diff
@@ -1 +1,2 @@
-<link rel="import" href="bower_components/d2l-image-banner-overlay/src/d2l-image-banner-overlay.html" />
+<link rel="import" href="d2l-fetch.html">
```
dlockhart: Maybe just from [bsi.html](https://github.com/Brightspace/brightspace-integration/blob/master/bsi.html) instead of these 2 places?

dlockhart:
Finally, the culmination of so much! So exciting.

d2l-image-banner-overlay.html
ryantmer: I initially had it in there, but that would mean all pages are going to fetch `d2l-fetch`, no?

dlockhart: Hmm, yes that's true, although it's going to get included in `shared.html` regardless since it's included in 2 separate shards. It's just a little off to me that BSI is aware that both my-courses and the image-banner both need it.

ryantmer: Oh, of course, forgot that was the case. I'll move it back to where I had it, uno momento!

PR #662, opened 2017-10-16
not measuring page events until WC-ready and page load event have fired

dlockhart:
This is another attempt to fix the sporadic errors we're seeing during UI automation. The error we're getting is this one:

> Uncaught DOMException: Failed to execute 'measure' on 'Performance': 'responseEnd' is empty: either the event hasn't happened yet, or it would provide cross-origin timing information.

That's thrown from Blink's code -- [here's a test exposing the behaviour](https://github.com/primiano/blink-gitcs/blob/master/LayoutTests/fast/performance/performance-measure-exceptions.html#L28).

The theory is that our measurement of `d2l.page.visible` -- which uses `responseEnd` as a starting point is firing _before_ the `responseEnd` mark actually occurs. This could be possible if Chrome is doing crazy optimizations somehow.

The fix is to wait until both the `WebComponentsReady` AND `load` events have fired before taking any of the measures.

### DIFF
--- a/js/page-loading/web-components-ready.js
+++ b/js/page-loading/web-components-ready.js
@@ -1,17 +1,16 @@
 (function() {
 	'use strict';
 
-	var pageReady = false;
+	if (!window.performance
+		|| !window.performance.mark
+		|| !window.performance.measure
+		|| !window.performance.getEntriesByName) return;
+
+	var wcReadyFired = false,
+		pageLoadedFired = false,
+		measured = false;
 
 	function measure(name, startMark, endMark) {
-		if (!window.performance
-			|| !window.performance.mark
-			|| !window.performance.measure
-			|| !window.performance.getEntriesByName) return;
-		if (!endMark) {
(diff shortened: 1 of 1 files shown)


js/page-loading/web-components-ready.js
```diff
@@ -23,41 +22,47 @@
...
-			pageIsReady();
+			wcReady();
 		} else {
-			addEventListener('DOMContentLoaded', pageIsReady);
+			addEventListener('DOMContentLoaded', wcReady);
 		}
 	}
 
-	setTimeout(pageIsReady, 10000);
+	setTimeout(wcReady, 10000);
```
njostonehouse: Does `pageLoaded` need a timeout too?

dlockhart: No, and technically WCReady doesn't either -- this was originally here because we were keeping the page faded out until it fired and after 10s something must have gone wrong. But now this code just does performance measuring.

PR #670, opened 2017-11-02
D2l fetch cache option

dlockhart:


bsi.html
```diff
@@ -6,7 +6,20 @@
 <link rel="import" href="bower_components/d2l-fetch-simple-cache/d2l-fetch-simple-cache.html">
 
 <script>
-	window.d2lfetch.use({name: 'auth', fn: window.d2lfetch.auth});
+	var enableTokenCache = false;
+	try {
+		enableTokenCache = document.body.classList.contains('d2l-flag-us91146-fetch-cache-token-localstorage');
```
ryantmer: Oh huh, how is this added? Is that an LD-automatic thing?

dlockhart: Yeah there's a thing we have for client-side flags. It'll be introduce in an LP pull request after this goes in and is released (when I do the BSI bump).

PR #703, opened 2018-01-10
Do not hide overflow (Chrome fix DE24456) if feature flag on

capajon:
I chose to use a note selector so it could be easily removed later on.  There is a corresponding LP change to add the feature flag related css class, but this is safe without it.


capajon:
Might need to do another approach, might close this if so.

edwardg8:
interesting, let me know how it goes

capajon:
Yeah I found a much better solution it seems - sigh - I don't think I even need a feature toggle for it.

PR #726, opened 2018-02-08
Update... a bunch of things

ryantmer:
These are all the version bumps that stem from updating siren-parser to v7.0.0. Also includes an unrelated bump to d2l-icons.


ryantmer:
Confirmed that with this build, we no longer see any issues with UMD FRAs, nor did I see any unintended side effects (d2l-image-banner-overlay and d2l-my-courses still function as expected, both of which would be quite broken if the siren-parser wasn't working correctly).

This also removes all usages of the old `siren-parser-global.html` in favour of the (also-now-deumdified) `siren-parser.html`.

PR #727, opened 2018-02-08
no longer bundling vui-more-less 

dlockhart:
The feature flag was removed from code in the LMS in `10.7.6`.


dlockhart:
This save ~6KB from `bsi.js` (uncompressed).

PR #728, opened 2018-02-09
US94228-css for role switch, impersonation or both

edwardg8:


sass/navigation/personal-menu.scss
```diff
@@ -50,3 +50,32 @@
...
+[dir='rtl'] .d2l-navigation-s-personal-menu-text-wrapper {
+	text-align: right;
+}
+
+.d2l-navigation-s-personal-menu-text-bold {
+	font-weight: bold;
+}
+
+.d2l-navigation-s-personal-menu-description {
+	font-size: 12px;
```
dbatiste: Should this really be absolute units?  Normally this would be 0.6rem.

edwardg8: I can update this with REM

dbatiste: Cool.  Yeah, pretty much anytime we are writing `px`, we should be asking ourselves... "do we want this thing to scale when the user has chosen a different font-size via Account Settings.  

For things like `1px` it doesn't make sense, ever, because the difference is so subtle, and you can get less than 1px which can be buggy because the browser may or may not render.  For bigger values it's valid thing to consider.

PR #736, opened 2018-02-13
upgrading to hybrid version of simple-overlay

dlockhart:
This upgrades the `daylight-off` branch of BSI to use the hybrid version of `simple-overlay`. A few prior commits also upgraded it to the latest Polymer. Weirdly, `simple-overlay` is the only web component used pre-Daylight by a couple random places in quizzing.

Ideally, I'd like to coordinate the upgrade to this BSI bundle with the other switchover to the hybrid `simple-overlay` for the Daylight BSI. That way the testers can just test it with Daylight OFF and ON.


bower-locker.bower.json
```diff
@@ -23,5 +23,8 @@
     "d2l-performance": "~0.0.4",
     "d2l-telemetry": "~0.0.9"
   },
-  "private": true
+  "private": true,
+  "resolutions": {
+    "d2l-colors": "^3.1.2"
```
dlockhart: Unfortunately this resolution was needed because `jquery-vui-accordion` uses `d2l-colors` -- no breaking changes though as the Sass variables remained.

PR #746, opened 2018-02-26
Bump d2l-simple-overlay

ryantmer:
Contains a small fix for the size of the scrollable area.


ryantmer:
This... seems to have pulled in a good chunk of other bumps too. Mostly patch, a few minor.

bower.json
```diff
@@ -1,14 +1,13 @@
 {
   "name": "brightspace-integration",
   "dependencies": {
-    "Stickyfill": "https://github.com/wilddeer/stickyfill.git#10d9853eda95659364a6932a03831a4df6252bf0",
```
dlockhart: Can you revert the `Stickyfill` moving around? It seems like different machines sort it alphabetically differently so it keeps moving up to the top and then down to the bottom.

dlockhart:
Yeah I've been avoiding those for a while but probably now is as good a time as ever to pull them in. All patch/minor version bumps.

bower.json
ryantmer: Even worse, I'm using git bash on Windows, so I don't even know if that's a Windows thing or not hah.

PR #754, opened 2018-02-27
Adding Paint Perf Metrics

svanherk:
Initial testing shows this working with the `timingdebug` parameter (I don't have @dlockhart's LE changes in yet to test the full flow), but wanted to get this up early to see if this was what you guys had in mind.  It didn't feel like a `measure`, but let me know if I misunderstand how I was supposed to use that.

### DIFF
--- a/js/page-loading/performance-timings.js
+++ b/js/page-loading/performance-timings.js
@@ -10,15 +10,19 @@
 		pageLoadedFired = false,
 		measured = false;
 
+	function sendPerformanceMeasure(name, measure) {
+		document.dispatchEvent(new CustomEvent('d2l-performance-measure', {
+			bubbles: true,
+			detail: { name: name, value: measure }
+		}));
+	}
+
 	function measure(name, startMark, endMark, fireEvent) {
 		window.performance.measure(name, startMark, endMark);
 		// following should be replaced with PerformanceObserver when IE/Edge support it
 		var measure = window.performance.getEntriesByName(name, 'measure');
 		if (measure.length === 1 && fireEvent) {
-			document.dispatchEvent(new CustomEvent('d2l-performance-measure', {
-				bubbles: true,
-				detail: { name: name, value: measure[0] }
-			}));
+			sendPerformanceMeasure(name, measure[0]);
 		}
 	}
 
@@ -59,6 +63,26 @@
 			wcReady();
 		} else {
 			addEventListener('DOMContentLoaded', wcReady);
+		}
+	}
+
+	if ('PerformanceObserver' in window) {
+		var observer = new window.PerformanceObserver(function(observerList) {
+			var paintMetrics = observerList.getEntries();
+			if (paintMetrics !== undefined && paintMetrics.length > 0) {
+				paintMetrics.forEach(function(paintMetric) {
+					sendPerformanceMeasure('d2l.page.' + paintMetric.name, paintMetric);
+				});
+			}
+		});
+
+		try {
+			observer.observe({
+				entryTypes: ['paint']
+			});
+		}
+		catch (e) {
+			// Need to surround this in a try-catch for browsers that have not implemented the paint entryType
 		}
 	}
 
--- a/js/page-loading/performance-timings.js
+++ b/js/page-loading/performance-timings.js
@@ -10,7 +10,7 @@
 		pageLoadedFired = false,
 		measured = false;
 
-	function sendPerformanceMeasure(name, measure) {
+	function dispatchPerformanceMeasureEvent(name, measure) {
 		document.dispatchEvent(new CustomEvent('d2l-performance-measure', {
 			bubbles: true,
 			detail: { name: name, value: measure }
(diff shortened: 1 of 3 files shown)


js/page-loading/performance-timings.js
```diff
@@ -62,6 +66,26 @@
 		}
 	}
 
+	if ('PerformanceObserver' in window) {
+		var observer = new window.PerformanceObserver(function(observerList) {
+			var paintMetrics = observerList.getEntries();
+			if (paintMetrics !== undefined && paintMetrics.length > 0) {
+				paintMetrics.forEach(function(paintMetric) {
+					sendPerformanceMeasure('d2l.page.' + paintMetric.name, paintMetric);
```
svanherk: Added the `d2l.page` for consistency in the naming, but `paintMetric.name` will still have the original name.

js/page-loading/performance-timings.js
```diff
@@ -62,6 +66,26 @@
...
+				});
+			}
+		});
+
+		try {
+			observer.observe({
+				entryTypes: ['paint']
```
svanherk: This listens for all paint events, I wasn't sure if we wanted to specify only the two we expect and all new ones are added manually.

js/page-loading/performance-timings.js
```diff
@@ -62,6 +66,26 @@
 		}
 	}
 
+	if ('PerformanceObserver' in window) {
+		var observer = new window.PerformanceObserver(function(observerList) {
```
svanherk: Used a `PerformanceObserver` rather than measuring in the `tryMeasure` function because this measure occurs and is available much later than the others

js/page-loading/performance-timings.js
```diff
@@ -62,6 +66,26 @@
...
+				});
+			}
+		});
+
+		try {
+			observer.observe({
+				entryTypes: ['paint']
+			});
+		}
+		catch (e) {
```
dlockhart: Code style nit: `catch` on the same line as the closing brace (similar to `if`-`else`)

js/page-loading/performance-timings.js
dlockhart: Nit: should just be able to go `new PerformanceObserver` minus the `window`

js/page-loading/performance-timings.js
```diff
@@ -10,15 +10,19 @@
 		pageLoadedFired = false,
 		measured = false;
 
+	function sendPerformanceMeasure(name, measure) {
```
dlockhart: Nit: I might prefer calling this `dispatchPerformanceMeasureEvent`?

js/page-loading/timing-debug.js
```diff
@@ -27,9 +27,15 @@
 		D2L.FastDom.mutate(function() {
```
dlockhart: Probably worth a small refactor here to avoid the duplication of `appendChild`/`createTextNode`:

```
var value = (e.detail.value.entryType === 'paint') ? e.detail.value.startTime : e.detail.value.duration;
timingNode.appendChild(
    document.createTextNode(e.detail.value.name + ':' Math.floor(value))
);
```

js/page-loading/performance-timings.js
```diff
@@ -62,6 +66,26 @@
 		}
 	}
 
+	if ('PerformanceObserver' in window) {
```
dlockhart: I'm assuming there's no point in using `window.performance.getEntriesByType(...)` similar to the above because the browsers who support the paint entry types also support `PerformanceObserver`? Don't get me wrong, I like using observer much better.

js/page-loading/performance-timings.js
dlockhart: That's a good question. It'll currently just be `first-paint` and `first-contentful-paint` right? I'm good either way... this way it'll get new ones in the future without having to update any code. Downside being if browsers suddenly add hundreds of paint events we'll be stuck sending a lot of data to GA.

js/page-loading/performance-timings.js
svanherk: That was my thinking, it looks like it'll only be a problem if IE/Edge implements the Paint Timing API before they implement `PerformanceObserver` (and if so, nothing breaks, we just "miss out" on those events).  I _figured_ this was a safe bet based on not even finding the Paint API in the list in https://developer.microsoft.com/en-us/microsoft-edge/platform/status/, but it's definitely just my assumption/feeling.

js/page-loading/performance-timings.js
svanherk: This was giving me a linting error when I tried, although looking again it seems like we don't actually have an `npm lint` step and we have some other linting errors in those files as well.

js/page-loading/performance-timings.js
svanherk: Yup, just those two for now.

js/page-loading/performance-timings.js
dlockhart: Seems correct to me. 👍

js/page-loading/performance-timings.js
dbatiste: Looks like Calibre is recording first meaningful paint in addition to those.  I'm guessing they must be using some trickery to determine that.

dlockhart: Yeah, they use a video recording of the page loading, break it down into frames and then do image comparisons frame-by-frame to find the "meaningful" moment.

PR #756, opened 2018-03-01
adjusting the webComponentsReady measure to start at the time origin

dlockhart:
@svanherk @dbatiste Thoughts?

This switches the start time of the `webComponentsReady` measure to the beginning of the original navigation. Previously it used the `responseEnd`, or when it received the last byte of the response. This should bring the number in-line with the paint timings and TTI measures so that they're all comparable.

The downside to this approach is that network lookup, server processing and response download all have an impact on the WCR event. But they also impact the paint and TTI timings. To give us a bit more information, I've added the `d2l.page.timeToClient` measure, which tries to capture the time it took from `navigationStart` to `responseEnd`.

### DIFF
--- a/js/page-loading/performance-timings.js
+++ b/js/page-loading/performance-timings.js
@@ -27,6 +27,7 @@
 	}
 
 	function wcReady() {
+		if (wcReadyFired) return;
 		wcReadyFired = true;
 		window.performance.mark('webComponentsReady');
 		tryMeasure();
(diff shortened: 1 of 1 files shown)


js/page-loading/performance-timings.js
```diff
@@ -27,6 +27,7 @@
 	}
 
 	function wcReady() {
+		if (wcReadyFired) return;
```
dlockhart: This is a small bug fix -- I noticed the `webComponentsReady` mark was getting fired twice because of the `setTimeout(..., 10000)` that automatically calls `wcReady` after 10 seconds no matter what.

svanherk: Why do we have that `setTimeout`?  Do we never expect it to take longer than 10 seconds, and by doing this we can easily identify outliers or hung pages?

dlockhart: So yeah, by firing it ourselves we're not going to be able to identify problem pages -- although I'd argue 10 seconds is a problem page already. But the new paint timings and especially TTI should definitely help with this.

PR #758, opened 2018-03-08
Bump d2l-table

RyanSweetD2L:


dlockhart:
This doesn't look quite right -- the numeric versions got replaced with commit hashes. I think something went wrong with the `bower-locker` process.

RyanSweetD2L:
@dlockhart I'll get it fixed.  Thanks for noticing...

PR #759, opened 2018-03-08
Bump d2l-table

RyanSweetD2L:


omsmith:
@RyanSweetD2L Unlike in stash, you coulda force pushed here :P

PR #761, opened 2018-03-12
Smaller bsijs bundle

omsmith:
Revisited this from a long while back. Output seems more obviously correct, but haven't tested locally yet.

- [x] Test locally

bsi.js is larger since we were already minifying as a transform in the browserify bundle.

| File | Master | Rollup (% Master) | Rollup & Uglify3 (% Master) |
| ----- | --------- | -------- | ---------------------- |
| bsi.js | 24683 | 40231 (163) | 40231 (163) | 
| bsi.js.gz | 7027 | 9402 (134) | 9402 (134) |
| bsi.min.js | 24174 | 21597 (89) | 21398 (89) |
| bsi.min.js.gz | 6927 | 6318 (91) | 6299 (91) |

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,13 +1,20 @@
 'use strict';
 
+import './polyfill/Array.prototype.includes.js';
+import Lie from 'lie';
+if (typeof Promise === 'undefined') {
+	window.Promise = Lie;
+}
+
 window.D2L = window.D2L || {};
 
-require('../node_modules/lie/dist/lie.polyfill.min.js');
-window.D2L.FastDom = require('./d2l-fastdom.js');
-require('../bower_components/jquery-vui-accordion/accordion.js');
-require('../bower_components/jquery-vui-change-tracking/changeTracker.js');
-require('../bower_components/jquery-vui-change-tracking/changeTracking.js');
-require('../bower_components/jquery-vui-collapsible-section/collapsibleSection.js');
-require('../bower_components/jquery-vui-scrollspy/scroll-spy.js');
-require('./page-loading/timing-debug.js');
-require('./polyfill/Array.prototype.includes.js');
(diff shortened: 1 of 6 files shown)


dlockhart:
Seems like a nice improvement, both in output size and usability of the code.

omsmith:
Loaded it up in Chrome 65 and IE11 - nothing obviously wrong.

Was able to set breakpoints in lie's `Promise` and see that hit in IE11.
Was able to set breakpoints in timing-debug and see that hit.

js/index.js
```diff
@@ -1,13 +1,20 @@
 'use strict';
 
+import './polyfill/Array.prototype.includes.js';
+import Lie from 'lie';
+if (typeof Promise === 'undefined') {
+	window.Promise = Lie;
```
dlockhart: I'm a little surprised this was necessary and that the polyfill itself wasn't doing this. I guess because lie adds a bit more than a pure polyfill? Should we be worried about that?

omsmith: I'm not including the "Polyfill" file that was in lie anymore, since iirc it wasn't bundling correctly / was getting removed.

So this is just doing a polyfill sort of deal.

omsmith:
Ping me if anyone reports any issues once this makes it into a release. Can revert it if so.

dlockhart:
Probably worth applying this change to the other 3 branches once it merges? Otherwise keeping them in sync will be more difficult.

omsmith:
Applied against polymer-2 as well. If you'd like my to do that daylightoff, can do.

dlockhart:
I think so? `fastdom` isn't included in `daylight-off` (or `daylight-off-polymer-2`) so it should be slightly different fix.

PR #764, opened 2018-03-12
Optimize email images

omsmith:


omsmith:
Mm, oops. will open a seperate branch for polymer-2

PR #768, opened 2018-03-13
Update d2l-table to v1.4.0

ppaskaris-d2l:
This PR updates `<d2l-table>` to v1.4.0, which fixes the extra white-space atop tables in Edge/IE and restores the sticky scroll buttons to their former glory in all the browsers (and under Shadow DOM). There is one known regression in Edge16 RTL that is [being tracked here](https://github.com/BrightspaceUI/table/issues/159).

[Screenshots demoing it in the LMS](https://d2lmail-my.sharepoint.com/:f:/g/personal/ppaskaris_desire2learn_com/EswaC76oJrJDh7BxlaRZFI8Br88_GXXAEFNJJW7rJn_BXg).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "brightspace-integration",
   "dependencies": {
-    "Stickyfill": "https://github.com/wilddeer/stickyfill.git#10d9853eda95659364a6932a03831a4df6252bf0",
     "app-localize-behavior": "https://github.com/PolymerElements/app-localize-behavior.git#1c7b334b36508c939fa64d2b4dc8d7c412f536f8",
     "d2l-alert": "https://github.com/BrightspaceUI/alert.git#bc6b1ca7a783913cc880858722bf8c10d5a325a5",
     "d2l-button": "https://github.com/BrightspaceUI/button.git#ec051badc534f4eed0b568070ba4151950d5b59e",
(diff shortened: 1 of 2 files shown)


bower.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "brightspace-integration",
   "dependencies": {
-    "Stickyfill": "https://github.com/wilddeer/stickyfill.git#10d9853eda95659364a6932a03831a4df6252bf0",
```
ryantmer: It's a bleh thing, but I've been keeping this at the top - it keeps moving because of platform differences with capitalization I think :/

dlockhart: Yeah, can you revert this part out of the commit? Otherwise it's an endless loop of it moving to the top for Mac/Linux users and then moving down again for Windows users. Or vice-versa, I'm not actually sure.

ppaskaris-d2l: Yeah, I can move it back.

ppaskaris-d2l:
Is there anything else I'll need to do to get this into the product before the deadline? I see some other PRs are duplicated into a different branch: should I do that too?

dlockhart:
If you could also put the same change into the `polymer-2` branch it would be very much appreciated.

Other than that, this needs to merge, you need to do a new release of BSI from the "Releases" tab in GitHub and [the config file in the LMS](http://search.dev.d2l/source/xref/Lms/core/lp/_config/Infrastructure/D2L.LP.Web.UI.Html.Bsi.config.json) needs to be updated to point at that new release.

PR #776, opened 2018-03-19
build output to build directory instead of dist

dlockhart:
This switches build output from the `dist` directory to `build`. Reason being: `polymer build` (which we're now using) puts its stuff there anyway, so might as well just have a single output directory.


package.json
```diff
@@ -5,33 +5,32 @@
...
+    "build:sass:bsi": "node-sass --output-style expanded ./sass/bsi.scss ./build/bsi.css",
+    "build:sass:datagrid": "node-sass --output-style expanded ./sass/datagrid/datagrid.scss ./build/datagrid.css",
+    "build:sass:homepages": "node-sass --output-style expanded ./sass/homepages/homepages.scss ./build/homepages.css",
     "postbuild:sass": "npm run autoprefix && npm run uglify:css",
-    "build:icons": "cpy --parents --cwd=bower_components/d2l-icons \"images/**/*.*\" ../../dist",
-    "build:navicons": "cpy \"nav-icons/*.*\" dist/images/nav-icons",
+    "build:icons": "cpy --parents --cwd=bower_components/d2l-icons \"images/**/*.*\" ../../build",
+    "build:navicons": "cpy \"nav-icons/*.*\" build/images/nav-icons",
     "build:email-icons": "node build-email-icons.js",
-    "prebuild:wc": "rimraf tmp && rimraf tmp-wc",
```
omsmith: _tmp-wc_ was from the old polymer-build.js?

polymer.json
```diff
@@ -11,6 +11,9 @@
 	  "d2l-rubric.html"
   ],
   "sources": [],
+  "extraDependencies": [
+    "bower_components/webcomponentsjs/*.js"
```
omsmith: ?

dlockhart: This should have been in a separate PR later, I'll remove it.

package.json
dlockhart: Yeah. It used to create `tmp` as it was building, and if the build failed, `tmp` stuck around. Polymer CLI's `build` (which uses a newer `polymer-build`) doesn't have this issue (tested it).

`tmp-wc` was where we told `polymer-build` to put things, which is also not necessary anymore.

dlockhart: Updated: removed this line entirely, as neither cleanup is needed anymore.

package.json
```diff
@@ -5,33 +5,31 @@
...
     "build:wc": "polymer build",
-    "postbuild:wc": "cpy build/bundled/*.* dist && rimraf build && npm run jsonpify-imports",
+    "postbuild:wc": "cpy build/bundled/*.* build && rimraf build/bundled && npm run jsonpify-imports",
```
dlockhart: Next step is going to be to serve BSI web components directly from the `bundled` directory in the LMS. Further steps will allow for different browsers to get web components from different directories, like ES5 vs. ES6, bundled vs. unbundled and potentially an IE11-specific place as well.

PR #779, opened 2018-03-20
Egao/de29150

edwardg8:


sass/navigation/personal-menu.scss
```diff
@@ -26,6 +26,7 @@
 	margin-left: 10px;
 	max-width: 200px;
 	min-width: 40px;
+	line-height: 1.6;
```
edwardg8: I think line-height: 1.6 is the closest to what have now, not very noticeable about the difference. The margin-top: 5px isn't from any design specs, so I believe it's ok with slight fluctuations of the new spacing.

PR #780, opened 2018-03-20
centralize web components

dlockhart:
This change centralizes all the web components under `web-components` directory. This makes the separation clearer, but mostly it's so that we can enable `eslint` again. Because there are so many different types of things to lint (modules, scripts, HTML files), having them in separate directories is necessary.

This will require a change in the LMS as all HTML import assets will now be found under `{bsi}/bundled/web-components/`. I'll be applying the same change to the other branches and fixing everything in the LMS at once.


jbydeley-d2l:
@dlockhart Have we considered using [Prettier](https://prettier.io/) along with ESLint. There are few things I loath more than having CI fail because of a lint that could / should be auto-corrected on save.

omsmith:
@jbydeley-d2l `eslint --fix`?

dlockhart:
@jbydeley-d2l I'd be open to that, never used it myself. Currently BSI isn't even using `eslint` (not sure when that stopped), so this is an attempt to get things in place to start doing that again.

PR #781, opened 2018-03-21
re-introducing eslint

dlockhart:
Required moving a couple things around such that modules were together and scripts were together.


dlockhart:
This messed something up with the rollup build, closing for now and will re-open when I fix it.

PR #782, opened 2018-03-21
re-introducing eslint

dlockhart:
attempt #2


dlockhart:
I've done a side-by-side diff of the before & after output and they're identical.

PR #787, opened 2018-03-21
update d2l-rubric version

kieranderson:


dlockhart:
@kieranderson can you wait 12-18 hours before this gets into the product? I'm right in the middle of some major re-organization of how the BSI web components are laid out, and am making the same changes to all 4 branches and will then update Brightspace accordingly. I can pull in this change at the same time, but it probably won't be until later tonight or tomorrow morning that it all merges.

kieranderson:
@dlockhart Yes that will work.

dlockhart:
Awesome. So once you merge this, don't worry about also making the change in the `polymer-2` branch, releasing BSI or updating the LMS -- I'll handle that.

PR #797, opened 2018-04-04
ES5 transpile

dlockhart:
Two weeks in the making folks! Haha this is the BSI side, which actually ends up being really simple -- the fun stuff has been on the LMS side, which I'll open shortly.

All this does is produce a second `es5-bundled` build output from Polymer's build step, passing `compile: true` to it. Under the covers, that'll use Babel to transpile the ES6 class stuff back to ES5 for IE11 (and older evergreen browsers).

Still outstanding: the JSON-ify imports stuff. I'll deal with this as a separate story, but this is yet another IE11-specific thing that makes it treat the HTML files as JS files to get around a terrible caching bug. Will confirm if it's still an issue and take the appropriate action.


package.json
```diff
@@ -32,7 +32,7 @@
     "uglify:css:datagrid": "cleancss -s ./build/datagrid.css -o ./build/datagrid.min.css",
     "uglify:css:homepages": "cleancss -s ./build/homepages.css -o ./build/homepages.min.css",
     "uglify:js": "uglifyjs ./build/bsi.js --compress --mangle --source-map \"content='build/bsi.js.map',url='bsi.min.js.map'\" -o ./build/bsi.min.js",
-    "jsonpify-imports": "node jsonp-imports.js \"./build/**/*.html\"",
+    "jsonpify-imports": "node jsonp-imports.js \"./build/es5-bundled/**/*.html\"",
```
svanherk: Is this the part you mentioned about the IE11-specific caching problem?  So before, everyone was getting the html to js behaviour, but now it will only be those that need the es5 bundle (if it's still needed at all)?

dlockhart: So with the new `v1` polyfill, IE stopped using the JS files and went back to using the HTML files -- since the code that made it do that was in our fork of the `v0` polyfill. This doesn't change any of that, it just continues to build the JSONP files alongside the ES5 bundle. They won't be used unless we make the same change to the `v1` polyfill -- which I may end up having to do but want to look into it more first.

PR #812, opened 2018-04-26
Adding tile and demarcation styles

dbatiste:


sass/homepages/homepages.scss
```diff
@@ -1,6 +1,32 @@
 $d2l-vui-grid-system-gutters: 100/70;
 @import '../../bower_components/vui-grid-system/grid-system.scss';
```
dlockhart: Nit: SASS lets you nest things, like this:

```
.daylight-widget-demarcation {
    .d2l-homepage-pageheader {}

    .d2l-widget {}
}
```

sass/tile.scss
```diff
@@ -0,0 +1,19 @@
+@import '../bower_components/d2l-colors/d2l-colors.scss';
```
dlockhart: Don't think this import is used?

dbatiste: Ah, yeah I should not have added that yet.  It will be though.

sass/homepages/homepages.scss
dbatiste: Oh yeah.  Will update.

PR #817, opened 2018-05-02
DE29441 - Increase `d2l-more-less` version

svanherk:
- 4.1.0: 
  - Fixes auto-expanding when dynamic content added in Shadow DOM
  - Basic test framework setup
- 4.0.0:
  - Remove support for browsers without focusin/out

Also seem to be catching a number of other minor updates here.


svanherk:
@mpharoah-d2l Looks like https://github.com/BrightspaceUI/table/blob/f4ac70a8a08b488cee2c197722083bbbc92e6f40/d2l-tspan.html#L3 is causing an error.  The minor update is getting picked up here, is this not ready to go to production?

dlockhart:
@jwong-d2l Sure, or quickly patch it as version `1.5.1`.

PR #819, opened 2018-05-03
Updates for Daylight widget demarcation animation.

dbatiste:


bower-locker.bower.json
```diff
@@ -21,7 +21,7 @@
     "d2l-more-less": "^4.1.0",
     "d2l-my-courses": "Brightspace/d2l-my-courses-ui#^4.5.1",
     "d2l-offscreen": "^3.0.3",
-    "d2l-polymer-behaviors": "^1.3.0",
```
svanherk: I caught these changes in the release I just did, which is why we see no other updates in this file

dbatiste: Yeah thanks!  I noticed you were about to merge so I waited to get your changes.  ;)

PR #820, opened 2018-05-07
Import d2l-rubric-editor

wongvincent:


web-components/d2l-rubric.html
```diff
@@ -1 +1,2 @@
 <link rel="import" href="../bower_components/d2l-rubric/d2l-rubric.html">
+<link rel="import" href="../bower_components/d2l-rubric/editor/d2l-rubric-editor.html">
```
dlockhart: We chatted about this in Slack, but this should be a separate HTML import so that pages who don't need the entire editor don't include it.

bower.json
```diff
@@ -83,7 +86,6 @@
   },
   "private": true,
   "resolutions": {
-    "Stickyfill": "ec3a2ba7466515c7863eb108282ad18b10543a15",
```
dlockhart: Can you revert the `Stickyfill` change? Windows users have it go to the top and Mac/Linux users put it with the "S"s alphabetically.

bower.json
```diff
@@ -197,11 +202,13 @@
...
       "d2l-organization-hm-behavior": "2.0.2",
       "d2l-polymer-behaviors": "1.5.0",
-      "d2l-rubric": "0.26.0",
+      "d2l-rubric": "0.39.0",
       "d2l-scroll-spy": "0.0.3",
       "d2l-search-widget": "3.0.3",
       "d2l-simple-overlay": "2.1.2",
       "d2l-table": "1.5.1",
+      "d2l-text-input": "0.3.0",
+      "d2l-textarea": "b0d536179c",
```
dlockhart: This doesn't quite seem right... shouldn't it be using a released version?

wongvincent: So the textarea is temporary and is a repo in Martin's account until it's moved into BrightspaceUI, should we still create a release for it?

PR #830, opened 2018-05-11
Use latest d2l-rubrics

wongvincent:


bower.json
```diff
@@ -170,11 +170,11 @@
     "webcomponentsjs": "8a2e40557b177e2cca0def2553f84c8269c8f93e"
   },
   "bowerLocker": {
-    "lastUpdated": "2018-05-10T21:50:07.443Z",
+    "lastUpdated": "2018-05-11T19:05:47.569Z",
     "lockedVersions": {
       "app-localize-behavior": "2.0.1",
       "d2l-alert": "3.0.1",
-      "d2l-button": "4.0.3",
+      "d2l-button": "4.2.1",
```
dlockhart: @margaree FYI this would pull in your button changes into BSI. I _think_ that should be OK, since the LMS doesn't actually use the new icon or subtle buttons yet, and your changes to `image-action` aren't included here.

wongvincent: Made a new commit that won't be using the new d2l-button version because it was causing the CI to fail

PR #831, opened 2018-05-14
DE30092-add top margin for each panel

edwardg8:


sass/homepages/homepages.scss
```diff
@@ -64,7 +64,15 @@ $viewport-width-sm: 768px;
 		margin: 0;
 	}
 
-	.d2l-homepage {
+	.homepage-col-4 {
```
dbatiste: There is margin-top at the top of each widget, and there is CSS below here that overrides this to 0.  Not sure what the intent is here... there is some spacing to consider for when the banner is present, and also when there is page title visible.

edwardg8: So the defect is that when there are multiple rows layouts are selected, there is no margin between the panels. It looks fine when there is only one row (a left and right panel), but not when there are multiple because the top margin we had earlier is on entire `homepage`

PR #852, opened 2018-06-13
polyfill findIndex for IE11

mdgbayly:
@omsmith @ChrisKraljevicD2L We'll most likely try and get this merged into master tonight and deployed to the LMS.

But wanted to give you a heads up in case it causes any issues or in case there is a better way to handle this. I just followed what had been done to polyfill `Array.includes`.

This is required for some recent changes to add domain whitelisting to `d2l-fetch-siren-entity-behavior`. That lib was updated in BSI yesterday and is required for d2l-rubric.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,4 +1,5 @@
 import './polyfill/Array.prototype.includes.js';
+import './polyfill/Array.prototype.findIndex.js';
 import Lie from 'lie';
 if (typeof Promise === 'undefined') {
 	window.Promise = Lie;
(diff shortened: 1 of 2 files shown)


omsmith:
I think you'll want to revert https://github.com/Brightspace/brightspace-integration/pull/851, believe it broke BVT earlier today.

mdgbayly:
@omsmith oh really - thanks for the heads up - yeh - i noticed it had not been updated in the config but didn't check to see if it had been committed and reverted... but now I see it has... sigh - ok thanks...

PR #860, opened 2018-06-19
Fail when deployment fails

ryantmer:
According to the [Travis docs](https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build), a failure in `after_success` will not actually fail the build. This means that the deployment step failing would not actually fail the job, resulting in an erroneous success state.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,7 @@ sudo: false
 script:
 - npm run test
 - npm run build
-after_success:
-- '[ "$TRAVIS_SECURE_ENV_VARS" = "true" ] && [ "x$TRAVIS_TAG" != "x" ] && npm run
-  publish:cdn'
+- 'if [ "$TRAVIS_SECURE_ENV_VARS" == "true" ] && [ "x$TRAVIS_TAG" != "x" ]; then npm run publish:cdn; fi'
 env:
   global:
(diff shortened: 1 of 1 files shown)


ryantmer:
This occurred late last week, with `10.8.3-daylight.13` - [the deployment](https://travis-ci.org/Brightspace/brightspace-integration/builds/392709518) partially worked, but some files failed to upload, and the build was still green, giving no indication of the failure.

Unfortunately, I tried to re-run the same build, which means those errors are gone, but they're replaced by a "No files transferred because files already exists in lib/bsi/10.8.3-daylight.18/" message. I'm not clear whether _that_ would fail based on what I've got here (not sure if `frau-publisher` exists with 0 there), but this should now fail the BSI build if the deployment outright fails.

dlockhart:
I think this makes sense. If the build stages fail, hopefully it just stops and doesn't try to publish?

ryantmer:
Exactly, yeah - according to the docs at least, if anything in `script` fails, it'll skip the remaining `script` steps, whereas if something fails in `after_success`, it'll still show as green and not notify via email of a failure or anything.

dlockhart:
Cool, seems like an excellent change then!

PR #870, opened 2018-07-10
Use latest d2l-rubric

wongvincent:
Use latest d2l-rubric.

I reverted the change to use d2l-icons 5.x because other components are still depending on 4.x
image


dlockhart:
Including @margaree here as well. This will definitely break some stuff, as version `5.0.0` of icons comes with breaking changes -- the removal of the `<d2l-icon-button>` component. You have a couple options:

1. Stick with version 4 of icons in rubrics for now. This is a temporary solution.
2. Go through the other things in BSI that depends on `d2l-icons` and upgrade them all to use version 5. If they're not using `<d2l-icon-button>`, it'll be an easy change. If they are, they'll need to switch off it to `<d2l-button-icon>` from the `d2l-button` repo. It should be pretty easy to do a quick analysis and see who's using what.

wongvincent:
Thinking of doing the temporary solution. Trying to get our rubric work tested ASAP

wongvincent:
@margaree Sorry for asking you to merge the d2l-icons 5.x bump earlier. It's going to have to be done again because I reverted it back to 4.x.

margaree:
No worries! I'll take a look at bumping everything when I have a chance. It should be pretty straightforward since I think I removed all uses of `d2l-icon-button` from BSI things.

dlockhart:
Thanks @margaree 💯 !

PR #876, opened 2018-07-13
Add git instead of https info to troubleshooting guide

jeroach:


jeroach:
Thanks Dave! FYI I switched the ordering around so the fix command is `git config --global url."https://".insteadOf git://`. The other way prevented me from pushing changes as git tried to reference `git://github.com/Brightspace/brightspace-integration/`

PR #880, opened 2018-07-19
Remove vui-breadcrumbs

dlockhart:
I'm going to pull all the ancient `vui-` stuff directly into BSI, starting here with breadcrumbs.

My reasoning is that most of these will become full-fledged Brightspace UI components at some point (maybe), but in the meantime the old `vui-` repos are still referencing colors and icons and such... and they're so old that they're starting to break. And most of them don't build anymore to even be able to go in and update them.

### DIFF
--- a/sass/deprecated/breadcrumbs.scss
+++ b/sass/deprecated/breadcrumbs.scss
@@ -1,0 +1,45 @@
+@import '../../bower_components/d2l-colors/d2l-colors.scss';
+@import '../../bower_components/d2l-link/d2l-link.scss';
+
+.vui-breadcrumbs {
+	color: $d2l-color-ferrite;
+	font-family: inherit;
+	font-size: 0.7rem;
+	font-weight: 400;
+	line-height: 1rem;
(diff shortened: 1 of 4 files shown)


sass/deprecated/breadcrumbs.scss
```diff
@@ -0,0 +1,45 @@
+@import '../../bower_components/d2l-colors/d2l-colors.scss';
+@import '../../bower_components/d2l-link/d2l-link.scss';
```
dlockhart: This is copied from here:
https://github.com/Brightspace/d2l-breadcrumbs-ui/blob/master/breadcrumbs.scss

svanherk:
What does this mean for FRAs still using this, is the last version still available if you remove these repos?  Or do they need to update to the `d2l-*` versions next time they build?

dlockhart:
This won't impact FRAs at all -- the repos will still exist and continue to exist, they'll just be deprecated/archived to reflect the fact that they're antiquated and we need to re-invest in making web component versions.

PR #883, opened 2018-07-20
upgrading shadycss from 1.3.1 to 1.3.5

dlockhart:
Diff:
https://github.com/webcomponents/shadycss/compare/v1.3.1...v1.3.5


dlockhart:
@mdulat, [you originally added the resolution](https://github.com/Brightspace/brightspace-integration/commit/127e042bff3c4d109b3dc41bbc9e0129fb7e8c10) that was keeping the shadycss polyfill at version `1.2.0` -- do you recall why? I'd like to keep up with updates to the polyfill unless there's a bug.

dlockhart:
Actually, since it was `^1.2.0` in the resolution, I don't think that would have prevented it from upgrading to `1.3.x` at all. So the resolution is probably just not necessary unless there was originally a version conflict.

mdulat:
I think it was just a version conflict. If that's no longer the case and it's building fine without the resolution, then it's fine to remove it.

PR #885, opened 2018-07-20
US97722-remove color for link item

edwardg8:


sass/navigation/common.scss
```diff
@@ -35,7 +35,6 @@ $d2l-navigation-bp-title: 740px;
 a.d2l-navigation-s-link,
 a.d2l-navigation-s-link:link,
 a.d2l-navigation-s-link:visited {
-	color: $d2l-color-ferrite;
```
dlockhart: You shouldn't _need_ to remove this as long as you specify the exact same rule when you inline the override CSS into the page.

dlockhart: Ah right, but I see you're using the `d2l-branding-navigation-foreground-color` CSS class inline in the page. 👍

edwardg8: Ha, yea, this override that, so have to get rid of it

PR #888, opened 2018-07-23
US97722-have to use more specific css rule to override the text color when hover and focus

edwardg8:


edwardg8:
This probably will change again since Jeff mentioned we probably won't use the same hover/focus style any more, but he's still investigating.

PR #906, opened 2018-07-27
US97722-css for using link area colors in mobile and pully of image-b…

edwardg8:
…ased nav

### DIFF
--- a/sass/navigation/common.scss
+++ b/sass/navigation/common.scss
@@ -16,6 +16,9 @@ $d2l-navigation-bp-title-hidden: 488px;
 $d2l-navigation-bp-mobile-menu-compact: 380px;
 $d2l-navigation-bp-title: 740px;
 
+$d2l-branding-navigation-dark-foreground-color: #565A5C;
+$d2l-branding-navigation-light-foreground-color: #FFFFFF;
+
 
 .d2l-navigation-s-centerer {
 	margin: 0 auto;
@@ -38,8 +41,21 @@ a.d2l-navigation-s-link:visited {
 	cursor: pointer;
 	text-decoration: none;
 }
-a.d2l-navigation-s-link:hover,
-a.d2l-navigation-s-link:focus {
+
+.d2l-branding-navigation-dark-foreground-color a.d2l-navigation-s-link,
+.d2l-branding-navigation-dark-foreground-color a.d2l-navigation-s-link:link,
+.d2l-branding-navigation-dark-foreground-color a.d2l-navigation-s-link:visited {
+	color: $d2l-branding-navigation-dark-foreground-color
+}
+
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link,
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link:link,
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link:visited {
+	color: $d2l-branding-navigation-light-foreground-color
+}
+
+.d2l-navigation-s a.d2l-navigation-s-link:hover,
+.d2l-navigation-s a.d2l-navigation-s-link:focus {
 	color: $d2l-color-celestine;
 	text-decoration: underline;
 	outline: none;
--- a/sass/navigation/main-image-based.scss
+++ b/sass/navigation/main-image-based.scss
@@ -2,14 +2,14 @@
 @import '../../bower_components/d2l-typography/d2l-typography.scss';
 
 @mixin _d2l-dark-overlay() {
-	background:rgba(0,0,0,0.1);
-	content:'';
-	height:100%;
-	left:0;
-	opacity:0;
-	position:absolute;
-	top:0;
-	width:100%;
+	background: rgba(0,0,0,0.1);
+	content: '';
+	height: 100%;
+	left: 0;
+	opacity: 0;
+	position: absolute;
+	top: 0;
+	width: 100%;
 }
 
 .d2l-navigation-main-ib[ib-pully-focus],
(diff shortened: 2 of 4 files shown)


sass/navigation/common.scss
```diff
@@ -38,8 +41,21 @@ a.d2l-navigation-s-link:visited {
...
+.d2l-branding-navigation-dark-foreground-color a.d2l-navigation-s-link:link,
+.d2l-branding-navigation-dark-foreground-color a.d2l-navigation-s-link:visited {
+	color: $d2l-branding-navigation-dark-foreground-color
+}
+
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link,
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link:link,
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link:visited {
+	color: $d2l-branding-navigation-light-foreground-color
+}
```
dlockhart: Does this still work when your feature flag is OFF?

sass/navigation/mobile-menu.scss
```diff
@@ -72,6 +72,11 @@
 	font-weight: 700;
 }
 
+.d2l-navigation-s-mobile-menu-color-strip{
```
dlockhart: Nit: space

dlockhart:
Added @dbatiste to look at the image-based nav tray thing

sass/navigation/common.scss
edwardg8: yep, I thoroughly tested most scenarios, they still work the same as before with flag-off.

sass/navigation/mobile-menu.scss
edwardg8: Added it :)

sass/navigation/common.scss
edwardg8: BTW, with flag off, you can only get `.d2l-branding-navigation-dark-foreground-color`

sass/navigation/common.scss
```diff
@@ -16,6 +16,9 @@ $d2l-navigation-bp-title-hidden: 488px;
 $d2l-navigation-bp-mobile-menu-compact: 380px;
 $d2l-navigation-bp-title: 740px;
 
+$d2l-branding-navigation-dark-foreground-color: #565A5C;
```
dlockhart: This could be `$d2l-color-ferrite`

sass/navigation/common.scss
```diff
@@ -38,8 +41,21 @@ a.d2l-navigation-s-link:visited {
...
+
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link,
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link:link,
+.d2l-branding-navigation-light-foreground-color a.d2l-navigation-s-link:visited {
+	color: $d2l-branding-navigation-light-foreground-color
+}
+
+.d2l-navigation-s a.d2l-navigation-s-link:hover,
+.d2l-navigation-s a.d2l-navigation-s-link:focus {
 	color: $d2l-color-celestine;
```
dbatiste: Is the celestine color going away?

sass/navigation/main-image-based.scss
```diff
@@ -1,6 +1,16 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
 @import '../../bower_components/d2l-typography/d2l-typography.scss';
 
+@mixin _d2l-dark-overlay() {
+	background:rgba(0,0,0,0.1);
```
dbatiste: Super nit-picky... we usually put a space after `:` when setting properties.

sass/navigation/main-image-based.scss
```diff
@@ -230,10 +240,62 @@ div.d2l-navigation-ib-item-icon-group-icon {
...
+	    opacity:1;
+	}
+
+	.d2l-navigation-main-ib-pully-has-branding[shown]:focus .d2l-navigation-main-ib-pully-tab-button {
+		background-color: rgba(0,0,0,0.1);
+		opacity: 1;
+	}
+}
+
+.d2l-navigation-main-ib-pully-icon > d2l-icon {
```
dbatiste: I presume these `z-index` values are necessary since they are here... but it seems like they should not be needed.  

If we _do_ need these, it seems like the values of 6/7 are incremented from elsewhere.  Note that whenever a new stacking context is created, one could simply use 0,1,2, etc within that context.

If we _do_ need these, it seems like `> d2l-icon` should not be necessary on this selector.

I think would help to see a rendered DOM with any other z-index values.

edwardg8: I continued z-index from here, https://github.com/Brightspace/brightspace-integration/blob/master/sass/navigation/main-image-based.scss#L248. So the button on the pully has to be on top of others except the icon, hence 6,7. I'm not sure how we can get the same result without playing with z-index

dbatiste: So it looks like we will have:
```
tab (per above): z-index = 5
   button (not really a button): z-index = 6
   pully icon
      d2l-icon: z-index = 7
```
It doesn't really matter much, but it would be the same as:
```
tab (per above): z-index = 5
   button (not really a button): z-index = 0
   pully icon
      d2l-icon: z-index = 1
```
But it seems like it could really be:
```
tab (per above): z-index = 5
   button (not really a button): z-index = 0
   pully icon: z-index = 1
```
Elements that follow other elements in the DOM will stack on top of the previous siblings.  So really I am wondering if we need the z-index values on these children at all.

It would be good for me to inspect the DOM to be sure.  Much easier to see it composed.

edwardg8: I think your theory of z-index worked, I first tried out with 0/1, things still show up right. Then I removed both z-indexes completely. Things still look right to me. I'm kinda surprised how it works, I was expecting the cab which has higher z-index appears on top of the rectangular overlay (button), but it's not.

edwardg8: I think you already explained the magical part why it works without z-index "Elements that follow other elements in the DOM will stack on top of the previous siblings" and that was absolutely right

PR #908, opened 2018-07-30
update rubric version

stevenao:


mdulat:
Should bower-locker file also be updated?

stevenao:
^0.78.0 - the caret sign should pickup new minor version. But, I made the change just to be safe... :)

wongvincent:
because version is still 0.x, caret won't pick up new minor version, only the patches

stevenao:
good point.. yeah, i got mixed up with the ~ sign.

PR #913, opened 2018-08-02
Add d2l-navigation-separator reference and remove unused CSS

jeroach:
`d2l-navigation` PR: https://github.com/BrightspaceUI/navigation/pull/3
`platformTools navbars` PR: https://github.com/BrightspaceUI/navigation/pull/3

**Navigation PR needs to be merged before this passes CI.**


sass/navigation/divider.scss
```diff
@@ -1,17 +1,7 @@
-@import '../../bower_components/d2l-colors/d2l-colors.scss';
 @import 'common.scss';
 
 .d2l-navigation-s-divider {
```
dlockhart: I think we might be able to totally get rid of this class name (and the `<div>`) and replace it with something like:

```
.d2l-navigation-s d2l-navigation-separator {
    @media (max-width: $d2l-navigation-bp-separators) {
        display: none;
    }
}
```

jeroach: Yep, this works! Thanks!

sass/navigation/divider.scss
```diff
@@ -1,17 +1,9 @@
...
-		color: $d2l-color-mica;
-	}
-	@media (max-width: $d2l-navigation-bp-separators) {
- 		display: none;
- 	}
-}
-.d2l-navigation-s-divider-no-margin {
-	margin: 0;
+
+    @media (max-width: $d2l-navigation-bp-separators) {
```
svanherk: Super minor: The formatting here went a little weird

jeroach: Hmmm, Atom doesn't seem to show a difference between the formatting of this file and other `.scss` files. Perhaps it's just the GitHub diff?

PR #916, opened 2018-08-02
US97007 - Publish d2l-my-dashboard into BSI

mtjandra:
After this I believe there are two things left to do.
1. Publish a release
2. Update the LMS to use the latest BSI

Seems like I only need to updated `daylight-polymer-1` since we do not use polymer-2 or polymer-3? is there a reason why there are not many updates to the `daylight-off-polymer-X` ones?

### DIFF
--- a/bower-locker.bower.json
+++ b/bower-locker.bower.json
@@ -34,7 +34,8 @@
     "jquery-vui-collapsible-section": "^1.2.2",
     "jquery-vui-change-tracking": "^0.6.0",
     "polymer": "^1.11.3",
-    "susy": "^2.2.12"
+    "susy": "^2.2.12",
+    "d2l-my-dashboards": "^1.0.0"
   },
   "private": true,
   "resolutions": {
--- a/polymer.json
+++ b/polymer.json
@@ -5,6 +5,7 @@
     "web-components/d2l-image-banner-overlay.html",
     "web-components/d2l-more-less.html",
     "web-components/d2l-my-courses.html",
+    "web-components/d2l-my-dashboards.html",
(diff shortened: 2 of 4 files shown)


bower-locker.bower.json
```diff
@@ -19,6 +19,7 @@
     "d2l-menu": "^1.1.0",
     "d2l-more-less": "^4.3.0",
     "d2l-my-courses": "Brightspace/d2l-my-courses-ui#^5.6.1",
+    "d2l-my-dashboards": "Brightspace/d2l-my-dashboards#^1.0.0",
```
LiyueWang: `Brightspace/d2l-my-dashboards-ui`

seems we don't have the version for `d2l-my-dashboards-ui` yet?

mtjandra: I registered it as `d2l-my-dashboard` similar to how the `d2l-my-courses-ui` is registered as `d2l-my-courses`. I did release a version https://github.com/Brightspace/d2l-my-dashboards-ui/releases. You should be able to install it with bower

jeroach: I think this line needs to either be:
`"d2l-my-dashboards": "Brightspace/d2l-my-dashboards-ui#^1.0.0",` OR `"d2l-my-dashboards": "^1.0.0"`

polymer.json
```diff
@@ -5,6 +5,7 @@
     "web-components/d2l-image-banner-overlay.html",
     "web-components/d2l-more-less.html",
     "web-components/d2l-my-courses.html",
+    "web-components/d2l-my-dashboards.html",
```
mtjandra: I just added this.

PR #923, opened 2018-08-07
Add d2l-telemetry-browser-client

walinga:
For eventual use by `d2l-rubric` telemetry


omsmith:
* This doesn't add it into BSI in any meaningful way. It's simply being installed by bower and then being ignored
* Wouldn't this just come in as part of _d2l-rubric_?

walinga:
No need for this PR since these changes will come in as part of `d2l-rubric`

PR #929, opened 2018-08-13
update rubric

mdulat:


bower.json
```diff
@@ -206,9 +206,9 @@
       "d2l-organization-hm-behavior": "2.0.3",
       "d2l-organizations": "0.0.3-hybrid",
       "d2l-polymer-behaviors": "1.6.2",
-      "d2l-polymer-siren-behaviors": "0.0.2",
+      "d2l-polymer-siren-behaviors": "0.0.1",
```
wongvincent: reverted?

mdulat: Yes, this ref comes from rubrics, which originally didn't specify a version #, but there was only 1 version. Now there's version 0.0.2 which is causing havoc with rubrics.

PR #970, opened 2018-09-06
d2l my courses bump to v5.7.10

m6jones:


bower-locker.bower.json
```diff
@@ -4,6 +4,7 @@
     "d2l-alert": "^3.1.0",
     "d2l-button": "^4.7.3",
     "d2l-button-group": "^2.1.0",
+    "d2l-card": "BrightspaceUI/card#^4.1.9",
```
dbatiste: Is there anything else the depends on the `d2l-tile` path?  Hopefully we can remove that from here.

m6jones: I am not sure, I know another team was using the tile path.

m6jones: I'll can try and track down to see if there are any left. It be nice to remove it.

PR #980, opened 2018-09-13
Egao/semester course code

edwardg8:


edwardg8:
I think the changes have to go in with changes in LMS, otherwise UI may break abit.

PR #1001, opened 2018-09-25
Updated d2l my courses

m6jones:
I had to update `BrightspaceUI/inputs#^0.0.1` to `BrightspaceUI/inputs#^1.0.0` is this okay @dlockhart? Note this is going into version 10.8.6 as a cert fix.


dlockhart:
Eek... that makes me a little uncomfortable as there hasn't been much testing yet. Why does this have to be updated in 10.8.6? @svanherk FYI

svanherk:
That's odd, `my-courses` doesn't use any of the inputs and I'm not seeing an obvious shared dependency issue.  What conflict were you seeing?

svanherk:
It looks like a fresh `bower install` is trying to pull in the new `d2l-my-dashboards` minor version, which uses the new `d2l-inputs`.  But we don't `d2l-my-dashboards` to update in cert either, so we can just throw away those changes.  I'm also seeing `tile` and `card` updates, are those needed for this `my-courses` change?  Or is that repo the **only** one we expect to update?

m6jones:
So my courses doesn't require it but the `npm install` would fail because of the input version conflict. Increasing the v1.0.0 made it work. It couldn't find the `inputs#v0.0.1`. I can try reverting the input changes in the bower lock file and see if that would work. Is that a better solution?

m6jones:
This PR is being replaced by https://github.com/Brightspace/brightspace-integration/pull/1003

PR #1021, opened 2018-10-03
Jroach/d2l navigation bottom bar

jeroach:
As per US99018: https://rally1.rallydev.com/#/238990794704d/detail/userstory/241874923132

navigation PR: https://github.com/BrightspaceUI/navigation/pull/9
platformTools PR: https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/2634/overview

FYI - CI will fail until the navigation PR is merged.


sass/navigation/main.scss
```diff
@@ -132,27 +122,23 @@
...
 }
 
-.d2l-navigation-s-main-wrapper[data-more] .d2l-navigation-s-more {
-	margin-left: 0;
-	margin-right: 0;
-}
-
 @media (max-width: $d2l-navigation-bp-mobile-menu) {
-	.d2l-navigation-s-main {
+	.d2l-navigation-bottom-bar {
```
mtjandra: If this styling is used only in the navbar, why do we have it in BSI rather than in the LMS itself? Is it so we can use the variables like `d2l-navigation-bp-mobile-menu` ?

svanherk:
Have you tested all of this with image-based nav?  I don't see any updates to it in https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/2634/overview, but it also uses a lot of this stuff

jeroach:
@svanherk I have not - taking a look now

svanherk:
Currently, the height of the `bottom-bar` is being decided by the contents (specifically https://github.com/Brightspace/brightspace-integration/pull/1021/files#diff-188d1c5cdb721a664fb8cc4dcbe0052eR17), whereas the `d2l-navigation-header` defines a set height.  I wonder if we should pull this out of BSI and also put the height in the `bottom-bar` component itself.  Once we get to the immersive nav I think we'll want to make these heights override-able, but in order to control consistency it makes sense for the heights to live in the component,  Thoughts?

jeroach:
Yeah that does make sense for the height to live in the component - I'll make the update.

jeroach:
As discussed offline, we will let the child component specify the height.

Also, the plan is to add another web component to support image based nav.

jeroach:
WIP @svanherk @mtjandra - no need to review unless you're keen to 😄 

I'll update once ready.

sass/navigation/main.scss
```diff
@@ -9,16 +9,6 @@
...
-.d2l-navigation-s-no-branding .d2l-navigation-s-main {
-	border-top: 1px solid $d2l-color-gypsum;
-	border-bottom: 1px solid $d2l-color-gypsum;
-}
-
-.d2l-navigation-s-has-branding .d2l-navigation-s-main {
-	border-top: 1px solid $d2l-color-light-galena;
-	border-bottom: 1px solid $d2l-color-light-galena;
-}
-
```
mtjandra: Shouldn't these only be removed after platformTools goes in, otherwise things will be broken temporarily

sass/navigation/main.scss
```diff
@@ -163,6 +154,6 @@
 	right: auto;
 }
 
-.d2l-navigation-main-tb .d2l-navigation-s-edit-menu {
```
mtjandra: Same here shouldn't you only remove/modify this after platformTools. This PR should just had the new additions

jeroach: We typically merge the LP PR that bumps the BSI version at the same time as the platformTools PR. I suppose it could cause a race condition, but we've done it this way in the past.

@svanherk any thoughts on this?

sass/navigation/main.scss
```diff
@@ -9,16 +9,6 @@
 	border-bottom: 1px solid $d2l-color-light-galena;
 }
 
-.d2l-navigation-s-no-branding .d2l-navigation-s-main {
```
svanherk: Where do we account for setting the different border colours now?  https://github.com/BrightspaceUI/navigation/blob/master/d2l-navigation-bottom-bar.html only has the gypsum one

jeroach: We're not - since that logic has been in master for a long time, the Daves and Edward recommended on a previous PR that we just drop it.

sass/navigation/main.scss
jeroach: I'm going to just merge the two at the same time, rather than get into any trickery - sounds like neither of you feel strongly enough to object, but let me know if I'm incorrect in that assumption.

sass/navigation/main.scss
svanherk: Looks like this is controlled by the `daylight-navbar-link-area-color` feature flag, which was enabled in 10.8.6 (Wave 5 gets 10.8.6 next week). Talked to @dlockhart, they tend to leave a flag in the wild for a month, then start the code to remove it.  We need to either maintain this functionality until the flag is removed (probably next release), or we can make a pitch to Sandra (Gaudi's TPM) about removing it early

svanherk: Posted this comment before reading yours, we probably need to ask Sandra too and then we should remove the flag (in either this PR or another one) while we have the context

jeroach: @svanherk @mtjandra I've reached out to Sandra - if she's okay with removing the check, then I'll update the `platformTools` PR according to remove checking the flag.

If she wants it supported for a month, then I'll add back the style in the `BSI` PR.

Regardless only one of the PRs should have to change (regardless of the outcome).

jeroach: Going to add the style back here - and I lied - there is a small change to `navigation`

Will put up a small PR for that. `platformTools` PR is fine as is.

jeroach: 🤦‍♂️ well I lied again....I think all 3 repos will need a small change.

Working on it and will update when done.

jeroach: @svanherk @mtjandra @CarolBusman 
navigation: https://github.com/BrightspaceUI/navigation/pull/19
platformTools: https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/2634/commits/1b4f67b3078d65a5f96efd12593c00dc26c91aab

Also going to need to release a new version of `navigation` and update the version here.

jeroach:
Navigation PR merged, version bumped here @svanherk @CarolBusman @mtjandra

sass/navigation/main.scss
```diff
@@ -163,6 +167,6 @@
 	right: auto;
 }
 
-.d2l-navigation-main-tb .d2l-navigation-s-edit-menu {
-	top: 8.5px;
+.d2l-navigation-bottom-bar .d2l-navigation-s-edit-menu {
```
svanherk: This won't be applied in shadow DOM.  It works in shady DOM because the custom element name is copied to all children components as a class, in order to mimic a shadow DOM.

Shady:
image

Shadow:
image

I'm thinking what you want is just `d2l-navigation-bottom-bar .d2l-navigation-s-edit-menu`.

sass/navigation/main.scss
```diff
@@ -9,12 +9,15 @@
 	border-bottom: 1px solid $d2l-color-light-galena;
 }
 
-.d2l-navigation-s-no-branding .d2l-navigation-s-main {
+// remove this block (and m_daylightLinkAreaColorSwitch check in NavigationView.Renderer.cs ) in Nov 2018
+// add this style to the :host block in d2l-navigation-bottom-bar.html
```
svanherk: Don't we want the other one to be added to `d2l-navigation-bottom-bar.html`?  `d2l-navigation-s-has-branding` is the ON state for the feature flag.

sass/navigation/main-image-based.scss
```diff
@@ -12,6 +12,7 @@
 }
 
 .d2l-navigation-s .d2l-navigation-main-ib {
+	border-top: 1px solid $d2l-color-gypsum;
```
svanherk: Is this also controlled by the feature flag and needs two different versions here?

sass/navigation/main.scss
jeroach: Hmmm, I'm confused. The previous styling that was always being applied was the `gypsum` as per this merged PR:
https://github.com/BrightspaceUI/navigation/pull/19/files

But what you're saying makes sense - @edwardg8 can you please confirm which colour should be used when `m_daylightLinkAreaColorSwitch` returns `true` / `ON`? Thanks!

sass/navigation/main.scss
jeroach: Thanks, will fix.

sass/navigation/main-image-based.scss
jeroach: Yep, good catch.

sass/navigation/main.scss
jeroach: @svanherk I'll move the comment for now (since I think you're right) - but will wait for @edwardg8 to confirm before merging.

edwardg8: `--d2l-color-gypsum` is the old color, new color (after feature flag is enabled) is `galena`.

edwardg8:
I wish we could get rid of the feature flag before we move navbar outside LMS to WC, so we don't have to track two different states. Nevertheless, `gypsum` is the old color and `galena` is new color (after FF is enabled).

jeroach:
Thanks @edwardg8 !

PR #1039, opened 2018-10-12
build: add script to normalize bower.json and a test to make sure it …

omsmith:
…is run


omsmith:
Just trying to make it easier to maybe avoid merge conflicts (remove `lastUpdated`) and to stop caring about where `Sortable` ends up.

It would probably be desirable for `"name": "brightspace-integration"` to remain at the top, but... effort?

PR #1045, opened 2018-10-15
DE31707 - Fix long course names in 10.8.7

svanherk:
The BSI changes needed to fix long course names in 10.8.7.

Related to this PR: https://github.com/BrightspaceUI/navigation/pull/14
Will also update `d2l-navigation` here once the above PR merges.

### DIFF
--- a/sass/navigation/header.scss
+++ b/sass/navigation/header.scss
@@ -10,10 +10,16 @@
 	}
 }
 
+.d2l-navigation-header-left {
+	margin: -7px; // Needed for the overflow of the logo button highlight to expand past the left slot boundaries
+	overflow: hidden; // Needed for long course names
+}
+
 .d2l-navigation-s-header-open-button-wrapper {
 	display: inline-block;
 	flex: 0 1 auto;
 	height: 100%;
+	padding: 0 7px;
 	@media (min-width: ($d2l-navigation-bp-mobile-menu + 1)) {
 		display: none;
 	}
@@ -24,7 +30,6 @@
 	display: flex;
 	flex: 0 1 auto;
 	height: 100%;
-	margin: -7px;
 	overflow: hidden;
 	padding: 0 7px;
 
--- a/sass/navigation/logo.scss
+++ b/sass/navigation/logo.scss
@@ -2,10 +2,14 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
 
 .d2l-navigation-s-logo {
-	align-items: center;
-	display: inline-flex;
+	display: table;
 	height: 100%;
 	position: relative;
+}
+
+.d2l-navigation-s-logo-wrapper {
+	display: table-cell;
+	vertical-align: middle;
 }
 
 .d2l-navigation-s-logo-wrapper > a {
--- a/sass/navigation/mobile-menu.scss
+++ b/sass/navigation/mobile-menu.scss
@@ -166,6 +166,7 @@
 
 .d2l-navigation-s-mobile-menu-header .d2l-navigation-s-header-logo-area {
 	flex-grow: 1;
+	margin: -7px;
 
 	.d2l-navigation-s-logo-divider {
 		display: inline-block;		


sass/navigation/header.scss
```diff
@@ -24,7 +30,6 @@
 	display: flex;
 	flex: 0 1 auto;
 	height: 100%;
-	margin: -7px;
```
svanherk: Need to move this to the slot, otherwise it doesn't allow the button highlight to flow outside the slot's boundaries.

sass/navigation/header.scss
```diff
@@ -10,10 +10,16 @@
...
+.d2l-navigation-header-left {
+	margin: -7px; // Needed for the overflow of the logo button highlight to expand past the left slot boundaries
+	overflow: hidden; // Needed for long course names
+}
+
 .d2l-navigation-s-header-open-button-wrapper {
 	display: inline-block;
 	flex: 0 1 auto;
 	height: 100%;
+	padding: 0 7px;
```
svanherk: This is needed now because of the `overflow: hidden` on the slot, before it could overflow fine.

sass/navigation/logo.scss
```diff
@@ -2,12 +2,16 @@
 @import '../../bower_components/d2l-colors/d2l-colors.scss';
```
svanherk: The changes in this file are fixing the logo in IE11:
image

This is not a regression, it's currently broken in 10.8.6 as well and I'm fixing it while I'm here.  IE11 really doesn't like flex...

sass/navigation/mobile-menu.scss
```diff
@@ -166,6 +166,7 @@
 
 .d2l-navigation-s-mobile-menu-header .d2l-navigation-s-header-logo-area {
 	flex-grow: 1;
+	margin: -7px;
```
svanherk: We need to add this manually here to compensate for https://github.com/Brightspace/brightspace-integration/pull/1045/files#diff-d0ad1088eec01096ac65d0d0f57f2969R22, but the thing we _should_ do is have the mobile menu also use the `d2l-navigation-header` web component as well.  Not worrying about that for cert though.

sass/navigation/logo.scss
mtjandra: I guess this fix would also go to the main branch?

svanherk: Yeah, once I've got cert in I'll cherry pick these commits and put them in main BSI as well

sass/navigation/logo.scss
```diff
@@ -32,3 +36,7 @@
 .d2l-navigation-s-logo-link:focus {
 	box-shadow: 0 0 3px 0 rgba( $d2l-color-celestine, 0.4 );
 }
+
+.d2l-navigation-s-no-login .d2l-navigation-s-logo {
```
svanherk: Annoyed I had to add this, but I can't see another BSI-only way to fix this.  We'll need to add a custom class to avoid applying the margin hack on login pages, or we can wrap the login page logo in the `d2l-navigation-s-header-logo-area` class like the others.  For now this isn't horrible, confirmed with @edwardg8  `d2l-navigation-s-no-login` is only applied to the login page.

svanherk:
> Do you need to bump version of navigation first? BrightspaceUI/navigation#14 (comment) and release at same time?
(Release in a different branch)

Yup, although I'm just doing a patch release against master since all the changes that have gone in since the 10.8.7 branching are related to Carol's buttons, which aren't in the LE yet.  In the future we'd need to be more careful though

bower-locker.bower.json
```diff
@@ -22,7 +22,7 @@
     "d2l-loading-spinner": "^6.0.3",
     "d2l-menu": "^1.1.0",
     "d2l-more-less": "^4.3.0",
-    "d2l-navigation": "^0.7.0",
+    "d2l-navigation": "^0.8.1",
```
svanherk: Not concerned with the changes from 0.8.0 getting into cert, it was all updates to components not being used by the monolith yet: https://github.com/BrightspaceUI/navigation/compare/v0.7.0...v0.8.0

sass/navigation/logo.scss
dlockhart: I think the `no-login` variant would be rendered on any page where the user wasn't logged in? e.g. error pages, the system check and most notably the login page.

svanherk: Just checked, and yes I also see it on /d2l/lp/forgotPassword/forgotPassword.d2l and /d2l/systemcheck.  Which is also what I wanted, I just need custom code when the logo is not hoverable (does not have that special highlight with the overflow).  Which _I believe_ is all cases when you are not logged in

svanherk: I'll poke around some other cases I know of though, like self reg.

PR #1058, opened 2018-10-25
Adding styling for the new d2l-navigation-button so that the icon changes colour on hover and focus.

CarolBusman:
`d2l-icon` doesn't pick up the parent button's `color` changes on `:hover` or `:focus` by default. Changing the `color` in the personal menu to inherit when it's in a `d2l-navigation-button`.

### DIFF
--- a/sass/navigation/personal-menu.scss
+++ b/sass/navigation/personal-menu.scss
@@ -69,6 +69,11 @@
 	width: 24px;
 }
 
+d2l-navigation-button button:focus .d2l-navigation-s-personal-menu-wrapper d2l-icon,
+d2l-navigation-button button:hover .d2l-navigation-s-personal-menu-wrapper d2l-icon {
+	color: inherit;
+}
+
 .d2l-navigation-s-personal-menu-text-wrapper {
     display: flex;
     flex-direction: column;


svanherk:
So I think this is just about to change, @mtjandra is turning on the profile-image flag for dev and it will start to use his web component instead of just an icon.  You won't be able to target inside it the same way with shadow DOM, so the `profile-image` may need to be changed in order to take a hover state or make the color configurable by the parent with a css variable.  Maybe it already does that, otherwise it might currently be broken...

mtjandra:
Lets still target both. It is still behind a feature flag, so it should be able to work correctly if the flag is either on or off

jeroach:
@CarolBusman @svanherk @mtjandra Just want to make sure this PR isn't forgotten about (it's been idle for a while) - I don't have the context of how Miles' changes affect this PR, but are these changes good as is, or do they need to take Miles' work into account?

CarolBusman:
> @CarolBusman @svanherk @mtjandra Just want to make sure this PR isn't forgotten about (it's been idle for a while) - I don't have the context of how Miles' changes affect this PR, but are these changes good as is, or do they need to take Miles' work into account?

This is not viable in its current form as it doesn't work correctly in both shady and shadow. I'm working through getting the css right to have this working properly in both.

PR #1066, opened 2018-10-30
Jroach/d2l navigation

jeroach:
~~Still need to reference new version of `navigation` once ready. Will update here once https://github.com/BrightspaceUI/navigation/pull/22 is approved and merged.~~

Ready to go 😄


mtjandra:
Looks like the only update here is https://github.com/BrightspaceUI/navigation/commit/7f68b65044e3b4dd81f43bafab85eb0cc756e4f3

I'm guessing there is currently a bug in the LMS and this will fix it? Unsure if it needs to be linked with any other change

jeroach:
@mtjandra Yep that's the change. Also added in some comments to help preserve knowledge for future work

PR #1079, opened 2018-11-15
Updates to enrollment card, tooltips, card and my courses.

m6jones:


bower-locker.bower.json
```diff
@@ -8,7 +8,7 @@
     "d2l-colors": "^3.1.2",
     "d2l-course-image": "Brightspace/course-image#^2.1.5",
     "d2l-dropdown": "^6.6.1",
-    "d2l-enrollments": "BrightspaceHypermediaComponents/enrollments#^0.3.20",
+    "d2l-enrollments": "BrightspaceHypermediaComponents/enrollments#^0.3.21",
```
dbatiste: Only concern is if the `fit-content` (used enrollment-card for the tooltip width) is problematic in Edge and IE11.  It would be good to confirm whether that's an issue before proceeding with the rest of the updates.

PR #1104, opened 2018-11-29
[Cert] Update d2l-activity-alignments and d2l-outcomes-level-of-achievements

mpharoah-d2l:


bower.json
```diff
@@ -39,7 +39,7 @@
       "d2l-opt-in-flyout-webcomponent": "1.1.0",
       "d2l-organization-hm-behavior": "2.0.3",
       "d2l-organizations": "0.1.14-hybrid",
-      "d2l-outcomes-level-of-achievements": "1.0.10",
+      "d2l-outcomes-level-of-achievements": "1.0.12",
```
mpharoah-d2l: The BSI had version 10 here, but the version of `d2l-activity-alignments` it was using was pulling in version 11, and that's the only place its actually used, so this is really only going up 1 version, the only changes being the ones we want to bring into cert

sarahboyd:
This is for the accessibility cert defect right? we should wait for approval before we get this in 👍

PR #1110, opened 2018-12-03
update rubric to 0.141.0

mdulat:


emil-furniss:
You can run `npm run clean-bower-json` to bring the json back to the original format, so it;'s easier to see changes

PR #1117, opened 2018-12-06
fix rubric dependencies

mdgbayly:


bower-locker.bower.json
```diff
@@ -16,7 +16,7 @@
     "d2l-fetch-simple-cache": "Brightspace/d2l-fetch-simple-cache#^0.6.0",
     "d2l-icons": "^5.1.0",
     "d2l-image-banner-overlay": "Brightspace/d2l-image-banner-overlay#^3.3.2",
-    "d2l-inputs": "BrightspaceUI/inputs#^1.0.3",
+    "d2l-inputs": "BrightspaceUI/inputs#^1.0.5",
```
jeroach: If we're updating inputs, I would expect an update in the `bower.json` file as well to reference the new commit hash.

mdgbayly: It already had it, so this was just fixing up the locker file to match. If you install from the locked bower json or do a regular install it will pick up new sem ver dependencies anyway.

PR #1129, opened 2018-12-12
Queue for callbacks interested in being called after WCR

dlockhart:
I don't intend to merge this until after branching on Friday.

Currently we have a bunch of code waiting for the `WebComponentsReady` event, mostly to defer but sometimes to ensure that all the polyfills are there before using them. These places have ugly bits because `WebComponentsReady` may have already fired, so they need to check `d2lWCRDispatched` and conditionally add a listener or execute right away. That sucks.

Additionally, in Polymer 3 our web component bundles are going to load asynchronously. So just because `WebComponentsReady` has fired doesn't mean that our bundles have loaded, only that the polyfills are there. So we need a nice mechanism that knows when both have occurred in Polymer 3 land.

Enter: this pull request.

It's loosely modelled after [`WebComponents.waitFor`](https://github.com/webcomponents/webcomponentsjs/blob/92063e0956ad0d5ed1f527026029cc2f94977340/webcomponents-loader.js#L95):
```
D2L.OnWebComponentsReady(function() {
    // I'm confident WebComponentsReady has fired!
});
```
Essentially it keeps an in-order queue of callbacks and calls them all as soon as `WebComponentsReady` fires. If more things get added to the queue after that, they'll execute immediately.

When I bring this into Polymer 3, I'll add an additional check to make sure our core bundle has loaded.

### DIFF
--- a/js/d2l-web-components-ready.js
+++ b/js/d2l-web-components-ready.js
@@ -1,0 +1,35 @@
+var ready;
+
+function dcl() {
+	window.removeEventListener('DOMContentLoaded', dcl);
+	ready();
+}
+
+function wcr() {
+	window.removeEventListener('WebComponentsReady', wcr);
+	ready();
+}
+
+module.exports = {
+	WebComponentsReady: new Promise(function(resolve) {
+		ready = resolve;
+	}),
+	init: function() {
+		if (window.WebComponents && !window.WebComponents.ready) {
+			window.addEventListener('WebComponentsReady', wcr);
+		} else {
+			if (document.readyState === 'interactive' || document.readyState === 'complete') {
+				ready();
+			} else {
+				window.addEventListener('DOMContentLoaded', dcl);
+			}
+		}
+	},
+	reset: function() {
+		window.removeEventListener('WebComponentsReady', wcr);
+		window.removeEventListener('DOMContentLoaded', dcl);
+		this.WebComponentsReady = new Promise(function(resolve) {
+			ready = resolve;
+		}.bind(this));
+	}
+};
--- a/js/d2l-web-components-ready.js
+++ b/js/d2l-web-components-ready.js
@@ -30,6 +30,6 @@ module.exports = {
 		window.removeEventListener('DOMContentLoaded', dcl);
 		this.WebComponentsReady = new Promise(function(resolve) {
 			ready = resolve;
-		}.bind(this));
+		});
 	}
 };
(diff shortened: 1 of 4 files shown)


js/d2l-web-components-ready.js
```diff
@@ -0,0 +1,54 @@
...
+			window.addEventListener('WebComponentsReady', wcr);
+		} else {
+			if (document.readyState === 'interactive' || document.readyState === 'complete') {
+				readyFunc();
+			} else {
+				window.addEventListener('DOMContentLoaded', dcl);
+			}
+		}
+	},
+	reset: function() {
```
dlockhart: `init()` and `reset()` exist purely for testing purposes -- so I can choose when to run the initialization code and can reset things back to their initial state in between tests.

omsmith:
Consider something like this? Less logic and error handling?:

```js
module.exports = {

  WebComponentsReady: null,
  _ready: null,
  _cancel: null,

  init: function() {
    var self = this;

    self.WebComponentsReady = new Promise(function(resolve, reject) {
      self._ready = resolve;
      self._cancel = reject;
    });

    addEventListener('WebComponentsReady', function() {
      self._ready();
    });
  },

  reset: function() {
    self._cancel(new Error('D2L.WebComponentsReady was reset'));
    self.init();
  }
};
```

```
D2L.WebComponentsReady.then(function() {
    // I'm confident WebComponentsReady has fired!
});
```

dlockhart:
I do like the Promise/await based API better, I'll give it a shot!

dlockhart:
I've updated this to be Promise-based, thanks for the suggestion @omsmith

test/web-components-ready.js
```diff
@@ -0,0 +1,167 @@
...
+					.should.have.been.calledWith('DOMContentLoaded');
+			});
+
+			it('should execute callbacks when "DOMContentLoaded" fires', (done) => {
+				wcr.WebComponentsReady.then(done);
+				wcr.init();
+				dispatchEventListener();
+			});
+
+			it('should execute callbacks in order', (done) => {
```
omsmith: Is this a super-duper desired property that we want to test it specifically and not just "whatever Promises do"?

dlockhart: That's a good point. I guess in-order isn't _that_ important... although it's nice. But if someone needed order guaranteed they should chain their promises. I'll modify the test to just make sure they all get called.

PR #1135, opened 2018-12-19
not resolving WebComponentsReady promise until imports have loaded

dlockhart:
Applying the same treatment as we use in Polymer 3-land of waiting for both the `WebComponentsReady` event AND the end of the imports file to be reached.

People are noticing JS errors on Spark, mostly involving `d2lIntl` not being defined. It's a timing issue that's very tricky to reproduce. I was able to get Safari to exhibit it if I loaded the page and then hit back very quickly. Under this circumstance, `window.WebComponents` doesn't exist for some reason -- even though the polyfill file is first and loaded synchronously. Then, when this script runs it adds an event listener for `DOMContentLoaded` instead of `WebComponentsReady`, which subsequently fires _before_ all the imports (including `d2l-intl`) have loaded.

### DIFF
--- a/web-components/bsi.html
+++ b/web-components/bsi.html
@@ -59,3 +59,10 @@
 <link rel="import" href="../bower_components/d2l-navigation/d2l-navigation-separator.html">
 <link rel="import" href="../bower_components/d2l-users/all-imports.html">
 <link rel="import" href="navigation-icons.html">
+<script>
+	window.D2L = window.D2L || {};
+	window.D2L._webComponentsLoaded = true;
+	if (window.D2L.WebComponentsLoaded !== undefined) {
+		window.D2L.WebComponentsLoaded();
+	}
+</script>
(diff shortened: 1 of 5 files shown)


web-components/bsi.html
```diff
@@ -59,3 +59,10 @@
 <link rel="import" href="../bower_components/d2l-navigation/d2l-navigation-separator.html">
 <link rel="import" href="../bower_components/d2l-users/all-imports.html">
 <link rel="import" href="navigation-icons.html">
+<script>
+	window.D2L = window.D2L || {};
+	window.D2L._webComponentsLoaded = true;
```
dlockhart: This is needed for Chrome, since it actually loads the imports _before_ `bsi.js` because it supports them natively.

dlockhart:
Going to merge this to get all the subsequent PRs going as this is actively breaking Spark. Happy to discuss and make changes afterwards though.

PR #1144, opened 2019-01-02
Update d2l-navigation to get immersive nav changes

svanherk:
Making the immersive navigation fixed to the top of the screen.


bower.json
```diff
@@ -35,13 +35,13 @@
       "d2l-more-less": "4.3.1",
       "d2l-my-courses": "6.0.8",
       "d2l-my-dashboards": "1.0.6",
-      "d2l-navigation": "1.1.0",
+      "d2l-navigation": "2.0.0",
       "d2l-offscreen": "3.0.6",
-      "d2l-opt-in-flyout-webcomponent": "1.2.0",
+      "d2l-opt-in-flyout-webcomponent": "1.2.1",
```
svanherk: Pre Polymer 3 conversion changes

bower.json
```diff
@@ -35,13 +35,13 @@
...
       "d2l-organization-hm-behavior": "2.0.5",
       "d2l-organizations": "0.1.14-hybrid",
       "d2l-outcomes-level-of-achievements": "1.0.13",
-      "d2l-polymer-behaviors": "1.8.2",
+      "d2l-polymer-behaviors": "1.8.4",
```
svanherk: Fixing polymer lint errors ahead of the Polymer 3 conversion

bower.json
```diff
@@ -109,7 +109,7 @@
     "d2l-fetch-simple-cache": "https://github.com/Brightspace/d2l-fetch-simple-cache.git#a7e1b7ae02962639ffcf4af222af540203f35b59",
     "d2l-fetch-siren-entity-behavior": "https://github.com/Brightspace/d2l-fetch-siren-entity-behavior.git#a5cf7958165ca9f5f16bdda9e200b4a071ea66f5",
     "d2l-hierarchical-view": "https://github.com/BrightspaceUI/hierarchical-view.git#c0952ac31024c6098e25bc0fc61be79a048539f1",
-    "d2l-hm-constants-behavior": "https://github.com/Brightspace/d2l-hm-constants-behavior.git#0a64e279952d3e4adaaa5ac514cb6434cba34da6",
```
svanherk: Missed this, is there a way to stop this from flip-flopping?

dlockhart: Does running `npm run clean-bower-json` fix it?

svanherk: I did run that, rerunning it has no effect.

dlockhart: Hmm. I wonder if it's easy to update it to fix this inconsistency as well?

svanherk: I think it happens if the `url` in the `package.json` has `git+` at the beginning (https://github.com/Brightspace/d2l-hypermedia-constants/blob/master/package.json#L10, https://github.com/Brightspace/simple-overlay/blob/master/package.json#L13, etc).  Ones that don't have this (https://github.com/BrightspaceUI/status-indicator/blob/master/package.json, https://github.com/Brightspace/d2l-rubric/blob/master/package.json#L25, etc) use show up as just `https://...`.  Unless these are different for a reason, I don't mind going through and updating them all to be consistent.

dlockhart: _Interesting_. I can't think of a reason why these would be different, so yeah let's make 'em consistent if it's easy.

svanherk: Turns out this isn't the issue, I'm finding some examples that break this theory (like https://github.com/Brightspace/organization-hm-behavior/blob/hybrid/package.json#L17)

dlockhart: Well we only have about another month to deal with it anyway... `bower.json` is on its last legs!

PR #1154, opened 2019-01-09
Update package-lock to get latest version of activities branch

svanherk:
Checked on my local instance, I'm no longer seeing the "already registered" error

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1609,7 +1609,7 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#d7f13e5bd2672a88a3f6ff20a293ca0bda874082",
+      "version": "github:BrightspaceHypermediaComponents/activities#3af0716c8f3c9fc09de82e807e695e214903806a",
       "from": "github:BrightspaceHypermediaComponents/activities#add_evaluation-hub_skeleton_with_table",
       "dev": true,
       "requires": {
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1624,26 +1624,10 @@
...
-          "requires": {
-            "@polymer/iron-resizable-behavior": "^3.0.0-pre.18",
-            "@polymer/polymer": "^3.0.0",
-            "d2l-colors": "github:BrightspaceUI/colors#6a1c414d6f3685e3c286aff4147ede037bdbb693",
-            "d2l-icons": "github:BrightspaceUI/icons#7df2e338974029bb4e1005799d2dd88b6e83145c",
-            "fastdom": "^1.0.8",
-            "resize-observer-polyfill": "^1.5.0",
-            "stickyfilljs": "^2.1.0"
-          }
-        }
```
dlockhart: ^ this was the thing that was breaking stuff 👍

dlockhart: Which makes me wonder... maybe we could write a validator of some kind? It could go through `package-lock.json` and make sure that no `d2l-` top-level dependencies have nested `d2l-` dependencies. Hmm...

svanherk: Yeah, now that's we're moving to npm instead of bower, I think we need to define some similar steps for how to handle the lock file that we had in the README for bower locking, and write some validation checks around that.  At least while we're using semver, and not registering with npm

dlockhart: I've created `US103256` to track this.

PR #1156, opened 2019-01-09
Fix performance timings

dlockhart:
We want to use the new Promise-based `WebComponentsReady` stuff instead of the window event in both our `fastdom` wrapper and performance timing code. This is because in Polymer 3, the window event has often already fired by the time our code runs. The promise is where it's at.

But to use the promise, the performance timing code needs to move up into `index.js`, which means everything up there needs to switch over to being an ES6 module. Some updates to the tests in order to get mocha working with the modules and the new promise followed accordingly.

Once this merges, I'll be applying this pretty much as-is to the Polymer 3 branch.


js/index.js
```diff
@@ -1,8 +1,13 @@
 import './polyfills.js';
+import webComponentsReady from './d2l-web-components-ready.js';
+import FastDom from './d2l-fastdom.js';
+import './timing-debug.js';
+import './performance-timings.js';
+import '../bower_components/jquery-vui-change-tracking/changeTracker.js';
+import '../bower_components/jquery-vui-change-tracking/changeTracking.js';
```
dlockhart: Imports always end up at the top anyway, so this just makes it a bit clearer that the imports don't actually get mixed in with the code below. The order hasn't changed, but `performance-timings.js` got added in the middle.

PR #1163, opened 2019-01-14
replacing "define" with "define2"

dlockhart:
`define2` is the new `define`!

Problem: the `esm-amd-loader` uses a global `define` function to load the AMD modules that are created as part of `polymer build`'s conversion of ES6 modules to AMD. Unfortunately, our UMD FRAs include RequireJS, which also uses a global `define` variable.

Solution: after the build completes, go through the output and replace instances of `define` with `define2`, avoiding the conflict. I was pretty specific about what I'm replacing here, since we don't want to hit cases like `customElements.define()`.


dlockhart:
Thanks for trying it out! You found a bug for sure as I was running against the non-minified version of the esm-amd-loader while I was debugging and never switched back. Will put up a fix shortly.

I also forgot to mention that you need a 1-line change to your LMS -- [this line](http://search.dev.d2l/source/xref/Lms/lp/framework/web/D2L.LP.Web/UI/Html/Style/Default/StylePageElementWriter.cs#78) needs to write out `define2` instead of `define`.

dlockhart:
Should be fixed now. Thanks again @margaree

PR #1166, opened 2019-01-14
Fix navbar non-links

dlockhart:
Two fixes here:
1. Applying the branding colours to non-links like the course name, which end up just being inside a `<span>` and not an `<a>`
2. While verifying the fix I noticed that all the CSS in `navigation/common.scss` was getting included **8 TIMES**, since it was imported into 8 different SCSS files. It only needs to be imported once, haha. This saves ~44KB in download size, yikes!


dlockhart:
Yeah, exactly Stacey. If `common.scss` only contained variables, pulling it in many times wouldn't be a problem. But in this case, it includes CSS, so each time that gets imported all that CSS gets imported.

PR #1167, opened 2019-01-15
DE32567: Update My Courses to count total Enrollments better.

njostonehouse:


edwardg8:
I believe this also include the pinned tab changes I made, refetch when pinning a course and changing to other tabs that contains the course?

njostonehouse:
@edwardg8 Yes, although that's behind the flag? Or is it actual improvements to reloading? That might be good.

edwardg8:
Yep, I think the refetch mechanism isn't necessary just for pinned tab, so it's not behind the flag. All tabs should benefit from this.

PR #1173, opened 2019-01-18
switch bower to node and add d2l-rubric

stevenao:


dlockhart:
I fixed a few build errors -- `package-lock` needed to be updated and `eslint` wasn't configured yet to work with Polymer 3.

There's a new problem in that HTML editor is referencing `d2l-fetch-auth` without a version number, which causes multiple versions to get bundled. I have a fix up for that here:
https://github.com/Brightspace/d2l-html-editor/pull/173

PR #1174, opened 2019-01-18
removing references to "bower"

dlockhart:
This scrubs BSI of all things "Bower"! 🎉 

@stevenao had removed a bunch of this earlier as it was breaking builds, but I think it's time to actually remove it as it's not providing any value keeping the old `bower.json` around anymore.


README.md
```diff
@@ -44,44 +44,28 @@ To point your Brightspace instance at the local integration project:
...
-3. Make changes to `bower.json` manually or via `bower install <component>`
-4. Update dependencies in `bower_components` via `bower update` or by removing the directory and doing a fresh `bower install`
-5. Lock `bower.json` again by running `bower-locker lock`
-6. Run `npm run clean-bower-json` to normalize some oddities
-7. Inspect the diff to ensure the changes match your expectations
-
-[Read more in the `bower-locker` documentation...](https://github.com/infusionsoft/bower-locker)
-
-### NPM Locking
+We use a `package-lock.json` file to lock our NPM dependencies. This ensures we only pick up changes to dependencies when we explicitly ask for them and are prepared to test them.
```
svanherk: Should we mention here completely deleting `package-lock.json` before running `npm i`?

dlockhart: According to NPM's documentation, that's not actually necessary. Not sure if I've seen that in practice though...

svanherk: Yeah never seems to work for me, so I got into the habit of clearing that and `node_modules` every time.  But maybe it's better since I last tried

PR #1177, opened 2019-01-21
Add back rubrics

wongvincent:
Fixed in d2l-html-editor: https://github.com/Brightspace/d2l-html-editor/pull/175/files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1345,16 +1345,16 @@
       "dev": true
     },
     "autoprefixer": {
-      "version": "9.4.5",
-      "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-9.4.5.tgz",
-      "integrity": "sha512-M602C0ZxzFpJKqD4V6eq2j+K5CkzlhekCrcQupJmAOrPEZjWJyj/wSeo6qRSNoN6M3/9mtLPQqTTrABfReytQg==",
+      "version": "9.4.6",
+      "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-9.4.6.tgz",
+      "integrity": "sha512-Yp51mevbOEdxDUy5WjiKtpQaecqYq9OqZSL04rSoCiry7Tc5I9FEyo3bfxiTJc1DfHeKwSFCUYbBAiOQ2VGfiw==",
       "dev": true,
       "requires": {
-        "browserslist": "^4.4.0",
-        "caniuse-lite": "^1.0.30000928",
+        "browserslist": "^4.4.1",
+        "caniuse-lite": "^1.0.30000929",
         "normalize-range": "^0.1.2",
         "num2fraction": "^1.2.2",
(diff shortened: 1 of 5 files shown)


svanherk:
Should there be a `package-lock.json` change to get the latest rubrics?  Or was the fix outside the rubrics repo?

wongvincent:
Not sure why the package-lock.json npm i created add `optional: true`

svanherk:
This looks good to me!

dlockhart:
@mdgbayly @wongvincent 

Did a quick compare of `bsi.js` and `shared_bundle_1.js` with and without these changes. These changes add `482KB` (uncompressed) to those bundles.

This is what gets downloaded and executed on every page, and doesn't include what gets downloaded/executed on pages that use the rubrics and/or HTML editor web components.

That's... a lot, and accounts for about 48% of our web component bundle size. I'm pretty worried about what the impact of this might be for IE11 and Edge, who are known to have performance issues simply executing the JS that registers lots of web components.

How difficult would it be to get all the rubrics stuff into a single bundle, instead of the 3 that it's split across today? That would at least get the dependencies the 3 share out of our common shared bundle.

dlockhart:
Looks like @stevenao is working on that as we speak! 👍

dlockhart:
Cool. So mostly what would cause the increase in bundle sizes would be one (or both) of two things:

1) Dependencies that are new to BSI but are used in more than one of the rubrics bundles are now appearing in the shared bundle. This is definitely happening, and merging the editor bundles into one should fix some of this (like it'll move HTML editor out of shared). New dependencies shared across the newly combined editor and rubrics bundles will still now end up in shared and increase its size, but there's less we can do about this.

2) Dependencies that aren't new to BSI, but that were separated out into their own bundle (e.g. table, more-less, etc.) or only used inside one bundle that are used by one or more of the rubrics bundles is causing them to get moved from its own bundle into shared. Comparing all the `.js` file sizes before & after, this appears to be happening to about `16KB` of stuff from `my-courses` and `6KB` of things from the opt-in flyout.

Basically as soon as something is used twice, the Polymer bundler moves it into shared. So if rubrics just happens to be the unlucky one that uses something for the 2nd time, this happens.

As I said, I'm investigating the performance impacts of _not_ bundling at all for certain browsers, which would mean only the web components that are used on a particular page would get downloaded. But that'll mean A LOT of HTTP requests.

mdgbayly:
@dlockhart regarding the bundle size - just playing around with the pre Polymer 3 build, and I guess yeh - rubrics and rubric editor are big just by their complexity. It looks like the rubric preview/assess components are still ending up in the shared bundle. I just remembered that we now use the preview from the rubric editor (that was a fairly recent change too), so that is probably what is causing that.

So maybe there is value in merging the preview/editor bundles too. It just means that the preview pages would be loading all the editor stuff too. Possibly there are other things we can do to mitigate this, even using dynamic imports for the rubric preview stuff in the rubric editor.

PR #1178, opened 2019-01-21
[fix] Rubrics editor BSI integration is causing tinymce to be loaded …

stevenao:
…twice on pages using legacy LMS HTML editor

### DIFF
--- a/polymer.json
+++ b/polymer.json
@@ -12,8 +12,7 @@
     "web-components/d2l-scroll-spy.html",
     "web-components/d2l-table.html",
     "web-components/d2l-rubric.html",
-    "web-components/d2l-rubric-editor.html",
-    "web-components/d2l-rubric-structure-editor.html"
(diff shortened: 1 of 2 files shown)


mdgbayly:
> Should `d2l-rubric-editor.html` not need to be modified to include `d2l-rubric/editor/d2l-rubric-structure-editor.html`, or was that file not ever needed?

It should be being included already via the `d2l-rubric/editor/d2l-rubric-editor.html` in the `d2l-rubric` repo.

stevenao:
yes, I verified that in the bsi build folder d2l-rubric-editor.html, d2l-rubric-structure-editor is defined there.

stevenao:
we will have a smaller shared_bundle_1.html as d2l-html-editor is not included as a shared bundle anymore. It will be imported with the d2l-rubric-editor bsi import.

PR #1186, opened 2019-01-24
bump outcomes polymer 3 components

djwester:
Hopefully I am doing this correctly?


web-components/outcomes-level-of-achievement-ui.js
```diff
@@ -0,0 +1 @@
+import "outcomes-level-of-achievement-ui/d2l-outcomes-level-of-achievement.js"
```
svanherk: This file name needs to change to `d2l-outcomes-level-of-achievements.js`, to match hat's in `polymer.json`

web-components/d2l-activity-alignments.js
```diff
@@ -1,4 +1,2 @@
-/*
-<link rel="import" href="../bower_components/d2l-activity-alignments/d2l-activity-alignments.html">
-<link rel="import" href="../bower_components/d2l-activity-alignments/d2l-select-outcomes.html">
-*/
+import "d2l-activity-alignments/d2l-select-outcomes.js";
+import "d2l-activity-alignments/d2l-activity-alignments.js";
```
dlockhart: Nit: strings in JS should be single-quoted. This may fail the linter actually.

PR #1188, opened 2019-01-30
update my-courses version in bsi

ciss1995:
d2l-organizations and d2l-enrollments have already been updated from a previous PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1133,9 +1133,9 @@
       "dev": true
     },
     "ansi-escapes": {
-      "version": "3.1.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1872,9 +1872,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30000930",
```
njostonehouse: I think a bunch of these version bumps were unintentional. You probably don't want to include them unless you're sure we want them pulled in.
In bower, I would manually update the locked file to the previous version, then do a clean build to make sure everything continues to build properly.

PR #1199, opened 2019-02-04
Update dependencies.

dbatiste:


dbatiste:
@awikkerink : I noticed your change earlier switched from SHAs to semver versions in the package-lock file.  Was that intentional?  This change updates a couple of versions, but also switches back to SHAs.  Is this ok?

awikkerink:
I think the switch the versions happened in a newer version `npm`. I'll need to look into it. But switching to SHAs LGTM

dbatiste:
Thanks.  Currently I am on `node v9.3.0 (npm v6.4.1)`.

PR #1210, opened 2019-02-07
removing "d2l-navigation-s-no-branding"

dlockhart:
The code that put this CSS class on things [was removed](http://search.dev.d2l/source/diff/Lms/platformTools/navbars/D2L.PlatformTools.Navbars/Web/Desktop/Rendering/Views/Daylight/NavigationView.Renderer.cs?r2=%2FLms%2FplatformTools%2Fnavbars%2FD2L.PlatformTools.Navbars%2FWeb%2FDesktop%2FRendering%2FViews%2FDaylight%2FNavigationView.Renderer.cs%4084d0612c&r1=%2FLms%2FplatformTools%2Fnavbars%2FD2L.PlatformTools.Navbars%2FWeb%2FDesktop%2FRendering%2FViews%2FDaylight%2FNavigationView.Renderer.cs%4002c9f016), so there's no need for any CSS to target it.

I'll do a separate PR that also removes `d2l-navigation-s-has-branding` since it's now always present and so also doesn't need to be called out.

### DIFF
--- a/sass/navigation/main.scss
+++ b/sass/navigation/main.scss
@@ -10,12 +10,6 @@
 }
 
 // remove this block (and m_daylightLinkAreaColorSwitch check in NavigationView.Renderer.cs ) in Nov 2018
-.d2l-navigation-s-no-branding .d2l-navigation-main-footer-container {
-	border-top: 1px solid $d2l-color-gypsum;
-	border-bottom: 1px solid $d2l-color-gypsum;
-}
-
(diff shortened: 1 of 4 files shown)


sass/navigation/main.scss
```diff
@@ -9,12 +9,6 @@
 	border-bottom: 1px solid $d2l-color-light-galena;
 }
 
-// remove this block (and m_daylightLinkAreaColorSwitch check in NavigationView.Renderer.cs ) in Nov 2018
```
svanherk: Lol it hasn't been _that_ long since then... success!

PR #1211, opened 2019-02-08
Fix nav footer styles

dlockhart:
This picks up [this change to navigation](https://github.com/BrightspaceUI/navigation/pull/58) and goes along with [this PR in platform tools](https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/2881/overview), which moves the footer's border styles from BSI into the navigation component.


sass/navigation/main.scss
```diff
@@ -2,17 +2,6 @@
 @import '../common.scss';
 
 .d2l-navigation-s-main-no-links {
```
svanherk: Is this `border-bottom` still needed?  Trying to track down the purpose of `d2l-navigation-s-main-no-links`...

dlockhart: I'll confirm. I think the purpose of it is for when the navbar has no links in it, we don't render that bottom link area. And in that case, we need to apply a bottom border to the top area since the bottom area defines both the top and bottom borders,.

dlockhart: Confirmed that it's still needed for that case where there are no links:

screen shot 2019-02-08 at 10 38 44 am

svanherk: Thanks!  Makes sense!

PR #1216, opened 2019-02-13
US103342: add css style for Dashboard background type

johngwilkinson:


dlockhart:
Can you put this into `master` first and then merge into the `polymer-3` branch? That's how we've been keeping them in sync.

PR #1218, opened 2019-02-13
Us103342/polymer3/background type

johngwilkinson:


sass/page.scss
```diff
@@ -61,6 +61,10 @@ body[unresolved] {
 	padding-bottom: 10px;
 }
 
+.d2l-page-dashboard {
+	background-color: #f6f7f8;
```
jwong-d2l: I'm not sure if your designer has talked to anyone about this, but I think as a whole they wanted to move to a named colour, specifically regolith.

johngwilkinson: Regolith seems to be #f9fafb, which doesn't match what's on the homepage now or the design I was given (both have #f6f7f8). I'll double-check with Fiona.

dlockhart: When this came up last time (with Joseph), the designers involved did express an interest in changing it. No matter what they decide, at least this way it's in one place.

johngwilkinson: Updated to use regolith

PR #1222, opened 2019-02-15
Increase activities version

svanherk:


svanherk:
FYI to anyone looking at this - the plan is to wait until after branching.  Let me know if something in here needed to be this release.

jwong-d2l:
We're going to have to bump BSI in cert for our changes... is that okay?  What was the reason for waiting till after branching?

svanherk:
Mostly so there was no confusion in the naming, if I name it 10.8.11 it probably wouldn't get in on time, so I was gonna name it 20.19.4 to make it clear what release it got in

jwong-d2l:
so wait, we'll have to do two different releases of BSI to support our cert bypass work?

dlockhart:
`10.8.11` needs to handle the possibility that the Polymer 3 feature flag might be disabled in production.

svanherk:
Closing, we will do a fresh pull after branching

PR #1224, opened 2019-02-15
bump version of d2l-save-status

dabdullahd2l:


dlockhart:
If you want to get this into `10.8.11`, you'll want to coordinate with @jwong-d2l and @dbatiste.

package.json
```diff
@@ -92,7 +92,7 @@
     "d2l-outcomes-level-of-achievements": "Brightspace/outcomes-level-of-achievement-ui.git#semver:^2",
     "d2l-polymer-behaviors": "Brightspace/d2l-polymer-behaviors-ui#semver:^2",
     "d2l-rubric": "Brightspace/d2l-rubric#semver:^3",
-    "d2l-save-status": "Brightspace/d2l-save-status#semver:^3",
+    "d2l-save-status": "Brightspace/d2l-save-status#semver:^3.1",
```
dlockhart: You want to lose the `.1`, otherwise other components that don't reference it exactly like this will end up with copies.

dabdullahd2l: It did not pick up the latest release until I added the `1`. Should I still remove it?

package-lock.json
```diff
@@ -2939,7 +2939,7 @@
         "d2l-menu": "github:BrightspaceUI/menu#semver:^2",
         "d2l-offscreen": "github:BrightspaceUI/offscreen#semver:^4",
         "d2l-polymer-siren-behaviors": "github:Brightspace/polymer-siren-behaviors#semver:^1",
-        "d2l-save-status": "github:Brightspace/d2l-save-status#semver:^3",
+        "d2l-save-status": "github:Brightspace/d2l-save-status#1849a3b5ccd1b8109ad8f6db8bfd76dd5d685711",
```
dlockhart: Make sure you have the latest NPM version. The newer ones I think don't make this change. `npm install -g npm`.

package.json
dlockhart: It shouldn't need the `.1` since `^3` semver means all minor and patch versions should be picked up. Something else must be going on. I'd recommend `rm package-lock.json; rm -rf node_modules; npm install;`.

dlockhart:
@dabdullahd2l I'm going to close your PR since #1223 includes it already.

PR #1230, opened 2019-02-20
use activities version with vanity url fix

vix633:
use activities version with vanity url fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,22 +31,22 @@
       }
     },
     "@babel/core": {
-      "version": "7.2.2",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.2.2.tgz",
-      "integrity": "sha512-59vB0RWt09cAct5EIe58+NzGP4TFSD3Bz//2/ELy3ZeTeKF6VTD1AXlH8BGGbCX0PuobZBsIzO7IAI9PH67eKw==",
+      "version": "7.3.3",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.3.3.tgz",
+      "integrity": "sha512-w445QGI2qd0E0GlSnq6huRZWPMmQGCp5gd5ZWS4hagn0EiwzxD5QMFkpchyusAyVC1n27OKXzQ0/88aVU9n4xQ==",
       "dev": true,
       "requires": {
         "@babel/code-frame": "^7.0.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -2837,18 +2818,28 @@
       }
     },
     "d2l-opt-in-flyout-webcomponent": {
-      "version": "github:Brightspace/d2l-opt-in-flyout-webcomponent#48797829b99359b3442afd7ddb8128495ac9d912",
+      "version": "github:Brightspace/d2l-opt-in-flyout-webcomponent#242ee167fda31688e1e932d1035b906a445b1414",
```
svanherk: @wongvincent This seems to be using the latest release, but there's a build error still

wongvincent: It looks like `s-html` is being nested and being included multiple times inside components.
In `d2l-opt-in-flyout-webcomponent`, it's using `s-html#semver:^2`, some other places are using `s-html#semver:^2.0.0`. The `dependency-map.txt` file in polymer-3-converter has it listed as `semver:^2.0.0` so that's probably why there is a discrepency.

Quoting DLockhart from Slack:
> The special git `semver:^1.0.0` syntax we’re using will result in multiple copies of a dependency for each different pattern. So if there was `semver:^1.0.0` and `semver:^1.0.1` you’d have two copies.

> The solution we’ve been applying today to everything is to reference things without the minor or patch versions: `semver:^1`.

I'm in the process of fixing this. Hopefully, that resolves the issue.

jwong-d2l: s-html merged

svanherk: Hmm, I'm still seeing an issue with a fresh build here: https://github.com/Brightspace/brightspace-integration/pull/1235

svanherk: Maybe https://github.com/Brightspace/brightspace-integration/blob/07d64d95d2e911390d556a34172e0fcc37c7589c/package-lock.json#L2403?  Updates need in `d2l-activity-alignments`?

wongvincent: yep, was waiting for an approval and bump, which is done now!

BSI fresh build: https://github.com/Brightspace/brightspace-integration/pull/1233

svanherk: Looks like I wrote this message right as you were doing so.  Ignore me!

svanherk:
Changes made in https://github.com/Brightspace/brightspace-integration/pull/1235

PR #1233, opened 2019-02-20
Update package-lock with latest

wongvincent:


svanherk:
Looks like all of the `semver` stuff is turning back to commit SHAs, I think you need to update your version of `npm`

PR #1234, opened 2019-02-21
integrating fastdom's promise extension

dlockhart:
This will let us do things like:

```
D2L.fastdom.measure(() => {
    return document.body.offsetHeight;
}).then((bodyHeight) => {
    // use bodyHeight somehow
});
```

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -1,6 +1,7 @@
 import './polyfills.js';
 import webComponentsReady from './d2l-web-components-ready.js';
 import FastDom from './d2l-fastdom.js';
+import FastDomPromised from '../node_modules/fastdom/extensions/fastdom-promised.js';
 import './timing-debug.js';
 import './performance-timings.js';
 import '../node_modules/jquery-vui-change-tracking/changeTracker.js';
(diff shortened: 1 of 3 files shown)


js/index.js
```diff
@@ -19,3 +20,6 @@ if (window.d2lWCRDispatched) {
 }
 
 window.D2L.FastDom = FastDom;
+window.D2L.fastdom = FastDomPromised;
+window.D2L.fastdom.fastdom = FastDom;
+window.D2L.fastdom.initialize();
```
dlockhart: I wasn't sure whether to put this on `D2L.fastdom` or to be more explicit about it like `D2L.FastDomPromised`. Thoughts?

dbatiste: I was sort of thinking we would always use the extension.  All of our P3 components directly reference fastdom though.  We'd need a wrapper to avoid doing this in those places.

dlockhart: Well the P3 components will likely import it directly, call `.use()` to use the extension and be done with it. This is purely for the monolith.

PR #1236, opened 2019-02-22
bump sequence viewer and related dependencies

jwong-d2l:


package-lock.json
```diff
@@ -3033,7 +3033,7 @@
         "d2l-navigation": "github:BrightspaceUI/navigation#semver:^3",
         "d2l-polymer-siren-behaviors": "github:Brightspace/polymer-siren-behaviors#semver:^1",
         "d2l-typography": "github:BrightspaceUI/typography#semver:^7",
-        "s-html": "github:Brightspace/s-html#semver:^2.0.0"
+        "s-html": "github:Brightspace/s-html#semver:^2"
```
svanherk: I wonder if this is gonna cause issues, cause there's still other `"s-html": "github:Brightspace/s-html#semver:^2.0.0"` lines in this file...

dlockhart: So if other things reference it in the same bundle... it could be an issue, yes.

PR #1237, opened 2019-02-22
Us103342/update d2l my dashboards

johngwilkinson:


johngwilkinson:
The changes to package-lock here are pretty extensive, and I don't know how to tell if they are okay. :(

svanherk:
Seems to be removing all the `^`, I'm not sure why that would be happening... what version of node and npm do you have?

johngwilkinson:
npm 5.6.0
node v8.11.1

svanherk:
I have npm `6.8.0` and node `9.9.0`. Maybe try updating npm to `6.8.0`?  We've seen this fix issues for people, although this problem seems new

johngwilkinson:
Trying with 6.8.0 now.

svanherk:
Ok, I just merged a different PR so you may need to merge master in after

johngwilkinson:
Same issue, I think. Would it be Evil for me to edit package-lock manually? I could still test the built BSI before updating the LMS to use it.

svanherk:
The diff looks way better to me now, but it seems to include lines from the other PR, so git may be confused

johngwilkinson:
There---merged in upstream/master. I think this should be good now. Thanks for the help!

svanherk:
No problem!  I need to make a BSI release, are you merging shortly?  If so I'll wait and get your stuff in there too

johngwilkinson:
Yes, thanks---I'll merge when travis is done.

PR #1239, opened 2019-02-22
US103342: cherry-pick dashboards update from master

johngwilkinson:


johngwilkinson:
Reviewers: I chose you because it looks like you recently made changes to the 10.8.11 branch. If you haven't made a release yet I'd be happy to piggy-back; otherwise I'll take care of it.

PR #1240, opened 2019-02-25
US104330: Update My Courses to remove the (obsolete) touch menu.

njostonehouse:


njostonehouse:
@edwardg8 If you want your latest change to My Courses in this release, we can piggyback.

edwardg8:
I also wanted to add the changes for syncing the overlay and the widget for pinned tab actions, so I'll release new my courses after that is done.

edwardg8:
I'm not too sure if we want to get rid of the package-lock.json file though, if we do that, then all the patches we made to my courses will automatically be picked up by BSI, is that intended?

njostonehouse:
@edwardg8 The package-log.json file isn't deleted, it just gets funny about its updates for some reason. Click in the diff to view the change.

edwardg8:
Never mind, my bad, should have package-lock.json file change opened.

PR #1241, opened 2019-02-25
Checking for all Polymer dependencies

dlockhart:
This improves the logic in the checking we do that ensures our web component dependencies don't end up getting nested in the `package-lock.json`. If they do, that results in multiple versions of the same web component getting included in the bundle.

Previously, this was just looking for all `d2l-*` components. That misses out on non-D2L web components like `s-html` or `iron-icon`. Instead, I'm checking to see if a dependency requires Polymer, which likely means it's a web component. Additionally I added `d2l-hypermedia-constants` since it doesn't actually include Polymer.

### DIFF
--- a/test-package-lock.js
+++ b/test-package-lock.js
@@ -6,17 +6,21 @@ const fs = require('fs');
 const path = require('path');
 
 const packageLockPath = path.join(__dirname, 'package-lock.json');
+const packages = ['d2l-hypermedia-constants'];
 
-function validate(json, depth) {
+function validate(json, depth, parentKey) {
 	depth++;
 	for (const key in json) {
-		if (key.substr(0, 4) === 'd2l-' && depth > 1) {
-			console.error(`D2L sub-dependency detected: "${key}". All "d2l-*" web component dependencies must be at root level to avoid duplicate registrations. Check that the version ranges in "package.json" do not contain anything beyond the major version.`);
-			process.exitCode = 1;
+		if (depth > 1) {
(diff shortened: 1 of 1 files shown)


dlockhart:
Good point. It's maybe not the only one we care about... there are a bunch: `'d2l-fetch', 	'd2l-fetch-auth', 'd2l-fetch-dedupe', 'd2l-fetch-simple-cache', 'd2l-hypermedia-constants', 'd2l-intl', 'd2l-organization-hm-behavior', 'd2l-polymer-siren-behaviors', 'd2l-telemetry-browser-client'`.

Now... you won't get duplicate web component registrations, but including fetch multiple times isn't ideal.

So I can either add them to the list, or make it so that anything that starts with `d2l-` OR has Polymer as a dependency is bad?

At some point we may just want to give in and use Yarn here with a flat dependency tree.

dbatiste:
There's been some recent discussion about introducing a singleton flag to NPM packages that would address this issue.  It sounds like the NPM folks are "onboard", but were waiting for another change (for package overrides) to land before proceeding, and it was merged back on Jan 10th.  If it comes to fruition, I think much of the reason for moving to Yarn would be gone, with NPM singletons being the desired end-state.  The Yarn guy chimed in on the RFC as well - he supports the RFC, but also mentions there would be some ramifications to Yarn.

dlockhart:
Cool yeah -- thanks for reminder. I'd been originally following that issue but lost track of it.

PR #1257, opened 2019-03-06
bump rubric to 3.1.0

iFun:


svanherk:
Hey Bill, your commit changed all of the `semver` ranges to commit shas instead.  This means you have an older version of `npm`, and should update.  Otherwise we'll be constantly be flip-flopping back and forth depending on who runs it, which makes it hard to see what repos were actually updated

iFun:
will update npm to latest version

PR #1261, opened 2019-03-08
Pulling in new version of activity alignments

ChrisKraljevicD2L:


ChrisKraljevicD2L:
On second thought, there's some more changes that are gonna be coming in so I'll save this PR for later

PR #1265, opened 2019-03-11
update with new activities

vix633:

PR #1266, opened 2019-03-11
Pulling in latest changes in d2l-multi-select

ChrisKraljevicD2L:


ChrisKraljevicD2L:
Eeeh this probably shouldn't go in until #1265 does

PR #1280, opened 2019-03-18
DE33463 -- update d2l-activity-alignments (release branch)

mpharoah-d2l:


svanherk:
If this is a cert fix, shouldn't it be merged into the `20.19.4` branch?  (If that doesn't exist yet, we need to create it)

PR #1281, opened 2019-03-19
DE33480 - Releasing certfix for renaming master teacher to teacher

mtjandra:


mtjandra:
@svanherk is at my desk and has approved

PR #1289, opened 2019-03-22
Update d2l-table

svanherk:


svanherk:
FYI @mtjandra, the release of this will include your new nav icon

PR #1292, opened 2019-03-25
Automatically tagging releases

dlockhart:
Here's Part 1 of the BSI improvements.

The workflow works like this:
- Upon every commit, the autotag script runs
- If it's a tagged commit, a pull request or a fork, autotag aborts
- If it's on `master`, we reach out to Rally and query for the release number that's under active development (`20.19.5` today).
- If it's on a branch name that matches our release format (`YY.YY.MM`), we re-format that to drop a trailing `0` in the month name.
- Assuming either of those succeeded, we query GitHub releases to find the largest build number so far -- for example `20.19.5-3`.
- We bump that up by one, and tag the release
- This will kick off a second build, which will do the publish to the CDN as it does today

Part 2 will follow, which will hook this change into the LMS's config file.


cli/autotag.js
```diff
@@ -0,0 +1,193 @@
...
+		requestOptions: {
+			headers: {
+				'X-RallyIntegrationName': 'BSI autotag',
+				'X-RallyIntegrationVendor': 'D2L Corporation',
+				'X-RallyIntegrationVersion': '1.0'
+			}
+		}
+	});
+
+	// Add 12 hours to the current time to account for branching occurring at
```
svanherk: Are we worried about how branching is sometimes delayed by an hour or two (like last sprint)?  It probably won't come up, but technically could if someone was trying to rush something into the current version

dlockhart: It's definitely a thing that could happen. If it did, they'd simply need to manually make the release just like today. I don't see that as a huge deal. When the monorepo exists, we could change this to instead just check for the presence or absence of the branch.

svanherk:
> Upon every commit, the autotag script runs

Is it every commit, or every time Travis runs? Just verifying it wouldn't auto-release three times if a PR had three commits in it

cli/autotag.js
```diff
@@ -0,0 +1,193 @@
...
+		process.exitCode = 1;
+	}
+}
+
+async function main() {
+
+	console.log('Attempting to automatically tag BSI release...');
+
+	const isTaggedCommit = (process.env.TRAVIS_TAG !== '');
+	if (isTaggedCommit) {
```
svanherk: Is this to handle the Travis build that runs after a release is made?

cli/autotag.js
```diff
@@ -0,0 +1,193 @@
...
+
+	const branchName = process.env.TRAVIS_BRANCH;
+	const isMaster = (branchName === 'master');
+
+	let release;
+	if (isMaster) {
+		console.log('  Master branch detected.');
+		const release = tryGetActiveDevelopmentRelease();
+		if (release === null) {
+			console.log('  Aborting auto-tag.');
```
svanherk: Minor: The other console logs are super nice and descriptive, may want to update this one and line 184 with more info as well

cli/autotag.js
dlockhart: You got it. The merge happens, at which point the CI run will auto-tag. Then another build happens that does the release. Saying this out loud... I might look into combining these steps in a separate PR.

cli/autotag.js
dlockhart: So it's kind of weird, but `tryGetActiveDevelopmentRelease` outputs a nice message and this is just adding the fact that this caused an abort to occur.

dlockhart:
Yeah you're correct, it would only happen on each CI run, not every commit.

README.md
```diff
@@ -67,15 +67,19 @@ To integrate a new web component into BSI, perform the following steps:
 2. Add a JavaScript file (i.e. `d2l-my-component.js`) to the `web-components` directory that imports the new web component. (i.e. `../node_modules/my-component/my-component.js`)
 3. Reference the new JavaScript file from the fragments list in `polymer.json`
 
-## Publishing
+## Tagging & Publishing
```
dlockhart: Updated the README as well.

package-lock.json
```diff
@@ -2404,7 +2485,7 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#6752ed2e708eb131f831ccd4b1b3b1f6c0d4d743",
+      "version": "github:BrightspaceHypermediaComponents/activities#4476c3e4a608274e3a5cdeaca5ba3cda504d8b05",
```
svanherk: Minor: If it's easy to pick these out, may want to have this PR just be the BSI tagging changes and not an update as well.  We could then make an update of these files to check the conventional workflow

cli/autotag.js
svanherk: Cool, that works!

package-lock.json
dlockhart: Bah, I thought I did. I keep getting merge conflicts every time someone updates `master`, so it's been fun. :)

svanherk: Looks good to me now (no `d2l-*` updates)!

PR #1296, opened 2019-03-26
Bump d2l-rubic version

rgoessl:


rgoessl:
@gfrancu You mean the `from` property?

dabdullahd2l:
@rgoessl I think both under `from` and `requires`. You can try deleting `node_modules` and `package-lock.json` and then re run `npm i`.

svanherk:
Are you using `npm 6.8.0`?  That usually solves this problem

PR #1306, opened 2019-03-29
A faster build if you don't care about IE11

dlockhart:


dlockhart:
Haha yes, sorry I didn't do that initially... silliness.

PR #1310, opened 2019-04-02
[skip publish] temporarily fix

dlockhart:
It turns out `ReleaseStartDate` is the Last Call Date, so starting just now `20.19.6` is in range of our query, which is resulting in 2 releases coming back.

Temporary fix is to take the first one (which is `20.19.5`) just to get this working quickly for folks. I'm going to try and either sort them (so the first one is guaranteed to be the one we want) or look into querying milestones instead of releases.

### DIFF
--- a/cli/autotag.js
+++ b/cli/autotag.js
@@ -49,7 +49,7 @@ async function tryGetActiveDevelopmentRelease() {
 	try {
 		releases = await rallyApi.query({
 			type: 'release',
-			limit: 1,
+			limit: 10,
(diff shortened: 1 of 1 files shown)


cli/autotag.js
```diff
@@ -197,4 +197,5 @@ async function main() {
 
 }
 
-main();
+//main();
```
svanherk: Was this on purpose?  Or testing that slipped in?

dlockhart: lol no

PR #1313, opened 2019-04-02
[skip release] ordering the releases before taking the closest one

dlockhart:
Here's a more permanent fix for this. It orders the releases by dev-done date and picks the earliest one.


dlockhart:
There is not. :(

PR #1320, opened 2019-04-03
upgrading fetch polyfill

dlockhart:


dlockhart:
@svanherk no rush, but would you be able to do a quick test of this in IE11? Loading the homepage with my courses should suffice. OpenStack is down again so I can't access my machine. :(

PR #1303, opened 2019-03-28
Bump version of activities to 2.5.0

CarolBusman:

PR #1323, opened 2019-04-03
update rubric

mdulat:


emil-furniss:
Can we only update rubrics? Otherwise we should be notifying everyone who has changes in the following version bump...

mdulat:
> Can we only update rubrics? Otherwise we should be notifying everyone who has changes in the following version bump...

I think that comment has more to do with bumping LP to go from e.g. v20.19.5-21 directly to v20.19.5-23, in case the LMS isn't ready yet for changes in v20.19.5-22. I think we usually allow updates to dependencies, don't we?

emil-furniss:
@mdulat I think it's also good form to do it here
https://github.com/Brightspace/brightspace-integration/pull/1303#issuecomment-477735727

PR #1338, opened 2019-04-16
Fix Rally query

dlockhart:
This ended up being a super weird Rally bug. When we query Rally to find active releases using the current date, those queries don't seem to support passing in hours. So we can't rely on the query to filter the active release down for us.

I changed it so that the hours are zeroed out before the query, but then we'll get back the current and next release more often. I then use the current EST time to determine if the first release that comes back is still active and if it's past noon yet on release day.

I threw in `console.group()` to clean up the console logging a bit as I used it a lot to test this out.

### DIFF
--- a/cli/autotag.js
+++ b/cli/autotag.js
@@ -21,7 +21,8 @@ const skipReleaseFlag = '[skip release]';
 
 async function tryGetActiveDevelopmentRelease() {
 
-	console.log('  Fetching active development release from Rally...');
+	console.log('Fetching active development release from Rally...');
+	console.group();
 
 	//  https://github.com/RallyTools/rally-node/wiki/User-Guide
 	const rallyApi = rally({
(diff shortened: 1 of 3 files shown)


cli/autotag.js
```diff
@@ -135,35 +161,41 @@ async function createRelease(newTag) {
 		});
 	} catch (e) {
 		console.error(chalk.red(e));
+		console.groupEnd();
```
svanherk: Minor: I may have messed up the nesting, but is this one needed?  I can't seem to find it's matching beginning, wouldn't https://github.com/Brightspace/brightspace-integration/pull/1338/files#diff-c8ee634add401b47fe9eb5b861e70177R233 be the one that handles it?

dlockhart: Good catch!

PR #1340, opened 2019-04-18
switching to use moment instead of time library

dlockhart:
`time` was breaking Windows builds.


omsmith:
```
	if (releases.TotalResultCount === 2) {
		const releaseDate = new time.Date(releases.Results[0].ReleaseDate);
		releaseDate.setTimezone('America/Toronto');
```

line 77

omsmith:
But it's not good @CodeBaboon! :P

PR #1345, opened 2019-04-23
DE30100 - include d2l-dropdown-button and d2l-dropdown-button-subtle …

klhuff:
…in bsi.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
-node_js: node
+node_js:
+- "11"
 sudo: false
 script:
 - npm run test:ci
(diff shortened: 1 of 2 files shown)


klhuff:
> Crazy! Something else must have been including these at some point and stopped.

d2l-dropdown-button-subtle is clearly still being included in something (as it works inside the monolith), but I'd prefer to avoid it disappearing like d2l-dropdown-button did.

dlockhart:
Absolutely yeah. We should always be explicitly including everything here that gets used in the monolith. Sometimes things just get missed though as they "work" if something else is including them.

PR #1366, opened 2019-04-29
US106001 Update my courses version

ciss1995:


svanherk:
Shouln't we also see updates to the `organizations` and `enrollments` versions?  And `activities`?

edwardg8:
This looks good, Organizations and enrollments are only in my-courses package.json I think.

svanherk:
Agreed this looks good, we may want to wait until we have `activities` `3.1.0` to merge (need to update https://github.com/BrightspaceHypermediaComponents/activities/blob/master/components/d2l-activity-card/d2l-activity-card.js).  But merging this now shouldn't break anything, `d2l-activity-card` is only used in an FRA I believe.  Will leave that up to Cheick, he's working on this now I believe

ciss1995:
I think that I will merge this and come back. We need to fix the tests in d2l-activity-card

PR #1384, opened 2019-05-14
bump bsi

vix633:
Bumping `users` and `icons`.


package-lock.json
```diff
@@ -3470,6 +3470,32 @@
         "d2l-sequences": "github:BrightspaceHypermediaComponents/sequences#semver:^1",
         "d2l-typography": "github:BrightspaceUI/typography#semver:^7",
         "polymer-frau-jwt": "github:Brightspace/polymer-frau-jwt#semver:^1"
+      },
+      "dependencies": {
```
svanherk: @iFun We seemed to lose the `semver` in your sequence bump PR (https://github.com/Brightspace/brightspace-integration/pull/1381/files#diff-32607347f8126e6534ebc7ebaec4853dR3477), which might be causing this to appear here and fail the build.  Did anything change recently with dependencies in the sequences stuff?

PR #1386, opened 2019-05-15
wcag 2.1: input styles

margaree:
also cleaned up old search scss


sass/deprecated/input-helpers.scss
```diff
@@ -4,7 +4,7 @@
 		$background-color: $d2l-color-white,
 		$background-color-hover: null,
 		$border-radius: 0.3rem,
-		$box-shadow: inset 0 2px 0 0 rgba(185, 194, 208, .2),
+		$box-shadow: inset 0 2px 0 0 rgba(181, 189, 194, .2),
```
dlockhart: Could you throw a comment here `/* corundum */` just so we know in the future that this is what we expected?

sass/deprecated/input-helpers.scss
```diff
@@ -122,8 +122,7 @@
 	) {
 
 	@include _vui-input(
-		$background-color: $d2l-color-woolonardo,
-		$background-color-hover: $d2l-color-gypsum,
```
dlockhart: Hmm, so I guess this was wrong before?

sass/deprecated/inputs.scss
```diff
@@ -33,25 +32,6 @@ textarea.vui-input {
...
-.vui-input-search-container {
-	@include _vui-input-search-container();
-}
-.vui-input-search-button {
-	@include _vui-input-search-button();
-}
-.vui-input-search-clear-button {
-	@include _vui-input-search-clear-button();
-}
-
```
dlockhart: Thanks again for finding these!

sass/deprecated/input-helpers.scss
margaree: yup it was at least inconsistent with `d2l-input*` as well as the "before" design in the story

PR #1393, opened 2019-05-21
DE34402-update enrollment

edwardg8:


package.json
```diff
@@ -80,6 +80,7 @@
     "d2l-button-group": "BrightspaceUI/button-group#semver:^3",
     "d2l-colors": "BrightspaceUI/colors#semver:^4",
     "d2l-dropdown": "BrightspaceUI/dropdown#semver:^7",
+    "d2l-enrollments": "BrightspaceHypermediaComponents/enrollments#semver:^2.0.2",
```
svanherk: I think you'll need to remove `^` here and below to lock it to the version we want.  Hopefully that fixes the `package-lock.json` file, otherwise we will need that special `my-courses` release

PR #1404, opened 2019-05-27
Learning path homepage widget (My Learnings)

m6jones:
Learning path summary view

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1721,9 +1721,9 @@
       },
       "dependencies": {
         "core-js": {
-          "version": "2.6.8",
-          "resolved": "https://registry.npmjs.org/core-js/-/core-js-2.6.8.tgz",
-          "integrity": "sha512-RWlREFU74TEkdXzyl1bka66O3kYp8jeTXrvJZDzVVMH8AiHUSOFpL1yfhQJ+wHocAm1m+4971W1PPzfLuCv1vg==",
+          "version": "2.6.9",
+          "resolved": "https://registry.npmjs.org/core-js/-/core-js-2.6.9.tgz",
+          "integrity": "sha512-HOpZf6eXmnl7la+cUdMnLvUxKNqLUzJvgIziQ0DiF3JwSImNphIqdGqzj6hIKyX04MmV0poclQ7+wjWvxQyR2A==",
           "dev": true
         },
         "regenerator-runtime": {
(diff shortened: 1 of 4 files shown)


polymer.json
```diff
@@ -4,6 +4,9 @@
   "fragments": [
     "web-components/bsi.js",
     "web-components/d2l-activity-alignments.js",
+    "web-components/d2l-enrollments-collection-widget.js",
```
svanherk: These have a plural `enrollments` whereas the files below are singular.  And looks like `d2l-activity-alignments` is now here twice

m6jones: thank you!

m6jones:
Is there a reason `d2l-quick-eval` is in `bsi.js` should I be going this route as well?

svanherk:
> Is there a reason d2l-quick-eval is in bsi.js should I be going this route as well?

I'm not actually sure... @dlockhart, is the idea that big components only used in one place should go in their own file, and smaller reused ones should go in `bsi.js`?

m6jones:
(Adding my team to the review so they can start to see the process of front-end development)

package-lock.json
```diff
@@ -1721,9 +1721,9 @@
       },
       "dependencies": {
         "core-js": {
-          "version": "2.6.8",
-          "resolved": "https://registry.npmjs.org/core-js/-/core-js-2.6.8.tgz",
-          "integrity": "sha512-RWlREFU74TEkdXzyl1bka66O3kYp8jeTXrvJZDzVVMH8AiHUSOFpL1yfhQJ+wHocAm1m+4971W1PPzfLuCv1vg==",
+          "version": "2.6.9",
```
jkehres: Why did these versions change without a corresponding change in package.json?

jkehres: Did you intentionally want to bump these versions?

dlockhart:
Yeah exactly -- if things are only used once, then putting them in their own file creates a separate fragmented file for it. That way everyone doesn't download it until they actually visit the app. It does mean that you'll need to load that fragment from the page that hosts the app though -- there are lots of examples of this.

package-lock.json
m6jones: These will get updated automatically believe up to minor version changes. Majors won't happen. Since this is a dependancy that we don't own I am letting it get bumped up.

m6jones: When doing a BSI bump or reviewing. You need to go through the `package-lock` file and be aware of what is being updated.

PR #1424, opened 2019-06-03
Update package-lock.json

djwester:


djwester:
@svanherk I've limited it to just the repos I want to update translations for but still see issues - And ideas?

djwester:
actually it might working after all - things just initially looked red, but on page refresh it looks like the build is running

svanherk:
Yeah this seems good to me, no core UI components updated so shouldn't be any WCAG work leaking in here

svanherk:
(Unless rubrics has hardcoded in their darker border changes)

PR #1436, opened 2019-06-11
Update d2l-outcomes-user-progress

mpharoah-d2l:


rgoessl:
You should bring in the latest release of HM consts as well

PR #1454, opened 2019-06-21
Bump version

mtjandra:


package-lock.json
```diff
@@ -2970,7 +2970,7 @@
       }
     },
     "d2l-enrollments": {
```
svanherk: I think the changes here are going to cause problems until `activities` is updated to use the newest `organizations`

svanherk: @m6jones I think this is the localization changes?

mtjandra: I'll just revert this portion manually for now. Seems it is complaining about that dep as well.

> Polymer sub-dependency detected "d2l-organizations" in "d2l-enrollments". All Polymer dependencies must be at root level of "package-lock.json" to avoid duplicate registrations. Check that the version ranges in "package.json" do not contain anything beyond the major version.

PR #1467, opened 2019-07-04
add d2l-organizations

ctwomey1:
Adding BrightspaceHypermediaComponents/organizations so we can use the d2l-organizations-consortium-tabs component.

This component will be rendered just above the nav, for lms's that are part of a consortium and are configured to have user associations.


package.json
```diff
@@ -97,6 +97,7 @@
     "d2l-navigation": "BrightspaceUI/navigation#semver:^3",
     "d2l-offscreen": "BrightspaceUI/offscreen#semver:^4",
     "d2l-opt-in-flyout-webcomponent": "Brightspace/d2l-opt-in-flyout-webcomponent#semver:^2",
+    "d2l-organizations": "github:BrightspaceHypermediaComponents/organizations#semver:5",
```
svanherk: This looks good, but something about the `package-lock.json` file has messed up and wants a lower version of `organizations`, which fails the build.  Did you maybe have an older version?

ctwomey1: Could be something else is pulling organizations in, i'll give it a look.  

I know locally I had to blow away node_modules before bsi would build properly

ctwomey1:
To resolve the issues with nesting I had to:
- `ri ./node_modules -Recurse -Force`
- `ri package-lock.json`
- `npm i`

npm i doesn't seem to always know how to flatten or even pull in new dependencies, it looked like it was pulling in an old version of activities for sure since @m6jones had bumped organizations in activities ~1 week ago.

Hopefully this fixes ci :/

PR #1506, opened 2019-07-23
US108638 - Reference the iterator

jeroach:
[US108638](https://rally1.rallydev.com/#/detail/userstory/322123650748?fdp=true): [UI / LMS] Ensure that right slot of immersive nav is working as expected

Que the facepalm...


jeroach:
> hehe at least there wasn't a larger issue! I actually didn't know you had to edit this file to get your thing into BSI.

It's only for new components :) If you're updating an existing component then you don't need to do anything extra. I updated this many times in the past when we were adding components a lot, but it's been a while and I forgot lol

jeroach:
PR build keeps stalling (no test failures or anything) - not sure what's up but going to keep kicking it.

PR #1511, opened 2019-07-25
removing unused dependencies on iron-icon and iron-iconset-svg

dlockhart:
This seems safe, but could use a second opinion.

`iron-iconset-svg` is no longer used in BSI (@dbatiste cleaned that up recently) and isn't used in the monolith. Ditto for `iron-icon`. This change resulted in no change to the `package-lock` file, which isn't surprising since other things inside BSI are still using them.

The only "risk" I can think of is that when `d2l-icons` stops pulling these in, then in theory if another dependency in BSI was using one of them but not importing it, it would break.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,8 +62,6 @@
     "@brightspace-ui/core": "^0.2",
     "@octokit/rest": "^16.19.0",
     "@polymer/esm-amd-loader": "^1.0.3",
-    "@polymer/iron-icon": "^3.0.0-pre.21",
-    "@polymer/iron-iconset-svg": "^3.0.1",
     "@polymer/lit-element": "^0.6.1",
     "@polymer/polymer": "^3.2.0",
     "@webcomponents/webcomponentsjs": "^2.0.4",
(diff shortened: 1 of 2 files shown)


svanherk:
> The only "risk" I can think of is that when d2l-icons stops pulling these in, then in theory if another dependency in BSI was using one of them but not importing it, it would break.

Would that have failed the bundling/build if the file was imported but missing?

dlockhart:
Probably, yeah!

dbatiste:
> Would that have failed the bundling/build if the file was imported but missing?

Hmmm, if they are missing the import statement in their component, the bundler wouldn't try to pull it in, and I don't think it would break the BSI build.  That missing import statement may or may not have been evident in their demo.  At runtime the tag would get ignored, but if they try to access the API it would blow up.  The risk ought to be low though, and we could perhaps check usage?

svanherk:
Ah ok, I was thinking of a scenario where they were importing it, but not including it in their `package.json`

dbatiste:
Yeah in that scenario I agree, the build would fail.

PR #1514, opened 2019-07-25
Bump navigation to 3.10.0

jeroach:
https://github.com/BrightspaceUI/navigation/releases/tag/v3.10.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -931,9 +931,9 @@
       "dev": true
     },
     "@octokit/endpoint": {
-      "version": "5.3.1",
-      "resolved": "https://registry.npmjs.org/@octokit/endpoint/-/endpoint-5.3.1.tgz",
-      "integrity": "sha512-4mKqSQfeTRFpQMUGIUG1ewdQT64b2YpvjG2dE1x7nhQupdI/AjdgdcIsmPtRFEXlih/uLQLRWJL4FrivpQdC7A==",
+      "version": "5.3.2",
+      "resolved": "https://registry.npmjs.org/@octokit/endpoint/-/endpoint-5.3.2.tgz",
+      "integrity": "sha512-gRjteEM9I6f4D8vtwU2iGUTn9RX/AJ0SVXiqBUEuYEWVGGAVjSXdT0oNmghH5lvQNWs8mwt6ZaultuG6yXivNw==",
       "dev": true,
       "requires": {
         "deepmerge": "4.0.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -2858,7 +2858,7 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#43cd7f9c59bcbf6625aa76b5346026b98b4475d3",
+      "version": "github:BrightspaceHypermediaComponents/activities#3342221132a1b92658e2bc8db7ab8329e6f2374e",
```
jeroach: fyi @maboelsoudd2l

package-lock.json
```diff
@@ -2893,7 +2893,7 @@
       }
     },
     "d2l-activity-alignments": {
-      "version": "github:Brightspace/d2l-activity-alignments#0c02f47e11fea1526432263c5eba330425bd6243",
+      "version": "github:Brightspace/d2l-activity-alignments#94f750e4ae47264a72a9f975a79d03c484530167",
```
jeroach: fyi @djwester

package-lock.json
```diff
@@ -3494,7 +3494,7 @@
       }
     },
     "d2l-rubric": {
-      "version": "github:Brightspace/d2l-rubric#997cc3c39bda43b35a7987a011f1c9d159c6da0b",
+      "version": "github:Brightspace/d2l-rubric#a63004de4a9bf1da10bd858e8f998b10109cc0f6",
```
jeroach: fyi @kevinwang-d2l

jeroach:
> I thought we just we just updated this in BSI?

When adding a new web component, you also need to reference it in `bsi.js` so that it becomes available for use in the LMS.

Also, when bumping the BSI version, it's best to pickup all other _minor_ version changes from other linked repos. And then notify those who did the change (as I did). If we skip this process, weird things can end up happening - remember the issue with BSI that Viktor fixed last week :)

maboelsoudd2l:
> > I thought we just we just updated this in BSI?
> 
> When adding a new web component, you also need to reference it in `bsi.js` so that it becomes available for use in the LMS.
> 
> Also, when bumping the BSI version, it's best to pickup all other _minor_ version changes from other linked repos. And then notify those who did the change (as I did). If we skip this process, weird things can end up happening - remember the issue with BSI that Viktor fixed last week :)

Does this process of picking up other minor version happen automatically?
Also @mtjandra updated the activities version to a newer release in #1512 , wouldn't this PR bump it back?

jeroach:
Mohammed and I chatted offline :)

PR #1529, opened 2019-08-01
Add d2l-switch component

kieranderson:


ppaskaris-d2l:
I have no idea how this works. Since my review is invalid (I am not a repo contributor), I would recommend you reach out to someone in UXD or something to help facilitate this.

kieranderson:
+Dave, because Github suggested you :)

polymer.json
```diff
@@ -18,7 +18,8 @@
     "web-components/d2l-rubric.js",
     "web-components/d2l-rubric-editor.js",
     "web-components/d2l-outcomes-level-of-achievements.js",
-    "web-components/d2l-outcomes-user-progress.js"
+    "web-components/d2l-outcomes-user-progress.js",
+    "web-components/d2l-switch.js"
```
dbatiste: My current thinking is that this should be moved to `bsi.js`.  @dlockhart , thoughts?

dlockhart: The only reason to make it a separate shard would be if we only want to import it on pages that use the component.

kieranderson: That is what I was planning on doing, however in the future we most likely want to include it in all pages.

So I can move it now to `bsi.js` if preferred.

kieranderson: Moved into `bsi.js` Thanks.

kieranderson:
Updated Looking to get this in for use in a single Renderer for now. Let me know if this is good.

PR #1530, opened 2019-08-01
pulling in latest organizations and card

dlockhart:
These no longer use `iron-iconset-svg`, which means we can upgrade to Lit icons without issue.


svanherk:
Looks like this is contained in https://github.com/Brightspace/brightspace-integration/pull/1531

dlockhart:
Oh wow that was fast -- thanks!

PR #1539, opened 2019-08-07
using Lit buttons and more-less

dlockhart:
This should be the last of switching stuff over!


package-lock.json
```diff
@@ -2936,7 +2936,7 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#fbac3bc1eb538cc1dfa3800bcc9b674aba017e6d",
+      "version": "github:BrightspaceHypermediaComponents/activities#ad3d2923cf9df87865386654a7bf82d2e7164050",
```
dlockhart: @maboelsoudd2l FYI

package-lock.json
```diff
@@ -3111,7 +3107,7 @@
       }
     },
     "d2l-enrollments": {
-      "version": "github:BrightspaceHypermediaComponents/enrollments#a76b6eb4bd76e728d00c65f96c7677a0dc7e2f1e",
+      "version": "github:BrightspaceHypermediaComponents/enrollments#760bcd3eeac299337de989ddb26962e318ca965d",
```
dlockhart: @m6jones FYI

package.json
```diff
@@ -75,21 +75,18 @@
...
     "d2l-activity-alignments": "Brightspace/d2l-activity-alignments.git#semver:^1",
     "d2l-alert": "BrightspaceUI/alert#semver:^4",
-    "d2l-button": "BrightspaceUI/button#semver:^5",
     "d2l-button-group": "BrightspaceUI/button-group#semver:^3",
     "d2l-dropdown": "BrightspaceUI/dropdown#semver:^7",
     "d2l-fetch": "Brightspace/d2l-fetch.git#semver:^2",
     "d2l-fetch-auth": "^1.5.2",
     "d2l-fetch-dedupe": "^1.1.0",
     "d2l-fetch-simple-cache": "^1.0.0",
-    "d2l-icons": "BrightspaceUI/icons#semver:^6",
```
dlockhart: Forgot to remove this last time

polymer.json
```diff
@@ -7,7 +7,6 @@
     "web-components/d2l-enrollment-collection-widget.js",
     "web-components/d2l-enrollment-summary-view.js",
     "web-components/d2l-image-banner-overlay.js",
-    "web-components/d2l-more-less.js",
```
dlockhart: Removing this fragment will require a change in the LMS that I'll make. It's not needed since multiple things reference more/less, so it gets pulled into the shared bundle anyway.

sass/deprecated/button.scss
```diff
@@ -1,4 +1,73 @@
 @import '../../node_modules/@brightspace-ui/core/components/colors/colors.scss';
```
dlockhart: This contains a copy of button's Sass, since I didn't want to copy it into core. I think performance is good enough now that we can switch Brightspace over to use the web components instead of the Sass, and then we could delete this.

web-components/bsi.js
```diff
@@ -37,22 +37,23 @@ window.D2L.Telemetry = {
 	}
 };
 
+import '@brightspace-ui/core/components/button/button-icon.js';
+import '@brightspace-ui/core/components/button/button-subtle.js';
+import '@brightspace-ui/core/components/button/floating-buttons.js';
```
dlockhart: @margaree should [your PR to fix some of those floating buttons issues](https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/2078/overview) merge?

PR #1561, opened 2019-08-19
increasing core to 1.2.1

dlockhart:


package-lock.json
```diff
@@ -3199,7 +3207,7 @@
       }
     },
     "d2l-hypermedia-constants": {
-      "version": "github:Brightspace/d2l-hypermedia-constants#105af921b4f6beef1d82b48ed00fd5442d9725f6",
+      "version": "github:Brightspace/d2l-hypermedia-constants#8cd6b6caccc4b0831ba63f27cb8d859477500aba",
```
dlockhart: @ryantmer

package-lock.json
```diff
@@ -2918,7 +2926,7 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#7331a479b2a35d2f49cd1280db1dd76a49ccad33",
+      "version": "github:BrightspaceHypermediaComponents/activities#3a74703186fbbb70501b14356e29180e8e69173a",
```
dlockhart: @maboelsoudd2l @jeroach

package-lock.json
ryantmer: Additive change with some rels that have existed forever, but weren't in `d2l-hypermedia-constants` :)

dlockhart:
I've updated this to pull in v4 of navigation, and also updated the 3 places that were using v3 of navigation to also reference v4.

PR #1563, opened 2019-08-20
Bumping activities to 3.42.0.

cbusmand2l:


cbusmand2l:
fyi: @ryantmer, @dlockhart @ctwomey1

PR #1568, opened 2019-08-21
Bump Navigation

mtjandra:


package-lock.json
```diff
@@ -3400,7 +3400,7 @@
       }
     },
     "d2l-navigation": {
-      "version": "github:BrightspaceUI/navigation#079e045b64ffd7f21e128747b8a47430c49f0c7e",
+      "version": "github:BrightspaceUI/navigation#927b5d07ce25f31a603e50d8bb9195a540489c2f",
       "from": "github:BrightspaceUI/navigation#semver:^3",
```
jeroach: We should also bump `semver` to 4 to be consistent

mtjandra: Do I also bump the version in `package.json`? I remember you were saying some things depends on navigation, but do not reference it from BSI. So I was hesitant to change those version numbers

jeroach: yeah both places - the repos that depend on navigation, but do not reference from BSI will have their own `package.json` still pointing at `"github:BrightspaceUI/navigation#semver:^3"`

dlockhart: It all gets built together though... so I have no idea what this is going to do. I don't know which version you're going to end up with TBH.

mtjandra: Yeah I thought whatever is in BSI is what is used everywhere, (unless you are using an FRA or something). But if I'm interpreting Josh correctly, seems like that is not the case. Either way, I think the behavior should be the same as the previous wonky nav BSI release

mtjandra: @jeroach do you know if there is a page we can check to verify if things are working correctly or not to test out your theory?

PR #1600, opened 2019-09-10
Move all imports for d2l-tabs to single file

rgoessl:


web-components/d2l-tab-panel.js
```diff
@@ -1 +0,0 @@
-import 'd2l-tabs/d2l-tab-panel.js';
```
dlockhart: I can't tell from the diff... can this file just be deleted?

rgoessl: The import is now in `d2l-tabs.js`, don't need that file anymore. Is that not what you meant by putting them in the same bsi import?

rgoessl: The file was deleted if that's what you're asking

dlockhart: Haha yeah sorry was just asking if the file was deleted -- now I see where in the diff it says that. All good!

PR #1616, opened 2019-09-16
upgrading to latest intl-messageformat

dlockhart:


dlockhart:
Weird errors that I'll need to investigate. Reverting the change to `core` for now and closing this PR.

PR #1628, opened 2019-09-23
US110531 - add quick eval fragment

AlexBedley:
Problem: Quick Eval has problems when making API calls during page load
Problem: This was traced a race condition with d2lfetch auth
Problem: The root cause was because Quick Eval was being imported before d2lfetch auth was set up
Problem: Quick Eval is loaded on every single page that loads bsi
Solution: Integrate QE as a fragment instead

This is my attempt to do this without holding up BSI for the whole company.

First PR (this one): Add QE as a fragment, push BSI to LP
Second PR: Use QE fragment in LMS - this might cause QE to be loaded twice on the QE page but my assumption is that it will still work despite blowing up the console
Third PR: remove QE from bsi.js, push BSI to LP

### DIFF
--- a/polymer.json
+++ b/polymer.json
@@ -20,7 +20,8 @@
     "web-components/d2l-outcomes-level-of-achievements.js",
     "web-components/d2l-outcomes-user-progress.js",
     "web-components/d2l-program-outcomes-picker.js",
-    "web-components/d2l-tabs.js"
+    "web-components/d2l-tabs.js",
+    "web-components/d2l-quick-eval.js"
   ],
   "extraDependencies": [
     "node_modules/d2l-rubric/editor/images/**",
--- a/web-components/d2l-quick-eval.js
+++ b/web-components/d2l-quick-eval.js
@@ -1,0 +1,1 @@
+import 'd2l-activities/components/d2l-quick-eval/d2l-quick-eval.js';


maboelsoudd2l:
side question @svanherk : even though this is an easy change, is there an easy way for us to test that this works (before we find out there could be something wrong on quad)?

svanherk:
@maboelsoudd2l You should be able to test by hosting BSI locally, and pointing to that from your local LMS's config file (https://github.com/Brightspace/brightspace-integration#running-locally---bundled-production-build).  With this, you could test that it still works without LMS changes, and then add your updates and confirm that works as well.
If you're more concerned about just the regression test against current quad, you could also merge this and then update the override config `d2l.System.BsiEndpointOverride` on the latest quad site to point to this version and check.  I think the LMS update PR occurs automatically now, but you would at least know early and could merge the revert PR soon after if something was wrong.

dlockhart:
I don't love using the `BsiEndpointOverride` config as a common daily occurrence that's part of our dev/test workflows. Changing it is really disruptive to others that might be testing on that same quad site, and if they find bugs, the developers could end up being really confused as to why their changes got into a build they weren't expecting.

I'd much prefer we wait the couple hours (down from ~6 hours before the recent improvements) for the changes to really make it into a quad build. Let's reserve using the config for extenuating circumstances.

PR #1645, opened 2019-09-30
D2L activity exemptions

BraydonKains:
Adding new Polymer 3 version of the d2l-activity-exemptions component.


svanherk:
Hey @BraydonKains, are you using npm `6.8.0`?  We've seen issues with other versions, because of our `#semver:` syntax and it looks like a lot of stuff is being deleted that's still needed.

package.json
```diff
@@ -109,6 +109,7 @@
     "d2l-table": "BrightspaceUI/table#semver:^2",
     "d2l-tabs": "BrightspaceUI/tabs#semver:^1",
     "d2l-users": "BrightspaceHypermediaComponents/users#semver:^2",
+    "d2l-activity-exemptions": "Brightspace/d2l-activity-exemptions#semver:^2",
```
svanherk: Might also be an issue with a mismatch here? https://github.com/Brightspace/d2l-activity-exemptions/blob/master/package.json#L8

BraydonKains:
> 
> 
> Hey @BraydonKains, are you using npm `6.8.0`? We've seen issues with other versions, because of our `#semver:` syntax and it looks like a lot of stuff is being deleted that's still needed.

I am on 6.11 at the moment. Do you suggest I downgrade?

svanherk:
> I am on 6.11 at the moment. Do you suggest I downgrade?

Yeah unfortunately, `6.9` broke the functionality we were relying on.  I recommend using `nvm` (for Mac) or `nvm for Windows` (https://github.com/coreybutler/nvm-windows) to be able to switch between node versions quickly, which can have a specific npm version installed.

PR #1652, opened 2019-10-03
quick eval now imported as fragment

vix633:


vix633:
> I think we should test this with a local BSI before merging, but the change looks good!

This is tested locally and on sitelord site working, and fixing the 401 issue as a byproduct

AlexBedley:
> > I think we should test this with a local BSI before merging, but the change looks good!
> 
> This is tested locally and on sitelord site working, and fixing the 401 issue as a byproduct

PR #1656, opened 2019-10-03
Add d2l-activities Serge JSON files dependencies

ryantmer:
An unforeseen result of the work done in https://github.com/BrightspaceHypermediaComponents/activities/pull/415, since `d2l-activities` is using `fetch` rather than `import` to fetch its translations, the BSI build ends up excluding the files from the build. Sadly, adding them to the `extraDependencies` in that repo isn't sufficient, and they still get skipped, whereas adding them here does work.

This is probably another small factor that leans us towards a better solution, possibly a build script - having to explicitly include each repo's JSON language files here would be pretty messy. This PR does get things back to working, though.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -868,7 +868,7 @@
       }
     },
     "@d2l/audio": {
-      "version": "github:Brightspace/d2l-audio#e66feb688bc34e4a9fc573295a1accfe522a7580",
+      "version": "github:Brightspace/d2l-audio#ce6aeb909a5feb9a765fb2ec564b85ee6b2cf1e1",
       "from": "github:Brightspace/d2l-audio#semver:^3",
       "dev": true,
       "requires": {
@@ -885,7 +885,7 @@
       }
     },
     "@d2l/media-behavior": {
-      "version": "github:Brightspace/d2l-media-behavior#bcf72a7cc48f766cb52a2add3ed4b7cb2110cece",
+      "version": "github:Brightspace/d2l-media-behavior#8105ee13faf56edca5f7e8d913089622dfb76092",
       "from": "github:Brightspace/d2l-media-behavior#semver:^1",
       "dev": true,
       "requires": {
@@ -895,7 +895,7 @@
       }
     },
     "@d2l/seek-bar": {
-      "version": "github:Brightspace/d2l-seek-bar#dd6f619628e4ddaa6d807f9275a4cda1c271fbe3",
+      "version": "github:Brightspace/d2l-seek-bar#89ffcb054e3e3e3f308f9073627931f45d5d78b9",
       "from": "github:Brightspace/d2l-seek-bar#semver:^1",
       "dev": true,
       "requires": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -868,7 +868,7 @@
       }
     },
     "@d2l/audio": {
-      "version": "github:Brightspace/d2l-audio#e66feb688bc34e4a9fc573295a1accfe522a7580",
+      "version": "github:Brightspace/d2l-audio#ce6aeb909a5feb9a765fb2ec564b85ee6b2cf1e1",
```
ryantmer: FYI @EvanLemieux

package-lock.json
```diff
@@ -885,7 +885,7 @@
       }
     },
     "@d2l/media-behavior": {
-      "version": "github:Brightspace/d2l-media-behavior#bcf72a7cc48f766cb52a2add3ed4b7cb2110cece",
+      "version": "github:Brightspace/d2l-media-behavior#8105ee13faf56edca5f7e8d913089622dfb76092",
```
ryantmer: FYI @EvanLemieux

package-lock.json
```diff
@@ -895,7 +895,7 @@
       }
     },
     "@d2l/seek-bar": {
-      "version": "github:Brightspace/d2l-seek-bar#dd6f619628e4ddaa6d807f9275a4cda1c271fbe3",
+      "version": "github:Brightspace/d2l-seek-bar#89ffcb054e3e3e3f308f9073627931f45d5d78b9",
```
ryantmer: FYI @EvanLemieux

mdgbayly:
> This is probably another small factor that leans us towards a better solution, possibly a build script - having to explicitly include each repo's JSON language files here would be pretty messy. This PR does get things back to working, though.

I guess at one level you could say having to include the extra dependencies here is that `build script`. :-)

But yeh, it seems like the biggest stumbling blocks here are ECMAScript's inability to import JSON files, and the Serge toolings inability to generate JS files.

If either of those were supported then we could possibly go back to using dynamic imports to load the lang terms.

And if we had dynamic imports, then build tooling like the polymer bundling would automatically generate fragments for each lang file and we probably wouldn't have to bother adding them to the BSI build script and we could still avoid a build script in the component repos.

Possibly investigating the having the Serge tooling generate ES6 modules would be the most palatable solution.

ryantmer:
That would be a great solution - even if it was just the ability to stick the JSON into a string template in a JS file, that would be sufficient! Maybe there's an option to do that somewhere...

dbatiste:
I agree, there may be cleaner options if we can control how Serge parses, and what it outputs. Serge does support plugins, in at least some limited way.  We apparently have this plugin, [parse_d2l_fra](https://github.com/Brightspace/sp2-worker/blob/74a8dcc6787a54975a40609ae629e320300a4f6f/static/serge/parse_d2l_fra.pm). I don't find the documentation for their API very informative though. I think they have a plugin for JS files, but I'm not exactly sure what it does, or what it's limitations are.

dbatiste:
I'd be interested in possibilities for `core` as well, where we are likely to have many components with their own languages files, possibly only containing one or two terms.

ryantmer:
Wow, can't believe I missed this:

> Here are some of the things that you can do with Serge:
> ...
> - Use regular expressions to modify the localized file content before it gets saved
https://www.serge.io/docs/modular-architecture/

Assuming that includes being able to change the file extension, this may be easier than I previously thought... I went straight to the config docs previously, guess I should have read the whole set of them!

PR #1667, opened 2019-10-16
Update package-lock.json

djwester:


djwester:
Fixes the issue with overall levels not showing up with Dutch translations

PR #1671, opened 2019-10-16
bumping activities

maboelsoudd2l:
fyi @DiljotSG @BraydonKains @BhamelinD2L


BraydonKains:
Are we going to trigger a BSI release here? If so, can someone send me the version this is included in? One of my team members needs it for testing.

PR #1673, opened 2019-10-17
Bumping d2l-activity-exemptions version for bugfix

BraydonKains:


BraydonKains:
Not sure who else to put on this review (Joseph is OOO today)

PR #1678, opened 2019-10-18
Bump BHC/activities

ryantmer:
Fixes change event dispatching in the instructions component of `d2l-activity-assignment-editor`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -853,9 +853,9 @@
       }
     },
     "@brightspace-ui/core": {
-      "version": "1.14.1",
-      "resolved": "https://registry.npmjs.org/@brightspace-ui/core/-/core-1.14.1.tgz",
-      "integrity": "sha512-q1jQi7uoSKO/AUHngn8647QIsptYP/F/81M0oX0F/LsOu+rnnI22y6R+r1cx+WiYopj9CQ7bM0df8JYRxJII9A==",
+      "version": "1.14.2",
+      "resolved": "https://registry.npmjs.org/@brightspace-ui/core/-/core-1.14.2.tgz",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -3263,7 +3263,7 @@
         "fastdom": "^1.0.8",
         "merge-stream": "^1.0.1",
         "require-dir": "^1.2.0",
-        "siren-sdk": "github:BrightspaceHypermediaComponents/siren-sdk#semver:^1"
+        "siren-sdk": "github:BrightspaceHypermediaComponents/siren-sdk#semver:^1.4.2"
```
ryantmer: FYI @jlai-d2l, this appears to come from [this change](https://github.com/BrightspaceHypermediaComponents/activities/pull/441/commits/728c68026bcbb0820f850e373e38c7a6652d4697#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R96) - seems okay, but wanted to verify that it's not a surprise.

package-lock.json
```diff
@@ -853,9 +853,9 @@
       }
     },
     "@brightspace-ui/core": {
-      "version": "1.14.1",
-      "resolved": "https://registry.npmjs.org/@brightspace-ui/core/-/core-1.14.1.tgz",
-      "integrity": "sha512-q1jQi7uoSKO/AUHngn8647QIsptYP/F/81M0oX0F/LsOu+rnnI22y6R+r1cx+WiYopj9CQ7bM0df8JYRxJII9A==",
+      "version": "1.14.2",
```
ryantmer: FYI @dlockhart

package-lock.json
ryantmer: Looks like this causes the build to fail, in fact - @jlai-d2l mind removing this from activities and doing a new release?

svanherk: Yup, we need to keep this at `^1`, or else we'll get multiple versions of the sdk components which will cause registration errors due to name collisions

PR #1679, opened 2019-10-18
updated d2l-table

maboelsoudd2l:


package.json
```diff
@@ -106,10 +107,9 @@
     "d2l-scroll-spy": "BrightspaceUI/scroll-spy#semver:^1",
     "d2l-sequence-viewer": "Brightspace/d2l-sequence-viewer#semver:^1",
     "d2l-simple-overlay": "Brightspace/simple-overlay#semver:^3",
-    "d2l-table": "BrightspaceUI/table#semver:^2",
+    "d2l-table": "github:BrightspaceUI/table#semver:^2",
```
dlockhart: The `github:` shouldn't be necessary

maboelsoudd2l: why did npm add it?

dlockhart: Not sure.

dlockhart: It used to be required for GitHub dependencies, but that requirement was removed a while ago.

PR #1706, opened 2019-11-11
Update package-lock.json

djwester:


djwester:
Getting latest rubrics fixes into master

PR #1721, opened 2019-11-20
updates to support new select styles in the HTML editor

dlockhart:
Corresponding change coming in the LMS.

These changes support the concept of "bordered" HTML editor toolbar groups (all the existing ones) and non-bordered (`<select>`-based ones).

It also stops overriding the borders of `<select>`s (except toning down the colour to match the other items).

### DIFF
--- a/sass/html-editor/group.scss
+++ b/sass/html-editor/group.scss
@@ -3,11 +3,7 @@
 
 .d2l-htmleditor-group {
 
-	border: 1px solid #D3D3D3;
-	border-radius: $vui-htmleditor-border-radius;
-	box-shadow: 0 1px 1px 0 #EFEFEF;
 	display: inline-block;
-	filter: progid:DXImageTransform.Microsoft.gradient( startColorstr='#FFFFFF', endColorstr='#EEEEEE',GradientType=0 ); /* IE6-9 */
 	line-height: 1em;
 	margin-right: 0.3rem;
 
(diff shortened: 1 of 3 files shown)


sass/html-editor/group.scss
```diff
@@ -3,11 +3,7 @@
 
 .d2l-htmleditor-group {
 
-	border: 1px solid #D3D3D3;
-	border-radius: $vui-htmleditor-border-radius;
-	box-shadow: 0 1px 1px 0 #EFEFEF;
 	display: inline-block;
-	filter: progid:DXImageTransform.Microsoft.gradient( startColorstr='#FFFFFF', endColorstr='#EEEEEE',GradientType=0 ); /* IE6-9 */
```
dlockhart: Seems like an old IE holdover.

sass/html-editor/group.scss
```diff
@@ -24,19 +20,32 @@
...
 	.d2l-htmleditor-toolbar-item {
-		.d2l-select {
-			border-radius: 0;
+		.d2l-select,
+		.d2l-select:hover:disabled {
+			border-color: #D3D3D3;
+		}
+		.d2l-select:hover, .d2l-select:focus {
+			border-color: $d2l-color-celestine;
 		}
```
dlockhart: This makes `<select>`s have the same border colour as the other toolbar items (slightly lighter). It's the only modification we make now.

sass/html-editor/html-editor.scss
```diff
@@ -7,7 +7,7 @@
 @import 'menu.scss';
 
 .d2l-htmleditor {
-	border: 1px solid #CCC;
+	border: 1px solid $d2l-color-galena;
```
dlockhart: We're darkening the border of the editor itself so that it matches the other input types.

sass/html-editor/shared.scss
```diff
@@ -1,3 +1,3 @@
-$vui-htmleditor-icon-height: 16px;
+$vui-htmleditor-icon-height: 18px;
```
dlockhart: The icons are actually `18px`, so this was throwing off the height calculation and causing things to be `44px` high instead of `42px`.

PR #1725, opened 2019-11-21
Use latest siren-sdk and d2l-organizations

wongvincent:


polymer.json
```diff
@@ -27,7 +27,7 @@
   ],
   "extraDependencies": [
     "node_modules/d2l-activities/components/d2l-activity-editor/**/lang/*.js",
-    "node_modules/d2l-organizations/components/d2l-organization-availability-set/lang/*.json",
+    "node_modules/d2l-organizations/components/d2l-organization-availability-set/lang/*.js",
```
ryantmer: Looks like this jumping the gun a bit here, since [the config](https://github.com/BrightspaceHypermediaComponents/organizations/blob/master/organizations.serge.json#L107) is still going to create JSON files - unless I'm missing something?

wongvincent: I am just about to create the other PR to change that!

wongvincent: Actually this PR should include using the latest d2l-organizations, so hold off on reviewing this fornow

wongvincent:
d2l-organizations @svanherk

PR #1731, opened 2019-11-27
Update d2l-activities

kieranderson:


svanherk:
To avoid having to add a lot of new dependencies to BSI, to use the editor you probably just need to move https://github.com/BrightspaceUILabs/edit-in-place/blob/master/package.json#L42 to a dev-dependency

svanherk:
> To avoid having to add a lot of new dependencies to BSI, to use the editor you probably just need to move https://github.com/BrightspaceUILabs/edit-in-place/blob/master/package.json#L42 to a dev-dependency

FYI @BhamelinD2L

BhamelinD2L:
Oh! I'll clean that dependency up shortly.

PR #1742, opened 2019-12-02
Nav icons for Manager Dashboard and Learning Groups

chrisd2l:
[US111114](https://rally1.rallydev.com/#/detail/userstory/340584315068?fdp=true)


chrisd2l:
Thanks for the quick approval! 
Yeah, it is a bit funny ... the use case would be for K-12 customers who adopt some of the corporate features and are used to the icon based navigation. Also, it makes demos that show K-12 and corporate features a bit more polished.

PR #1755, opened 2019-12-09
Bumping activities version to 3.50.1

cbusmand2l:
FYI @dlockhart, @mdgbayly, @BhamelinD2L @wongvincent, @xiangqiyan

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -982,18 +982,6 @@
         "d2l-tooltip": "github:BrightspaceUI/tooltip#semver:^3",
         "lit-element": "^2",
         "siren-sdk": "github:BrightspaceHypermediaComponents/siren-sdk#semver:^1"
-      },
-      "dependencies": {
-        "siren-sdk": {
-          "version": "github:BrightspaceHypermediaComponents/siren-sdk#9b04aee997ec69dd4f80ff6df6cd626985afbbf3",
-          "from": "github:BrightspaceHypermediaComponents/siren-sdk#semver:^1",
-          "dev": true,
-          "requires": {
-            "@brightspace-ui/core": "^1",
-            "@polymer/polymer": "^3.0.0",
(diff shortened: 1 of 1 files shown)


AlexBedley:
Ok I give up. Here's what I got so far:

1. CI is failing because it cannot find a file in d2l-attachments
2. However, that file exists and the code looks like it should work
3. `npm run build` fails locally with the same error. The weird thing there is it's actually looking for the file relative to my current directory instead of the file that's importing it (`attachment-content.js`). So of course the file is not there. But the file is in the correct spot (`../helpers/attachment-utils.js`) relative to `attachment-content.js`
4. Just for fun I opened `../helpers/attachment-utils.js` to see what was in it. `defaultLink`, which is what `attachment-content.js` is specifically importing, is not in that file.

My guess is that d2l-attachments is using some weird babel thing which causes `defaultLink` to not be transpiled which makes `import` think it can't find the file (although the file is "there") since it's missing `defaultLink`. However I'm so deep into this I'm having a hard time keeping it straight, a second pair of eyes would be nice.

@mdgbayly you are the sole author of https://github.com/Brightspace/attachment. Would you mind taking a look and see if I forgot something obvious?

mdgbayly:
@AlexBedley 

> @mdgbayly you are the sole author of https://github.com/Brightspace/attachment. Would you mind taking a look and see if I forgot something obvious?

That's pretty weird. Just taking a look...

defaultLink is in `attachment-utils` so not sure why it is causing the polymer build issues. We've just used it in a webpack based build prior to @ryantmer  adding it to BSI, so maybe some weird transpiler issue as you say.

https://github.com/Brightspace/attachment/blob/master/helpers/attachment-utils.js#L69

dlockhart:
I have a PR up to fix the problem here:
https://github.com/Brightspace/attachment/pull/8

mdgbayly:
> I have a PR up to fix the problem here:
> [Brightspace/attachment#8](https://github.com/Brightspace/attachment/pull/8)

@dlockhart Aha! Thanks!

PR #1756, opened 2019-12-09
Expose the announce method.

dbatiste:
Considering starting to update the places that use iron-a11y-announcer.  There are quite a few however, which could be a big clean-up task to update & test them all. So, my current goal is to get the LMS impl updated.

### DIFF
--- a/web-components/bsi.js
+++ b/web-components/bsi.js
@@ -89,6 +89,9 @@ import 'd2l-users/components/d2l-profile-image.js';
 import { setCancelSyntheticClickEvents  } from '@polymer/polymer/lib/utils/settings.js';
 setCancelSyntheticClickEvents(false);
 
+import {announce} from '../node_modules/@brightspace-ui/core/helpers/announce.js';
+window.D2L.Announce = announce;
+
 import {clearDismissible, setDismissible} from '../node_modules/@brightspace-ui/core/helpers/dismissible.js';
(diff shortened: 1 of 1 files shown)


web-components/bsi.js
```diff
@@ -89,6 +89,9 @@ import 'd2l-users/components/d2l-profile-image.js';
 import { setCancelSyntheticClickEvents  } from '@polymer/polymer/lib/utils/settings.js';
 setCancelSyntheticClickEvents(false);
 
+import {announce} from '../node_modules/@brightspace-ui/core/helpers/announce.js';
+window.D2L.Announce = announce;
```
dlockhart: Will you just switch [D2L.LP.Web.UI.Aria.LogMessage](https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web/UI/Html/Aria/Aria.js?r=61376c12#9) to call this?

dbatiste: Yep, that's what I've tested out locally so far.  Sound ok, or do you think I should use that here?

Legacy also has a method, something like "MakeLiveRegion"... I'm not sure if I should switch those.  For whatever reason, they were not updated to use the MVC LogMessage.

dlockhart: Seems reasonable. `MakeLiveRegion` seems to be doing something else... turning an existing control into a live region, so it's probably best left as-is.

PR #1763, opened 2019-12-12
Updates to dialog CSS selectors.

dbatiste:
These updates are needed so that these styles do not target internal elements of d2l-dialog web components when shadow-DOM is not supported (Edge). 

The classes that are in conflict are:

* d2l-dialog
* d2l-dialog-inner
* d2l-dialog-footer

A couple updates for the d2l-dialog-inline were needed to maintain specificity for proper dimensions of MVC inline dialogs.

### DIFF
--- a/sass/dialog.scss
+++ b/sass/dialog.scss
@@ -65,7 +65,7 @@ $dialog-resize: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.
 	to   { transform: translateY(100%); }
 }
 
-.d2l-dialog {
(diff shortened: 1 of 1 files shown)


sass/dialog.scss
```diff
@@ -108,14 +108,14 @@ $dialog-resize: url("data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.
 	padding: 1.5rem;
 }
 
-.d2l-dialog-inline {
+.d2l-dialog-mvc.d2l-dialog-inline {
```
dbatiste: This is needed to override the `width: 400px` above.

PR #1764, opened 2019-12-12
Bumping activities to 3.50.23

cbusmand2l:
fyi @dlockhart, @BraydonKains, @m6jones

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2189,9 +2189,9 @@
       }
     },
     "aws-sdk": {
-      "version": "2.588.0",
-      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.588.0.tgz",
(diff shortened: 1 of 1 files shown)


AlexBedley:
@dlockhart I think https://github.com/BrightspaceUI/core/pull/309 is causing some weirdness for this PR. Do I need to run something here to fix it?

dlockhart:
Yep sorry I intended to release that and an update to BSI at almost the same time but got delayed. If you can wait 5 minutes I'll have a fix up.

AlexBedley:
> Yep sorry I intended to release that and an update to BSI at almost the same time but got delayed. If you can wait 5 minutes I'll have a fix up.

Thanks for taking a look! No worries

dlockhart:
Ok fixed -- I restarted your build.

PR #1773, opened 2019-12-16
Add BSI unbundle script

DiljotSG:
Created a script to unbundle your BSI build.


dlockhart:
Sorry, what's the script for exactly? Like why would someone use it?

DiljotSG:
@dlockhart The script is for unbundling your BSI build which has quite a few steps and can be a pain to do repeatedly. (Especially when you do a `full_all` and need to unbundle your BSI once again, when you are doing front-end work.)

dlockhart:
Ah ok... it's assuming that your Windows LMS machine and the machine that BSI is running on are the same, so I guess it's only useful for Windows users.

DiljotSG:
@dlockhart You can actually run it separately for just the front-end or back-end. So if you had a Windows VM with an LMS you could run it in the VM to update your config and restart IIS, and then run it again in your other OS with your BSI components to link them, etc.

dmackenz:
Tested this, works very well.

README.md
```diff
@@ -74,6 +74,47 @@ There is experimental support for running an unbundled BSI to improve the local
...
+```shell
+python unbundle.py --component-path="<path of component to unbundle>"
+```
+
+For example, if I wanted to unbundle my Activities repo, the command could look like this.
+```shell
+python unbundle.py --component-path="C:\Users\d2l-employee\Developer\activities"
+```
+
+#### Advanced Script Usage
```
AlexBedley: it might help to add some example calls using some common combinations of the flags as you have done above for the default case

README.md
```diff
@@ -74,6 +74,47 @@ There is experimental support for running an unbundled BSI to improve the local
 
 The unbundled workflow uses the `@open-wc/es-dev-server` to serve the web components and the other BSI static assets.
 
+### Using the Script to Unbundle your build
+
+You can unbundle your build using the `unbundle.py` script located at the root of the repository.
+
+Make sure you have Python 3.X installed, if not, you can install it using chocolatey (for Windows).
```
AlexBedley: does this work with python 2?

README.md
DiljotSG: Good idea

README.md
DiljotSG: Nope sorry 😢

jeroach: Python 2 is mega old - no concerns from me with Python 3 only

unbundle.py
```diff
@@ -0,0 +1,225 @@
...
+
+    # Run a command and print it
+    def run_command(self, command):
+        self.print_cmd(command)
+        if not self.dry_run:
+            os.system(command)
+
+    # Change directory and print it
+    def change_dir(self, path):
+        self.print_cmd("cd {}".format(path))
```
jeroach: Should there be an `if not self.dry_run:` check here?

unbundle.py
```diff
@@ -0,0 +1,225 @@
...
+        os.chdir(path)
+    
+    # Check if the script is running as admin
+    @staticmethod
+    def is_admin():
+        try:
+            return ctypes.windll.shell32.IsUserAnAdmin()
```
jeroach: Will this work on a non-windows box?

jeroach: Oh, it is only called if we confirm that the user is on Windows - fair enough

DiljotSG: Yeah you can ask the script just to do the `--front_end_only` if you're running it on Mac, if you're doing any of the LMS stuff for the unbundling that implies Windows.

unbundle.py
```diff
@@ -0,0 +1,225 @@
...
+
+            # Delete .npmrc
+            try:
+                self.print_cmd("rm .npmrc")
+                os.remove(".npmrc")
+            except:
+                pass
+
+            # Delete node_modules
+            self.remove_dir("node_modules")
```
jeroach: Do we also need to delete `package-lock.json`?

unbundle.py
DiljotSG: Good call

unbundle.py
DiljotSG: I haven't needed to remove that when unbundling.

jeroach: ah I thought that was part of the "refresh BSI" process, but it's actually to fix another type of issue:
https://github.com/Brightspace/brightspace-integration#package-lockjson-refresh

I just got into a habit of doing it everytime I manually unbundled.

Anyway, no need for it here

dmackenz: I would add rm package-lock.json to this script just to be safe.

DiljotSG:
`python -m unittest unbundle.TestUnbundler` will run the unit tests for you (I will edit the README to have this command.

jeroach:
@DiljotSG and I chatted offline re:
- moving all `unbundled` stuff to it's own folder
- using python modules to allow the tests to exist in a separate file

this should also fix the test issues with the last CI run

jeroach:
I have also tested this out locally - it works :) Full approval from me.

PR #1799, opened 2020-01-07
Bump activities

jeroach:


.npmrc
```diff
@@ -1 +0,0 @@
-engine-strict=true
```
jeroach: @dlockhart this got deleted when I did the `rm -rf node_modules` and `npm i` 

Was that intended?

package-lock.json
```diff
@@ -3576,7 +3576,7 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#8a013dd7c98d2f1a681850b693186619a9e135d9",
+      "version": "github:BrightspaceHypermediaComponents/activities#8ed5de1042d70965ded509bacb18c10b05a58d2d",
```
jeroach: Activities:
https://github.com/BrightspaceHypermediaComponents/activities/commit/8ed5de1042d70965ded509bacb18c10b05a58d2d

.npmrc
dlockhart: I don't think it should go away... AFAIK it's there to make sure NPM stays `<=6.8.0` to avoid all the issues we get with higher versions.

jeroach: kk, thanks - will put it back

package-lock.json
AlexBedley: Github even lets you see the difference between 2 commits: https://github.com/BrightspaceHypermediaComponents/activities/compare/8a013dd7c98d2f1a681850b693186619a9e135d9...8ed5de1042d70965ded509bacb18c10b05a58d2d

jeroach: cool :)

PR #1801, opened 2020-01-07
upgrading to core 1.28

dlockhart:
This picks up the version of core that doesn't have our Intl stuff in `LocalizeMixin`, as well as the version of activities that no longer uses them.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -937,9 +937,9 @@
       }
     },
     "@brightspace-ui/core": {
-      "version": "1.27.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -3576,12 +3576,13 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#8ed5de1042d70965ded509bacb18c10b05a58d2d",
+      "version": "github:BrightspaceHypermediaComponents/activities#5d8da68ca8e2a56095643c040a8061550ae3a8c8",
```
dlockhart: @jeroach FYI this picks up my recent change

PR #1807, opened 2020-01-08
Updated d2l-navigation

Scy-D2L:


Scy-D2L:
Everything switching from semvers to commits makes me suspect there's a problem here. Double checking things...

svanherk:
What node and npm version are you using?  I've seen this before with some other versions, what tends to work for me is node 9-11 and only npm version `6.8.0`

Scy-D2L:
> What node and npm version are you using? I've seen this before with some other versions, what tends to work for me is node 9-11 and only npm version `6.8.0`

That was the issue. Had a bunch of problems getting the exact version of npm because nvm-windows ruins npm updates but this looks much better.

package-lock.json
```diff
@@ -3907,7 +3907,7 @@
       }
     },
     "d2l-html-editor": {
-      "version": "github:Brightspace/d2l-html-editor#b515597c4f0dda90fbcfe3b9f67e6207a6d2d9d1",
+      "version": "github:Brightspace/d2l-html-editor#a51bf21f50abb95c893f1ef020c61603575c7d22",
```
dlockhart: @ianlaidlaw FYI

svanherk:
> That was the issue. Had a bunch of problems getting the exact version of npm because nvm-windows ruins npm updates but this looks much better.

Oh yes, updating npm when using nvm-windows is such a pain.  Not sure what you did, I tend to do the "keep deleting the files it complains about until it finally works" method 😒

Scy-D2L:
> > That was the issue. Had a bunch of problems getting the exact version of npm because nvm-windows ruins npm updates but this looks much better.
> 
> Oh yes, updating npm when using nvm-windows is such a pain. Not sure what you did, I tend to do the "keep deleting the files it complains about until it finally works" method 😒

The magic of google: https://gist.github.com/johnmcase/d31b799b9030327091a0e74880e4c530 a script that handles it.

PR #1820, opened 2020-01-14
US111909 Update activities and siren-sdk

ryantmer:
Updates the activity editor to be support audio note attachments.

Also includes a bunch of updates to Babel, apparently.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,9 +4,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@babel/cli": {
-      "version": "7.8.0",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.8.0.tgz",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -3608,7 +3608,7 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#5fb12f05bd58ca1f52c28b7801617ab8b1f18a24",
+      "version": "github:BrightspaceHypermediaComponents/activities#239300ed794782fe1a5aa8b210355fa3bce23320",
```
AlexBedley: fyi @dmackenz @vix633

ryantmer: ah whoops, thanks Alex - saw there was an update just before this so figured it was only my changes in it!

PR #1828, opened 2020-01-16
bumping bsi

dmackenz:


package-lock.json
```diff
@@ -4084,65 +4020,6 @@
         "d2l-typography": "github:BrightspaceUI/typography#semver:^7"
       }
     },
-    "d2l-my-courses": {
```
AlexBedley: this seems suspicious... it's still listed in package.json

svanherk: ➕➕ Yeah something definitely seems wrong here... 🤔

PR #1833, opened 2020-01-20
Bump d2l-rubric

rgoessl:


djwester:
does this need to go into the 20.20.2 branch as well?

rgoessl:
> 
> 
> does this need to go into the 20.20.2 branch as well?

No, just small patches. Nothing in 20.20.2 is broken

PR #1834, opened 2020-01-21
Bumping core version to 1.28.4

cbusmand2l:
FYI @dlockhart, @mdulat, @BhamelinD2L, @m6jones, @rgoessl


jeroach:
Maybe I'm blind, but I don't see a bump to `core`

cbusmand2l:
Closing for now while we figure out why core didn't update its version.

PR #1871, opened 2020-02-06
bump activities

mdgbayly:


mdgbayly:
Actually just realized might need to bump `siren-sdk` for your changes too. Also someone else has a PR in the works to bump BSI which is picking up some of our changes too. Will validate we have everything updated that we need.

kieranderson:
Looks like that one is also bumping `siren-sdk`

mdgbayly:
> Looks like that one is also bumping `siren-sdk`

Yep - just checking where they are - as they have all the changes we need. Will just close this one if that one gets through.

dlockhart:
FYI this is going to conflict with #1873, which also bumps activities.

PR #1877, opened 2020-02-12
Adding user-feedback to BSI

MykolaGalian:
Adding component github.com/BrightspaceUILabs/user-feedback for using like web-components for feedback dialog from MVC page (according to US111466).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -987,6 +987,22 @@
         "lit-element": "^2.2.1"
       }
     },
+    "@brightspace-ui-labs/user-feedback": {
+      "version": "github:BrightspaceUILabs/user-feedback#6b74fbf37a893e4371e3489bbe7477c97415a7ea",
+      "from": "github:BrightspaceUILabs/user-feedback#semver:^1",
+      "dev": true,
+      "requires": {
+        "@brightspace-ui/core": "^1",
+        "@webcomponents/webcomponentsjs": "^2",
+        "d2l-alert": "github:BrightspaceUI/alert#semver:^4",
+        "d2l-fetch": "github:Brightspace/d2l-fetch#semver:^2",
+        "d2l-hypermedia-constants": "^6.30.0",
+        "d2l-inputs": "github:BrightspaceUI/inputs#semver:^2",
+        "d2l-loading-spinner": "github:BrightspaceUI/loading-spinner#semver:^7",
+        "lit-element": "^2",
+        "siren-parser": "^8.2.0"
+      }
+    },
     "@brightspace-ui/core": {
       "version": "1.29.5",
       "resolved": "https://registry.npmjs.org/@brightspace-ui/core/-/core-1.29.5.tgz",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -110,6 +110,7 @@
     "d2l-tabs": "BrightspaceUI/tabs#semver:^1",
     "d2l-tooltip": "BrightspaceUI/tooltip#semver:^3",
     "d2l-users": "BrightspaceHypermediaComponents/users#semver:^2",
+    "@brightspace-ui-labs/user-feedback": "BrightspaceUILabs/user-feedback#semver:^1",
```
dlockhart: Most of the labs components deploy to NPM, although I see it's not configured to do so. If you'd like to set that up, I can help.

package-lock.json
```diff
@@ -4046,8 +4062,8 @@
       }
     },
     "d2l-hypermedia-constants": {
-      "version": "github:Brightspace/d2l-hypermedia-constants#3d2291ef284591d26e2ea93bb76e363a0bbf7814",
-      "from": "github:Brightspace/d2l-hypermedia-constants#semver:^6",
+      "version": "6.32.0",
+      "resolved": "github:Brightspace/d2l-hypermedia-constants#3d2291ef284591d26e2ea93bb76e363a0bbf7814",
```
dlockhart: Hmm, this doesn't look quite right. Looks like it is [on NPM](https://www.npmjs.com/package/d2l-hypermedia-constants) but nobody else is referencing it from there -- they're all referencing it from GitHub. For now, could you do the same? Ideally everyone uses the NPM reference but we can't have a mix of both.

package-lock.json
```diff
@@ -987,6 +987,22 @@
...
+      "from": "github:BrightspaceUILabs/user-feedback#semver:^1",
+      "dev": true,
+      "requires": {
+        "@brightspace-ui/core": "^1",
+        "@webcomponents/webcomponentsjs": "^2",
+        "d2l-alert": "github:BrightspaceUI/alert#semver:^4",
+        "d2l-fetch": "github:Brightspace/d2l-fetch#semver:^2",
+        "d2l-hypermedia-constants": "^6.30.0",
+        "d2l-inputs": "github:BrightspaceUI/inputs#semver:^2",
+        "d2l-loading-spinner": "github:BrightspaceUI/loading-spinner#semver:^7",
```
dlockhart: Loading spinner and inputs are both in `@brightspace-ui/core`. The old `d2l-inputs` and `d2l-loading-spinner` repos are archived.

dlockhart: Also, `webcomponentsjs` should be a `devDependency`, not an actual dependency.

package.json
MykolaGalian: Thanks, please help

package-lock.json
MykolaGalian: ok

package-lock.json
MykolaGalian: ok

PR #1880, opened 2020-02-12
removing unused CSS for vui-dropdown-menu

dlockhart:
All of this was used before we switched the monolith over to use our menu web components. Now that the feature flag is gone, it's not needed.

Since she's the expert, I'll wait to merge this and the corresponding PR in the monolith until Margaree has had a chance to review.


sass/deprecated/context-menu.scss
```diff
@@ -1,63 +0,0 @@
...
-
-.vui-dropdown-menu > ul > li.d2l-contextmenu-item-hover-next-visible {
-	border-top: 1px solid transparent;
-}
-
-.d2l-contextmenu-ph-anchor {
-	@include _vui-dropdown-context-menu();
-}
-
-.d2l-contextmenu-ph {
```
dlockhart: The `d2l-contextmenu-ph` CSS _is_ used, but I'm just going to move these 2 lines into the monolith.

dlockhart:
Removing this resulted in a 6% reduction in the uncompressed and compressed size of `bsi.css`!

PR #1900, opened 2020-02-21
upgrading rollup plugins and preparing for multi-stage build

dlockhart:
I'm going to be porting this existing rollup build process over so that it handles more and more of the BSI build. That'll eventually allow us to move the web component build in here as well and abandon Polymer CLI, enabling unbundled builds and using `es-dev-server` for dev "builds".

I've tested this by doing a side-by-side compare of the output before & after -- it was identical.


rollup/rollup.config.js
```diff
@@ -1,18 +1,16 @@
...
 	],
 	output: {
+		dir: 'build',
 		format: 'iife',
 		name: 'BSI',
 		sourcemap: true
 	}
 };
+
+export default legacyConfig;
```
dlockhart: Eventually this is going to export an array of configs, so this is the first of multiple.

PR #1905, opened 2020-02-25
Rollup for CSS/Sass

dlockhart:
Moving the CSS/Sass build steps and minification to use rollup.

There's a breaking change here that removes the concept of having a non-minified version of `bsi.js` and the CSS files, so this won't merge until a corresponding PR in the monolith merges.


package.json
```diff
@@ -4,20 +4,13 @@
   "private": true,
   "main": "src/index.js",
   "scripts": {
-    "autoprefix": "postcss ./build/bsi.css -o ./build/bsi.css -u autoprefixer",
     "start": "cross-env BROWSERSLIST_CONFIG=./.browserslistrc-es-dev-server es-dev-server --config=./es-dev-server.config.js --compatibility auto --watch",
     "build": "npm run build:wc && npm run build:common",
     "build-no-es5": "npm run build:wc-no-es5 && npm run build:common",
-    "build:common": "npm run build:js && npm run build:sass && npm run build:icons && npm run build:navicons && npm run build:email-icons && npm run build:babelHelpers && npm run build:esm-amd-loader && npm run build:regenerator-runtime",
-    "build:js": "rollup -c ./rollup/rollup.config.js",
+    "build:common": "rollup -c ./rollup/rollup.config.js && npm run build:icons && npm run build:navicons && npm run build:email-icons && npm run build:babelHelpers && npm run build:esm-amd-loader && npm run build:regenerator-runtime",
```
dlockhart: I merged `build:js` into `build:common` since really that's what it is now... and when I'm done `build:common` will just be rollup stuff.

rollup/rollup.config.js
```diff
@@ -1,24 +1,46 @@
...
+				use: [
+					['sass', {outputStyle: 'expanded'}]
+				]
+			})
+		],
+		output: {
+			file: `./build/${name}.css`
+		}
+	};
+});
```
dlockhart: Rollup isn't super well suited to only handling CSS files on their own -- they tend to want you to import them from your JS. Having a config per file was the recommended way to do this, so I used this map syntax to avoid a lot of duplication.

rollup/rollup.config.js
```diff
@@ -1,24 +1,46 @@
...
+				extract: true,
+				minimize: true, /* uses cssnano */
+				plugins: [autoprefixer()],
+				use: [
+					['sass', {outputStyle: 'expanded'}]
+				]
+			})
+		],
+		output: {
+			file: `./build/${name}.css`
```
dbatiste: Looks like this will assume unique names across specified paths.  A scenario easily spotted in the list of paths if it were to come up though.

dlockhart: It does yeah. I don't really see us adding any more Sass stuff in here beyond what's already there.

PR #1907, opened 2020-02-26
Update frau-publisher

ciss1995:
This PR 
- Updates the version of frau-publisher
- Call frau-publisher without passing aws keys and secrets because frau-publisher can lookup the aws key, secret and session token directly from the env variables.


j3parker:
Should this go too? https://github.com/Brightspace/brightspace-integration/blob/master/.travis.yml#L18

ciss1995:
@j3parker We should be able to remove `CDN_SECRET`. It does not seem to be used anywhere else.

package-lock.json
```diff
@@ -6367,9 +6367,9 @@
       }
     },
     "frau-publisher": {
-      "version": "2.7.10",
-      "resolved": "https://registry.npmjs.org/frau-publisher/-/frau-publisher-2.7.10.tgz",
-      "integrity": "sha512-U+NA5FZPWKjcOPXXXN2YcyygRWH4+79fTxtXkeL4sHfDceU/DSuq1P8nRWlcsmKm/lRKdGNsPiy9CuYN3XZY3w==",
+      "version": "2.7.11",
+      "resolved": "https://registry.npmjs.org/frau-publisher/-/frau-publisher-2.7.11.tgz",
+      "integrity": "sha512-PyhdIKvPyQVuOPJjKJ8HhY4L+wHn41pO1+MDF1FQLRtBjIJZdEqaPzKqPC3KNHR0TWT6ctWgrL9Z73xpXuSoHQ==",
```
dlockhart: I get a different set of dependencies listed out here when I do a fresh package lock reset...

PR #1911, opened 2020-02-26
Set a version for awards leaderboard

sgrinwis-d2l:
Not having a version set for awards leaderboard was causing issues


package.json
```diff
@@ -79,7 +80,7 @@
     "d2l-activity-alignments": "Brightspace/d2l-activity-alignments.git#semver:^1",
     "d2l-activity-exemptions": "Brightspace/d2l-activity-exemptions#semver:^2",
     "d2l-alert": "BrightspaceUI/alert#semver:^4",
-    "d2l-awards-leaderboard-ui": "Brightspace/awards-leaderboard-ui",
+    "d2l-awards-leaderboard-ui": "github:Brightspace/awards-leaderboard-ui#semver:0.1",
```
dlockhart: You shouldn't need the `github:` as it recognizes `OrgName/repoName`.

PR #1914, opened 2020-02-26
Rollup to copy files

dlockhart:
This PR switches BSI from using the `cpy-cli` build script to copy various icons and files to using a rollup plugin. Not only does this streamline the process, but will allow us to set up a watch & repeat portions of it when files change.

This does **NOT** touch the email-icons script, which converts some of our SVGs to PNGs and copies those -- I'll be tackling that task separately.

I did a before & after filesystem diff and this results in no changes.


package-lock.json
```diff
@@ -4995,7 +5014,7 @@
       }
     },
     "d2l-sequences": {
-      "version": "github:BrightspaceHypermediaComponents/sequences#7e893c7cade327ab5f4ccb6ac86d8472c2c386fc",
+      "version": "github:BrightspaceHypermediaComponents/sequences#120a14fa9558dd01f7964b33801685c6b406130e",
```
dlockhart: @sarahboyd FYI

PR #1916, opened 2020-02-27
checking email icons into source so they get built manually

dlockhart:
This build step takes about 10 seconds, and only really needs to run whenever any of our icons change -- which is almost never. So this makes it a manual script and checks the icons themselves into source control.


cli/build-email-icons.js
```diff
@@ -20,7 +20,7 @@ const svg2png = require('svg2png');
 	fs.readFile(`node_modules/d2l-icons/images/${filePath}.svg`)
```
svanherk: I was going to update this path in my PR, but might as well do it here if we ever need to regenerate these?

dlockhart: Good call -- updated.

package-lock.json
```diff
@@ -3167,12 +3165,6 @@
         }
       }
     },
-    "call-me-maybe": {
```
dlockhart: LOL I'm sad this is no longer a dependency... but here's my number....

PR #1922, opened 2020-02-27
DS7123 20.20.4 (master)

mpharoah-d2l:


package-lock.json
```diff
@@ -4917,7 +4917,7 @@
       }
     },
     "d2l-rubric": {
-      "version": "github:Brightspace/d2l-rubric#8777e25a2bc2f614a269cab1af61bff3f13b238e",
+      "version": "github:Brightspace/d2l-rubric#b2f0a3833cb8406386d055de4dd836ab9d269978",
```
jordankay13: Gunna be honest - don't know what this is doing since I'm not familiar with BSI workflows.

However, this string is the same as the 20.20.3 version. Is that ok?

mpharoah-d2l: Yeah, there were no other changes in d2l-rubric since 20.20.3, so both the 20.20.3 and master BSIs will point to the same commit hash in d2l-rubric

PR #1948, opened 2020-03-12
DE38110 backport pdf printing to 20.20.2

ChrisKraljevicD2L:


package-lock.json
```diff
@@ -4454,7 +4454,7 @@
       }
     },
     "d2l-sequences": {
-      "version": "github:BrightspaceHypermediaComponents/sequences#5036a5017b1a118f3283e1fd246a4bfbb789d651",
+      "version": "github:BrightspaceHypermediaComponents/sequences#361de81e809cbd961ff39cfbb8414c20fc5f2247",
```
sarahboyd: should we be backporting your sequences changes to the sequences release branch instead of using master here? I think there might be some stuff on master we don't want in 20.20.2

ChrisKraljevicD2L: Good point. I'll do that :+1:

ChrisKraljevicD2L:
I don't know if this should get

PR #1990, opened 2020-03-25
DE38224 - Bump d2l-activity-alignments - 20.20.4

lvyuan95-d2l:


lvyuan95-d2l:
If a new dependency has been introduced, just add that to `package.json` and that's should be good to go, right?

rgoessl:
> If a new dependency has been introduced, just add that to `package.json` and it should be good to go, right?

No, you have to add it to package.json, delete the package-lock, then run `npm i` to make sure that anything else using that dependency is referring to the same one

package.json
```diff
@@ -52,6 +52,7 @@
     "@polymer/esm-amd-loader": "^1.0.3",
     "@polymer/lit-element": "^0.6.1",
     "@polymer/polymer": "^3.2.0",
+    "@polymer/iron-collapse": "^3.0.1",
```
rgoessl: I think this should be removed

mpharoah-d2l: It's a new dependency, so I think this is correct?

package-lock.json
```diff
@@ -4176,7 +4176,7 @@
       }
     },
     "d2l-activity-alignments": {
-      "version": "github:Brightspace/d2l-activity-alignments#90107c3d532f67003a4afb0748430dedd8193d71",
+      "version": "github:Brightspace/d2l-activity-alignments#ddeae1fc858e9f0ccaf0e6eae2559e2f22cfecd9",
```
mpharoah-d2l: This is using the same version as the 20.20.5 one. I this correct, or should we be targeting a different branch?

package.json
rgoessl: Other repos use it, so it's not new to BSI

package-lock.json
lvyuan95-d2l: I think we want those fixes on both 20.20.4 and master, not sure if anything in cert 20.20.4 will be merged in master automatically?

lvyuan95-d2l:
So the weird thing is that `@polymer/iron-collapse` has been used in other components as well as in `package-lock.json`, but not in `package.json`

Should I just bump the `d2l-activity-alignments` and add `@polymer/iron-collapse` to its `requires` in `package-lock.json`?

rgoessl:
> Should I just bump the `d2l-activity-alignments` and add `@polymer/iron-collapse` to its `requires` in `package-lock.json`?

If you bump it using `npm i` it should automatically update all dependencies. But you would have to delete the package-lock first

PR #1992, opened 2020-03-25
DE38224 - Bump d2l-activity-alignments - 20.20.5

lvyuan95-d2l:
Bump the version in master too as the picker is in the master branch already

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "@polymer/esm-amd-loader": "^1.0.3",
     "@polymer/lit-element": "^0.6.1",
     "@polymer/polymer": "^3.2.0",
-    "@polymer/iron-collapse": "^3.0.1",
     "@rollup/plugin-commonjs": "^11",
     "@webcomponents/webcomponentsjs": "^2.0.4",
     "autoprefixer": "^9",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
     "@polymer/esm-amd-loader": "^1.0.3",
     "@polymer/lit-element": "^0.6.1",
     "@polymer/polymer": "^3.2.0",
+    "@polymer/iron-collapse": "^3.0.1",
```
svanherk: Do you need to use this directly in the lms somewhere?  If not, you don't need to add to the `package.json` file, this should be added automatically in `package-lock.json` when you follow the refresh steps (https://github.com/Brightspace/brightspace-integration#package-lockjson-refresh)

svanherk: Hmmm, the reason it might not be is because it's listed under `devDependencies` here: https://github.com/Brightspace/d2l-activity-alignments/blob/master/package.json#L24
If you use it in a component, you should move it to the `dependencies` section

lvyuan95-d2l: Thanks Stacey! Also should I manually add `"@polymer/iron-collapse": "^3.0.1"` to `requires` sections of the component too?

svanherk: Yeah those steps will do that for you (and anything else you may need), but the versions listed there are important (we're trapped on `npm 6.8` for now).  Let me know if you have any issues and I can try to help troubleshoot!

PR #1997, opened 2020-03-26
Update package-lock.json

djwester:


djwester:
@lvyuan95-d2l  @max-hao-d2l 

you both currently have open prs to merge changes into these repos. Do you want to dump those in favour of this one? It should have your changes in it I think

lvyuan95-d2l:
I think we still need to make the dependency change for `d2l-activity-alignments`

PR #2011, opened 2020-03-31
Add oslo to bsi build

ppaskaris-d2l:
This PR adds scripts the generates an XML language package for the included web components and a manifest for the Olso component in the LMS. The build artifacts will be committed into the monolith by the bsi-sync jenkins job.

We have a mapping file of component to serge file, which we then parse and traverse to discover all of the terms. We take care to encode the term names for the LMS, and to use a more or less stable identifier for the collection names so that components can be rejiggered (within reason...). This is similar to what RE's thing does, but less complicated because it doesn't actually implement any translation workflow.

Caveats: only works if the serge config is present in the node_modules directory, which happens to always be the case if it's a git dependency 🎉.

### DIFF
--- a/oslo/generate-monolith-xml
+++ b/oslo/generate-monolith-xml
@@ -1,0 +1,109 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const fs = require('fs').promises;
+const path = require('path');
+
+const Lang = require('./Lang');
+const LangPackage = require('./LangPackage');
+const Serge = require('./Serge');
+
+const Languages = [
+	new Lang('ar-sa', 'ar', 'ar-SA'),
+	new Lang('da-dk', 'da', 'da-DK'),
+	new Lang('de-de', 'de', 'de-DE'),
+	new Lang('es-mx', 'es', 'es-MX'),
+	new Lang('fr-ca', 'fr', 'fr-CA'),
+	new Lang('ja-jp', 'ja', 'ja-JP'),
+	new Lang('ko-kr', 'ko', 'ko-KR'),
+	new Lang('nl-nl', 'nl', 'nl-NL'),
+	new Lang('pt-br', 'pt', 'pt-BR'),
+	new Lang('sv-se', 'sv', 'sv-SE'),
+	new Lang('tr-tr', 'tr', 'tr-TR'),
+	new Lang('zh-cn', 'zh', 'zh-CN'),
+	new Lang('zh-tw', 'zh-TW', 'zh-TW'),
+];
+const PackageName = 'WebComponents';
+const TranslationReplacement = '%CULTURE:LC%';
+
+const DefinitionsPath = path.join(__dirname, '../build/langterms/definitions/', PackageName + '.xml');
+const ConfigPath = path.join(__dirname, '../build/langterms/D2L.LP.Oslo.config.json');
+const MappingPath = path.join(__dirname, '../.serge-mapping.json');
+const ModulesRoot = path.join(__dirname, '../node_modules');
+const TranslationPath = path.join(__dirname, '../build/langterms/translations/', TranslationReplacement, PackageName + '.xml');
+
+async function main() {
+
+	const mappingText = await fs.readFile(MappingPath, 'utf-8');
+	const mappingObject = JSON.parse(mappingText);
+	const mapping = Object.entries(mappingObject);
+
+	const definitions = new LangPackage(PackageName);
+	const manifest = [];
+	const translations = new Map();
+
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -1,7 +1,9 @@
 language: node_js
 node_js:
-- "11"
+- "12"
```
ppaskaris-d2l: So this is a noteworthy and unrelated change that I can do as a separate PR first, but I wrote my code assuming v12 because we're otherwise up to date, and I took advantage of that before realizing that the build wasn't.

oslo/generate-serge-mapping
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env node
```
kieranderson, line 1: I see that this is run by `build:serge-mapping`, which isn't specifically called out in the build steps. Did you imagine running this every time something was added with terms to update the serge-mapping file? In the `readme` you specify to just update the serge-mapping file directly.

ppaskaris-d2l: I imagined running this when all of the component dependencies work. That is not currently the case.

oslo/generate-monolith-xml
```diff
@@ -0,0 +1,109 @@
...
+	new Lang('es-mx', 'es', 'es-MX'),
+	new Lang('fr-ca', 'fr', 'fr-CA'),
+	new Lang('ja-jp', 'ja', 'ja-JP'),
+	new Lang('ko-kr', 'ko', 'ko-KR'),
+	new Lang('nl-nl', 'nl', 'nl-NL'),
+	new Lang('pt-br', 'pt', 'pt-BR'),
+	new Lang('sv-se', 'sv', 'sv-SE'),
+	new Lang('tr-tr', 'tr', 'tr-TR'),
+	new Lang('zh-cn', 'zh', 'zh-CN'),
+	new Lang('zh-tw', 'zh-TW', 'zh-TW'),
```
dlockhart, line 25: So I guess we'll just have to be vigilant when new languages are added to the monolith to make sure we add them here as well. And to components, which will be fun.

oslo/generate-monolith-xml
```diff
@@ -0,0 +1,109 @@
...
+];
+const PackageName = 'WebComponents';
+const TranslationReplacement = '%CULTURE:LC%';
+
+const DefinitionsPath = path.join(__dirname, '../build/langterms/definitions/', PackageName + '.xml');
```
dlockhart, line 30: Nit: string literals are fun:

    `${PackageName}.xml`

package.json
```diff
@@ -5,14 +5,16 @@
...
-    "build": "npm run build:wc && npm run build:common",
+    "build": "npm run build:wc && npm run build:common && npm run build:lang",
     "build-no-es5": "npm run build:wc-no-es5 && npm run build:common",
     "build:common": "rollup -c ./rollup/rollup.config.js && npm run build:babelHelpers",
     "build:babelHelpers": "babel-external-helpers > ./build/babel-helpers.js",
     "build:rename-define": "node ./cli/rename-define.js \"./build/**/*.js\"",
     "build:email-icons": "node ./cli/build-email-icons.js",
     "build:wc": "cross-env NODE_OPTIONS=--max_old_space_size=4096 polymer build",
     "build:wc-no-es5": "cross-env NODE_OPTIONS=--max_old_space_size=4096 polymer build --name \"es6-bundled\"",
+    "build:lang": "node ./oslo/generate-monolith-xml",
```
dlockhart, line 16: What's the speed of this script like? Since I'm guessing we don't need it for local BSI builds, wondering if it's something we should just call directly from `.travis.yml` and not make it part of the generic `build` task.

oslo/generate-monolith-xml
ppaskaris-d2l: We have the same issue here: https://github.com/BrightspaceUI/intl/blob/master/lib/common.js#L3

I did consider sourcing it, but I needed the capitalization.

package.json
ppaskaris-d2l: If you blink, you will miss it.

ppaskaris-d2l:
> Was this the area you were looking at adding those tests? Those parsers seem like a great spot to add some comprehensive tests.

No, I was adding tests to the JavaScript that will be included with the product.

ppaskaris-d2l:
Okiedokie, I wrote a test and I caught a bug in the JS parsing. Curious what kind of unescaping Serge does, but for now, we only have quotes escaped, so I'm satisfied. The other parsers just use JSON.parse so I'm not too worried about them.

PR #2021, opened 2020-04-03
added .nvmrc to 12.13

maboelsoudd2l:
This should help align devs to run bsi with only one version of node.

CAVEAT: since BSI restricts npm < 6.8.0, `npm i` will fail immediately because nvm automatically installs the latest version of npm. So devs will have to `npm i -g npm@6.8.0`. I wish .nvmrc would have the ability to specify an npm version. If you guys have any suggestions let me know because this outcome is not ideal 😄


ppaskaris-d2l:
Cool but can it do "12.x" or "latest LTS" as well? Why 12.13 specifically? Isn't the current 12.16.1?

maboelsoudd2l:
so the instructions now to get it BSI to install is:
```
nvm install
nvm use
npm i -g npm@6.8.0
npm install
```

PR #2059, opened 2020-04-28
upgrading various dependencies to latest

dlockhart:


package-lock.json
```diff
@@ -4794,7 +4792,7 @@
       }
     },
     "d2l-my-courses": {
-      "version": "github:Brightspace/d2l-my-courses-ui#5f7538a42ffb6cf3b5eaa691ff8b9626ee557a10",
+      "version": "github:Brightspace/d2l-my-courses-ui#c33c71aae0db59d7b6ee452ae192f5974b696b6f",
```
dlockhart, line 4795: @svanherk FYI

package-lock.json
```diff
@@ -4890,7 +4888,7 @@
       }
     },
     "d2l-organization-hm-behavior": {
-      "version": "github:Brightspace/organization-hm-behavior#bdb2122bac7890e46a56b8f57345187adf4cb674",
+      "version": "github:Brightspace/organization-hm-behavior#7c48b167c24bce3f499f4340d09924f26e65e472",
```
dlockhart, line 4891: @omsmith FYI

package-lock.json
```diff
@@ -5003,17 +5001,28 @@
       }
     },
     "d2l-program-outcomes-picker": {
-      "version": "github:Brightspace/program-outcomes-picker#8ce7568bed6b526a3e2f94b67abe4ef6c357ec55",
+      "version": "github:Brightspace/program-outcomes-picker#01de38971f89c9d47997e622b33454cedad7d12d",
```
dlockhart, line 5004: @rgoessl FYI

package-lock.json
```diff
@@ -5155,7 +5164,7 @@
       }
     },
     "d2l-simple-overlay": {
-      "version": "github:Brightspace/simple-overlay#b893b64645fb1695837c3c6e3a0836ffb4148e81",
+      "version": "github:Brightspace/simple-overlay#62dd71f55cd17711afcc77471073e268186e260d",
```
dlockhart, line 5167: @svanherk FYI

package-lock.json
```diff
@@ -5003,17 +5001,28 @@
...
-        "d2l-button": "github:BrightspaceUI/button#semver:^5",
-        "d2l-icons": "github:BrightspaceUI/icons#semver:^6",
-        "d2l-inputs": "github:BrightspaceUI/inputs#semver:^2",
-        "d2l-loading-spinner": "github:BrightspaceUI/loading-spinner#semver:^7",
+        "@brightspace-ui/core": "^1.43.2",
+        "d2l-alert": "github:BrightspaceUI/alert#semver:4.13.0",
         "lit-element": "^2.2.1"
+      },
+      "dependencies": {
+        "d2l-alert": {
```
svanherk, line 5013: This isn't good...

package-lock.json
```diff
@@ -16080,6 +16087,554 @@
           "resolved": "https://registry.npmjs.org/fs.realpath/-/fs.realpath-1.0.0.tgz",
           "integrity": "sha1-FQStJSMVjKpA20onh8sBQRmU6k8="
         },
+        "fsevents": {
```
svanherk, line 16090: So Allan and I were a bit concerned with this giant chunk last week but we were both getting it too, so hopefully its fine?

package-lock.json
dlockhart: Yeah, I think it's because [alert is pinned](https://github.com/Brightspace/program-outcomes-picker/blob/master/package.json#L13). @rgoessl was this done for a reason? Could it be `BrightspaceUI/alert#semver:^4` like the others?

dlockhart: Also, I wouldn't recommend having a `package-lock` in `program-outcomes-picker` since it's just a dependency of other things, not an app itself. Usually the lock files are for the top-level thing (the app) that needs to manage what versions of what gets built together.

svanherk: I found this: https://github.com/Brightspace/program-outcomes-picker/commit/71d6f6a60e8ecbbd60f2bb70cbdb9ac44af0960e
@mpharoah-d2l Do you have context here?  We can only have one version of `d2l-alert` in BSI

mpharoah-d2l: Demo page was broken when using the latest version of d2l-alert because it didn't seem to play well with Lit. Forgot that this would be an issue when adding to BSI

package-lock.json
```diff
@@ -4580,7 +4578,7 @@
       }
     },
     "d2l-enrollments": {
-      "version": "github:BrightspaceHypermediaComponents/enrollments#61177f648968d6f6965f27dbf9fa4c5e3e66c71d",
+      "version": "github:BrightspaceHypermediaComponents/enrollments#5073a7bcc7559a09163b651e2223d25bf937caab",
```
dlockhart, line 4581: @svanherk FYI

package-lock.json
dlockhart: @rgoessl @svanherk @mpharoah-d2l Ok I've pinned outcomes picker back at `3.0.6` for now. If there are actual issues with the newer releases of alert please let @AllanKerr know and we can resolve them. Once that's all fixed up, feel free to unpin and release your changes into BSI whenever you like.

mpharoah-d2l: I'll do some investigation to see if the latest d2l-alert breaks Lit components for real, of it it only breaks demo pages

package-lock.json
dlockhart: Yeah I think it's just a matter of the `polyfills-loader` (used by the open-wc stuff) pulling in `fsevents` which just happens to have a lot of sub-dependencies.

package-lock.json
svanherk: Hehe yay!  Thanks 😝

package-lock.json
mpharoah-d2l: It does seem to actually break things. Using the latest d2l-alert causes an error about d2l-icon being defined twice, but reverting it to version 4.13.0 fixes it. I'm not completely sure why though

PR #2061, opened 2020-04-28
removing broken no-ES5 build

dlockhart:
I don't think this has been working since Polymer 3. Ripping it out now to simplify the migration to Rollup-based unbundled builds.


package.json
```diff
@@ -4,14 +4,11 @@
   "private": true,
   "main": "src/index.js",
   "scripts": {
-    "build": "npm run build:wc && npm run build:common && npm run build:lang",
-    "build-no-es5": "npm run build:wc-no-es5 && npm run build:common",
-    "build:common": "rollup -c ./rollup/rollup.config.js && npm run build:babelHelpers",
-    "build:babelHelpers": "babel-external-helpers > ./build/babel-helpers.js",
+    "build": "npm run build:wc && npm run build:common && babel-external-helpers > ./build/babel-helpers.js && npm run build:lang",
+    "build:common": "rollup -c ./rollup/rollup.config.js",
```
dlockhart, line 8: I rolled the `babelHelpers` stuff into `build` since they're only required as part of the Polymer builds and soon `build:common` is going to be shared between Polymer and Rollup builds.

PR #2064, opened 2020-04-29
renaming Polymer builds to "legacy"

dlockhart:
The plan is for `npm run build` to eventually run:
1. legacy: today's Polymer build
2. unbundled: the new Rollup-based unbundled builds, and
3. common: things needed by both

Then when we're done with the legacy Polymer build, it'll be easy to just kill that script and refactor.

This change takes `build:wc` and renames it `build:legacy` while adding the babelHelpers (only needed for legacy builds) step and the old post-build step of renaming `define` -> `define2`.

I needed to scope the define-renaming to just the legacy files, since we want it to leave the new unbundled stuff (which will go in `build/wc`) alone when it comes in. That required a tweak to that script so that it could accept multiple globs.

I also moved the `build:lang` step into the common build, since it'll be needed with the Rollup builds too.


rollup/rollup.config.js
```diff
@@ -11,8 +11,7 @@ const jsConfig = {
 		copy({
 			targets: [
 				{src: 'node_modules/@brightspace-ui/core/components/icons/images', dest: './build'},
-				{src: 'email-icons', dest: './build/images'},
-				{src: 'node_modules/@polymer/esm-amd-loader/lib/esm-amd-loader.min.js', dest: './build'}
```
dlockhart, line 15: I moved the `esm-amd-loader` bit out into the `build:legacy` step so that it'll be there for the rename-define stuff. Also it makes it so that this Rollup config doesn't do anything "legacy".

package-lock.json
```diff
@@ -2686,7 +2705,6 @@
           "resolved": "https://registry.npmjs.org/is-accessor-descriptor/-/is-accessor-descriptor-1.0.0.tgz",
           "integrity": "sha512-m5hnHTkcVsPfqx3AKlyttIPb7J+XykHvJP2B9bZDjlhLIoEq4XoK64Vg7boZlVWYK6LUY94dYPEE7Lh0ZkZKcQ==",
           "dev": true,
-          "optional": true,
```
svanherk, line 2689: I feel like I see this `"optional": true,` line come and go from every package over and over again, I'm not sure what causes it

dlockhart: Yeah I know, it's weird.

PR #2066, opened 2020-04-30
removing script to build email icons

dlockhart:
AFAIK, these email icons have never changed, and if they did it would be easy enough to convert the SVGs to PNGs manually.

So... I'd like to remove this script (which no longer gets called as part of the build anyway), and is causing a bunch of extra dependencies to be installed.


dlockhart:
Haha you jinxed it!

PR #2067, opened 2020-04-30
Cert: DE38817 core template width, DE38899 d2l-activities langterms

emil-furniss:


package-lock.json
```diff
@@ -4232,7 +4233,7 @@
       }
     },
     "d2l-activities": {
-      "version": "github:BrightspaceHypermediaComponents/activities#ac21edc082429f2546ae1537ebcc496363aeb881",
+      "version": "github:BrightspaceHypermediaComponents/activities#b87d0edb2622a3fc36b88333dfb2e5fc5f86694f",
```
dlockhart, line 4236: You wanted this change as well I assume?

emil-furniss: yeah, it's the updated langterms for d2l-activities. Sneaking this in here so we don't have to do another release just for that.

PR #2069, opened 2020-04-30
aligning browserlistrc files

dlockhart:
Renaming the `es-dev-server` one as I want to use it for the new open-wc Rollup builds as well.

Also simplified it a bit so that it can be identical to the `.browserlistrc` except with the additional `not IE 11` (open-wc handles IE11 for us).

### DIFF
--- a/.browserslistrc-es-dev-server-legacy
+++ b/.browserslistrc-es-dev-server-legacy
@@ -1,9 +1,0 @@
-last 2 Chrome major versions
-last 2 ChromeAndroid major versions
-last 2 Firefox major versions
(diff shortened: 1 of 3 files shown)


.browserslistrc
```diff
@@ -1,5 +1,6 @@
 last 2 versions
 Firefox ESR
+Edge 18
```
dlockhart, line 3: Technically this isn't needed as for some reason `last 2 versions` still includes legacy-Edge. Probably because they're worried about all the things it's going to break. But I thought I'd include it just to protect us in case they change their minds.

PR #2072, opened 2020-05-01
update activities version

spandanabaruah:


spandanabaruah:
@emil-furniss @astadelbauer @kieranderson

PR #2073, opened 2020-05-01
using rollup to build web components

dlockhart:
Ok, this is (mostly) it! [Corresponding (but not required) PR in the monolith](https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/11286/overview)

This doesn't actually change BSI for developers yet, as running the unbundled build is still a manual step that only happens from Travis. Once it's 100% ready, I'll run it in parallel for a bit, and then later when we're confident in it, it'll be the only build running.

The `build` directory will now look like this:
- `es5-bundled/*`: legacy bundled builds for non-ES6 browsers (IE11)
- `es6-bundled/*`: legacy bundled builds for everyone else
- `unbundled/*`: new unbundled builds
- `images/`, `bsi.js`, `bsi.css`, `datagrid.css`, `homepages.css`, `babel-helpers.js`, `esm-amd-loader.min.js`: static non-web component files that are unrelated

The feature flag in the monolith -- which is still OFF for now -- controls whether assets are served from the bundled or unbundled directories.

There's also "developer" mode, which doesn't even have a build step and instead just serves everything from the local filesystem -- enabling hot reloading. You'll notice that it needs some special handling for the non-web component files. This is very similar to what Martin set up.


.travis.yml
```diff
@@ -7,6 +7,7 @@ before_install: if [[ `npm -v` != 6.8.0 ]]; then npm i -g npm@6.8.0; fi
 script:
 - npm run test:ci
 - travis_wait npm run build
+- npm run build:unbundled
```
dlockhart, line 10: Eventually I'll move this to be part of `npm run build` -- I just don't want to slow down developer builds until it's actually being used.

package-lock.json
```diff
@@ -779,6 +779,18 @@
         "@babel/helper-plugin-utils": "^7.8.3"
       }
     },
```
dlockhart, line 781: Ignore `package-lock` for now. Before merging, I'll do a rebase and re-generate it properly.

package.json
```diff
@@ -9,14 +9,17 @@
...
     "build:lang": "node ./oslo/generate-monolith-xml",
     "build:serge-mapping": "node ./oslo/generate-serge-mapping",
+    "build:unbundled": "BROWSERSLIST_CONFIG=./.browserslistrc-open-wc rollup -c ./rollup/rollup-wc.config.js",
     "lint": "npm run lint:js",
     "lint:js": "eslint *.js web-components/*.js js/** rollup/** test/** cli/** oslo/**.js",
     "publish:autotag": "node ./cli/autotag.js",
     "publish:cdn": "frau-publisher -m lib -t bsi -f \"./build/**/*\" --versionvar TRAVIS_TAG",
     "preserve": "npm run build",
-    "serve": "http-server build --cors",
+    "serve": "npm run start:prod",
```
dlockhart, line 18: Switching the main `serve` command to just use `es-dev-server`'s "prod" configuration, which is essentially a basic HTTP server.

rollup/rollup-wc.config.js
```diff
@@ -0,0 +1,49 @@
...
+
+const componentFiles = [
+	'./web-components/bsi-unbundled.js',
+	'./node_modules/@brightspace-ui/core/components/button/button-icon.js',
+	'./node_modules/d2l-navigation/d2l-navigation-band.js',
+	'./web-components/d2l-opt-in-flyout-webcomponent.js',
+	'./web-components/d2l-scroll-spy.js',
+	'./web-components/d2l-simple-overlay.js',
+	'./web-components/d2l-table.js',
+	'./web-components/d2l-tabs.js',
```
dlockhart, line 11: This list will eventually include EVERY component used in the monolith, which will each be responsible for importing these files.

Also, once the legacy build is gone, I'd like to switch things in `./web-components` that just point at another file to point directly to the asset(s) in `node_modules`. There's no need to create these "fragments" anymore.

web-components/bsi-unbundled.js
```diff
@@ -0,0 +1,68 @@
+import 'fastdom';
```
dlockhart, line 1: `bsi.js` imported a lot of "core" things which it can no longer do, so I needed to create an unbundled version of it. I moved everything "shared" into here.

rollup/rollup-wc.config.js
```diff
@@ -0,0 +1,48 @@
...
+	'./web-components/d2l-sequence-viewer.js',
+	'./web-components/d2l-rubric.js',
+	'./web-components/d2l-rubric-editor.js',
+	'./web-components/d2l-outcomes-level-of-achievements.js',
+	'./web-components/d2l-outcomes-user-progress.js',
+	'./web-components/d2l-program-outcomes-picker.js',
+	'./web-components/d2l-quick-eval.js',
+	'./web-components/d2l-content-store.js',
+	'./web-components/d2l-consistent-evaluation.js',
+	'./web-components/d2l-user-feedback.js',
```
dlockhart, line 34: The one file missing from here that's in `polymer.json` is `mobx`. It was only put into its own bundle because Polymer was dumping it in the shared bundle previously, which we didn't want. Now everything is unbundled anyway, so that's not necessary.

es-dev-server-prod.config.js
```diff
@@ -0,0 +1,18 @@
...
+	port: 8080,
+	compatibility: 'none',
+	dedupe: true,
+	middlewares: [
+		cors()
+	],
+	nodeResolve: false,
+	rootDir: 'build',
+	start: false,
+	watch: false
```
dlockhart, line 17: This is basically a glorified static web server meant to emulate the production build environment -- no transformations just CORS.

rollup/rollup-wc.config.js
```diff
@@ -0,0 +1,48 @@
...
+const config = createBasicConfig({
+	developmentMode: false, /* forces tree-shaking, minify ON */
+	legacyBuild: true, /* required for IE11 and legacy-Edge support */
+	outputDir: 'build/unbundled'
+});
+config.input = componentFiles.concat(appFiles);
+config.output[0].entryFileNames = '[name].js';
+config.output[0].chunkFileNames = '[name].js';
+config.output[1].entryFileNames = 'nomodule-[name].js';
+config.output[1].chunkFileNames = 'nomodule-[name].js';
```
dlockhart, line 46: This makes the out filenames predictable so we know how to import them from the monolith. Otherwise they contain hashes that would normally be used for cache busting I think.

PR #2076, opened 2020-05-04
fixing warning for deprecated oktokit import method

dlockhart:
Was seeing this warning in the build output:
> [@octokit/rest] `const Octokit = require("@octokit/rest")` is deprecated. Use `const { Octokit } = require("@octokit/rest")` instead


dlockhart:
Confirmed that the warning is no longer appearing in the build output.

PR #2109, opened 2020-05-15
update sequences and sequence-viewer version

sarahboyd:


svanherk:
So I think this PR is going to be a problem... you've updated `sequences` to version 2, but `d2l-enrollments` and `d2l-organizations` are still trying to use version 1.  The only reason CI isn't failing is because you manually created this PR instead of using `npm i`

svanherk:
Are there a lot of breaking changes in version 2?  If not, I think the quickest way to fix things would be to update `organizations` and `enrollments`.

sarahboyd:
> Are there a lot of breaking changes in version 2? If not, I think the quickest way to fix things would be to update `organizations` and `enrollments`.

checking with my team right now

PR #2152, opened 2020-05-29
updating script to also apply "dev" environment

dlockhart:
@DiljotSG would you be able to test this out for me? I run BSI from my Mac so the script itself doesn't work for me.

Essentially when the new unbundled builds are ready, `"env": "dev"` needs to get added to the monolith config file, which will do the same thing as `"import-style": "esm"` did. So this sets them both for now, and then when unbundled builds are always-on I'll remove the import-style stuff.


DiljotSG:
@dlockhart I will be happy to test this out for you and report back

dlockhart:
Thanks!

PR #2157, opened 2020-06-01
Awards Leaderboard bump to v1.0.4

cbelan-d2l:


cbelan-d2l:
Merged in after `20.20.7.22674`

PR #2164, opened 2020-06-02
Configure dependabot

j3parker:
See [the docs](https://dependabot.com/docs/config-file/) for more info


j3parker:
Feel free to disable it if it sucks!

PR #2177, opened 2020-06-02
Reduce redundant Travis builds

j3parker:
See [the docs](https://docs.travis-ci.com/user/pull-requests/#double-builds-on-pull-requests)

It looks like each PR does two builds: one for the branch, and one for the PR event. This adds extra load on Travis and causes queuing.

We use this setting elsewhere, for example in [Brightspace/cdn-manager](https://github.com/Brightspace/cdn-manager/blob/master/.travis.yml#L6).

Give it some thought before merging. I don't know the particular details for this repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: node_js
 node_js:
 - "12"
 sudo: false
+branches:
+  only:
+  - master
 # npm > 6.8.0 doesn't support our github semver references, so downgrade
 before_install: if [[ `npm -v` != 6.8.0 ]]; then npm i -g npm@6.8.0; fi
 script:


j3parker:
It looks like this is compatible with your branch protection rule:

image

.travis.yml
```diff
@@ -2,6 +2,9 @@ language: node_js
 node_js:
 - "12"
 sudo: false
+branches:
+  only:
+  - master
```
svanherk, line 7: I think we'd need to add the cert/hotfix build branches here too?

j3parker: Yes

j3parker: It looks like it supports a regular expression. https://circleci.com/docs/2.0/configuration-reference/#branches

What do the branch names look like?

svanherk: Should... seems like we could do something like `/20.*/`

PR #2180, opened 2020-06-02
removing IE11 array polyfills

dlockhart:
`findIndex` support:

Screen Shot 2020-06-02 at 4 11 11 PM

`includes` support:

Screen Shot 2020-06-02 at 4 11 58 PM


dlockhart:
I'm good with it yeah. We were going to block Firefox < 59, Safari < 10.1, Chrome < 61 -- these are a lot older.

PR #2178, opened 2020-06-02
Update deps

dlockhart:
This should handle most of the things Dependabot is complaining about.

PR #2184, opened 2020-06-03
Reverse latest deps

dlockhart:
Undoing the changes from [this pull request](https://github.com/Brightspace/brightspace-integration/pull/2178) until we can figure out which change is causing BSI to fail. I kept `d2l-activities` at the latest version from Emil's change yesterday.


j3parker:
:( ow

dlockhart:
I've also confirmed that a `npm ci` with these changes fixes things. Now the archaeology begins...

PR #2185, opened 2020-06-03
upgrading eslint

dlockhart:
Going to do these over, one by one.


package-lock.json
```diff
@@ -6231,70 +6247,451 @@
...
+        }
+      }
+    },
+    "eslint-config-brightspace": {
+      "version": "0.8.0",
+      "resolved": "https://registry.npmjs.org/eslint-config-brightspace/-/eslint-config-brightspace-0.8.0.tgz",
+      "integrity": "sha512-f6Cp68napCNFVcKndfeEGJ3F8ECryrX037N7WGXDIab+25TbWJecq7UkOkoT61kl+Z8Y5MgkQS8stCy7KJgxMA==",
+      "dev": true,
+      "requires": {
+        "eslint": "^5.15.1"
```
svanherk, line 6452: Looks like we can cleanup some of this dependency duplication if we get `eslint` up-to-date in our config repo at some point

dlockhart: Good catch. @dbatiste how hard do you think that'd be?

PR #2188, opened 2020-06-03
re-upgrading dependencies

dlockhart:
Now that consistent-eval is locked, we can re-introduce this.


omsmith:
would you care to bump frau-publisher in this? :smile:

dlockhart:
Haha I can yeah... haha another 30 minute wait for me!

omsmith:
Thanks

PR #2195, opened 2020-06-04
making profile image unbundled

dlockhart:
Goes along with [this LMS change](https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/12282/overview).

### DIFF
--- a/rollup/rollup-wc.config.js
+++ b/rollup/rollup-wc.config.js
@@ -18,6 +18,7 @@ const componentFiles = [
 	'./node_modules/d2l-navigation/d2l-navigation-main-header.js',
 	'./node_modules/d2l-navigation/d2l-navigation-separator.js',
 	'./node_modules/d2l-navigation/d2l-navigation.js',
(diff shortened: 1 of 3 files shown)


rollup/rollup-wc.config.js
```diff
@@ -18,6 +18,7 @@ const componentFiles = [
 	'./node_modules/d2l-navigation/d2l-navigation-main-header.js',
 	'./node_modules/d2l-navigation/d2l-navigation-separator.js',
 	'./node_modules/d2l-navigation/d2l-navigation.js',
+	'./node_modules/d2l-users/components/d2l-profile-image-base.js',
```
dlockhart, line 21: The `<d2l-profile-image-base>` component is the one that's actually used in the LMS, so including `d2l-profile-image.js` which wraps it didn't seem necessary.

svanherk: Yeah I think Miles did use the `d2l-profile-image` originally, but the delay due to HM loading was seen as not ideal for the profile image, so he changed it back to just the UI component.

PR #2197, opened 2020-06-04
upgrading to rollup v2

dlockhart:
Not sure why we weren't already using v2, it's fully supported by the open-wc stuff.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1455,12 +1455,12 @@
       "dev": true
     },
     "@octokit/plugin-rest-endpoint-methods": {
-      "version": "3.14.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -4524,7 +4517,7 @@
       }
     },
     "d2l-awards-leaderboard-ui": {
-      "version": "github:Brightspace/awards-leaderboard-ui#f10354be714b15d84f81d1f04cd6ad48c19342d4",
+      "version": "github:Brightspace/awards-leaderboard-ui#4554c48e56b64518ee95b079c8e51906d978415b",
```
dlockhart, line 4520: @cbelan-d2l FYI

PR #2226, opened 2020-06-12
more unbundling

dlockhart:
Another round of unbundling... don't expect this will get merged before the flag flips on this weekend. This shaves off about `17 kB` from the typical page's BSI download.

There'll be a corresponding PR in the monolith.


rollup/rollup-wc.config.js
```diff
@@ -6,9 +6,16 @@ import merge from 'deepmerge';
...
+	'./node_modules/@brightspace-ui/core/components/button/floating-buttons.js',
+	'./node_modules/@brightspace-ui/core/components/inputs/input-search.js',
 	'./node_modules/@brightspace-ui/core/components/inputs/input-time.js',
 	'./node_modules/@brightspace-ui/core/components/button/button-icon.js',
+	'./node_modules/@brightspace-ui/core/components/more-less/more-less.js',
+	'./node_modules/@brightspace-ui/core/components/tooltip/tooltip.js',
 	'./node_modules/@d2l/switch/d2l-switch.js',
+	'./node_modules/d2l-activities/components/d2l-subtitle/d2l-subtitle.js',
+	'./node_modules/d2l-button-group/d2l-button-group.js',
+	'./node_modules/d2l-button-group/d2l-action-button-group.js',
```
svanherk, line 18: Minor: Looks like we're trying to keep these alphabetical and the last two should be switched.  Although we seem to break that above with `button-icon`...

PR #2233, opened 2020-06-17
Test self hosted runners in BSI

j3parker:
do not merge!

### DIFF
--- a/.github/workflows/is-bsi-cursed.yml
+++ b/.github/workflows/is-bsi-cursed.yml
@@ -1,0 +1,7 @@
+name: Are we cursed?
+on: push
+jobs:
+  cursedornot:
+    runs-on: self-hosted
+    steps:
+      - run: echo 'not cursed'


.github/workflows/is-bsi-cursed.yml
```diff
@@ -0,0 +1,7 @@
+name: Are we cursed?
+on: push
```
j3parker, line 2: Your workflow uses a kind of event that doesn't run in PRs... I don't think that's related (i hope not!!!) but I wanted something to play with

j3parker:
wtf

j3parker:
i guess "yay for reproduction"

j3parker:
Oh, it might be because the repo is public?

j3parker:
https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories

image

Ahhh I might have even configured that at one point.

The error message isn't wrong, but it could be more helpful!

j3parker:
Not cursed, just burdened with the public spotlight.

PR #2255, opened 2020-06-24
Switch to import Lit d2l-switch from core and remove Polymer d2l-switch.

dbatiste:


rollup/rollup-wc.config.js
```diff
@@ -14,9 +14,9 @@ const componentFiles = [
 	'./node_modules/@brightspace-ui/core/components/inputs/input-time.js',
 	'./node_modules/@brightspace-ui/core/components/button/button-icon.js',
 	'./node_modules/@brightspace-ui/core/components/more-less/more-less.js',
+	'./node_modules/@brightspace-ui/core/components/switch/switch.js',
```
dbatiste, line 17: Now that nothing uses the Polymer d2l-switch in the monolith, we can safely switch these imports.  The new switch will be used shortly.

PR #2264, opened 2020-07-03
Use github-script rather than peter-evans/repository-dispatch

j3parker:


.github/workflows/trigger-lms-sync.yml
```diff
@@ -38,9 +38,16 @@ jobs:
...
+        script: |
+          github.repos.createDispatchEvent({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            event_type: bsi-sync,
+            client_payload: {
+              "bsi-version": "${{ steps.verify-release.outputs.bsi-version }}",
+              "author-email": "${{ steps.verify-release.outputs.author-email }}"
+            }
+          });
```
j3parker, line 52: I noticed that this was available in actions/github-script. Sending this change because it would be good to consolidate on actions/github-script when possible (because its more flexible). If we can get this in/working I'll remove peter-evans/repository-dispatch from third-party-actions.

(Sorry for the back-and-forth, I didn't think to look for this in github-script.)

j3parker: https://octokit.github.io/rest.js/v18

PR #2269, opened 2020-07-07
updating core

dlockhart:


package-lock.json
```diff
@@ -1147,9 +1157,9 @@
       }
     },
     "@brightspace-ui/core": {
-      "version": "1.57.1",
-      "resolved": "https://registry.npmjs.org/@brightspace-ui/core/-/core-1.57.1.tgz",
-      "integrity": "sha512-LubKANNBQErUtZgsj48YVmEu/i2O1MaPba78cH/BnBamwOV+z32IAdkR0T7aApIIkIDJZr8HZT8UHdSnt5ALEA==",
+      "version": "1.57.4",
```
dlockhart, line 1160: @AllanKerr FYI error state stuff in here

package-lock.json
```diff
@@ -4409,7 +4425,7 @@
       }
     },
     "d2l-activity-alignments": {
-      "version": "github:Brightspace/d2l-activity-alignments#c9214bf12faf2a39a4dbc16ac8e1f07ddaa172d0",
+      "version": "github:Brightspace/d2l-activity-alignments#dcbe6db5374ccd29652eb36442f6756de63fdb23",
```
dlockhart: @emil-furniss Hmm... so because `d2l-activity-alignments` isn't actually creating releases anymore, this reverting backwards to the last time a release was made is going to keep happening. I'll undo this change for now, but this will keep happening until some kind of automatic release process is in place for this repo.

package-lock.json
```diff
@@ -5257,7 +5273,7 @@
       }
     },
     "d2l-sequences": {
-      "version": "github:BrightspaceHypermediaComponents/sequences#30ffcc54e917c0270e707c85690c35b57d46a956",
+      "version": "github:BrightspaceHypermediaComponents/sequences#7a0681ee21f12a9bc88bdebef3136ee6bfc5962c",
```
dlockhart, line 5276: @imcdonald-d2l FYI

imcdonald-d2l: Thanks! 😄

package-lock.json
emil-furniss: Darn, thanks for raising that. I'll manually create a release for now so that this doesn't keep reverting.

PR #2273, opened 2020-07-09
waiting for web fonts to be loaded before fading page in

dlockhart:
Corresponding PR in Brightspace:
https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/13415/overview

This change will delay the page "fade-in" until after our web fonts have loaded in browsers that support the FontFace API. It does so in a similar mechanism to how we wait for web components today. The hope is that this helps prevent a "flash of unstyled text" from occurring in some instances.

If things take too long (>2s), we assume something has gone wrong loading the fonts and we just fade-in regardless.

### DIFF
--- a/js/bsi.js
+++ b/js/bsi.js
@@ -4,6 +4,7 @@ import './performance-timings.js';
 
 window.D2L = window.D2L || {};
 
+window.D2L.FontsLoaded = webComponentsReady.FontsLoaded;
 window.D2L.WebComponentsLoaded = webComponentsReady.WebComponentsLoaded;
 window.D2L.WCRDispatched = webComponentsReady.WCRDispatched;
 window.D2L.WebComponentsReady = webComponentsReady.WebComponentsReady;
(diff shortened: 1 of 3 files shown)


dbatiste:
I know the LMS PR is merged already, but what are your thoughts on baking that logic into typography, so that it lives with the code that is responsible for loading the fonts?  Or setting `d2lFontsLoaded` from there too late?  https://github.com/BrightspaceUI/core/blob/master/components/typography/typography.js

dlockhart:
The tricky part with moving this somewhere like into `typography.js` is going to be timing. We want to tell the browser to load these fonts as early as possible, which is why it's in a snippet of JS in the `<head>` of our pages. If we move it, then it'll have to wait for `typography.js` to load, which is done by the BSI loader, which waits for all the polyfills first.

dbatiste:
I see.  Yeah, was hoping to avoid specifying the fonts in two different places.  Thanks for entertaining the thought.  😄

dlockhart:
Yeah it's a struggle for sure. If we had a more cohesive "app", I could foresee a more centralized place for things to register that they'd like to be on the page (like `IBsiLoader`) and we could therefore do a lot more useful preloading/prefetching of important resources early in the HTML output. That's not where we are today though.

PR #2276, opened 2020-07-13
Bumping activities version

astadelbauer:


astadelbauer:
@wongvincent  @spandanabaruah  FYI

PR #2286, opened 2020-07-15
Bumping activities version

astadelbauer:


astadelbauer:
Yep I'll wait, this one's CI usually takes like 20 minutes anyway ;)

PR #2287, opened 2020-07-15
Bumping CPD version to fix a regression

BraydonKains:


BraydonKains:
I'm going to update CPD version manually instead of regen the package lock to avoid a merge conflict

PR #2325, opened 2020-07-24
dnd-sortable bump

patel-kajal:
Context:
DE39798
Fixed js error thrown in Edge when attaching a new rubric
updated d2l-dnd-sortable.js

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4563,7 +4563,7 @@
       }
     },
     "d2l-dnd-sortable": {
-      "version": "github:Brightspace/dnd-sortable#38bc8fc34b0e2d8b8faf2e35514eb002fc2e2bbc",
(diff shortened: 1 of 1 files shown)


pureooze:
Consider making a new release here https://github.com/Brightspace/dnd-sortable/releases and using its hash in the lockfile instead of the commit hash 😃 (I think that repo requires them to be manually created 😢 )

pureooze:
Sorry caused some confusion there, in other repos the bot makes a new commit when it creates the release so the hash does change for those cases 😅

astadelbauer:
> Consider making a new release here https://github.com/Brightspace/dnd-sortable/releases 😃 (I think that repo requires them to be manually created 😢 )
> 
> edit: hash wont change but its a good idea to make releases

Cool, I didn't even know to do that lol

patel-kajal:
this was merged here: https://github.com/Brightspace/brightspace-integration/pull/2329

PR #2339, opened 2020-07-28
US114907 Set WebComponent lang type to ICU in the XML file

pureooze:
Added attribute for setting langtype for webcomponent lang xml, we can parse this in the LAIM tool to determine what to set the term type to.

The new XML:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<package name="WebComponents" type="Language" langtype="ICU" toolid="0" version="0.0.0.0">
  <collection name="d2l-activities\activityCard" type="Standard">
    <langTerm sortOrder="1" name="course">
      <defaultValue><![CDATA[Course]]></defaultValue>
      <description><![CDATA[]]></description>
    </langTerm>
    ...
```

### DIFF
--- a/oslo/LangPackage.js
+++ b/oslo/LangPackage.js
@@ -45,7 +45,8 @@ class LangPackage {
 
 		serializer.writeTagBegin('package');
 		serializer.writeAttribute('name', this._name);
-		serializer.writeAttribute('type', 'Language'); // Could be IcuLanguage?
+		serializer.writeAttribute('type', 'Language');
+		serializer.writeAttribute('langtype', 'ICU');
 		serializer.writeAttribute('toolid', '0');
 		serializer.writeAttribute('version', '0.0.0.0');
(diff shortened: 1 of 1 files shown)


oslo/LangPackage.js
```diff
@@ -45,7 +45,8 @@ class LangPackage {
 
 		serializer.writeTagBegin('package');
 		serializer.writeAttribute('name', this._name);
-		serializer.writeAttribute('type', 'Language'); // Could be IcuLanguage?
+		serializer.writeAttribute('type', 'Language');
+		serializer.writeAttribute('langtype', 'ICU');
```
JordanCorner, line 49: Is there any chance that this should be more term or package specific rather than blindly saying ICU? Or is ICU the guaranteed future for the forseeable future of lang terms?

pureooze: I think its fine for now, we could always add an if statement here if we need to support SML or something else.

PR #2346, opened 2020-07-29
20.20.8 DE39964 Switch to specific version of awards v1.0.15

cbelan-d2l:
Switch to older version of core version before list component updates
-- Issues described in DE39964

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4484,16 +4484,16 @@
       }
     },
     "d2l-awards-leaderboard-ui": {
-      "version": "github:Brightspace/awards-leaderboard-ui#44a049d6465540d2cedd8f88757556cff1084506",
-      "from": "github:Brightspace/awards-leaderboard-ui#semver:^1",
+      "version": "github:Brightspace/awards-leaderboard-ui#5e9f6bfc5fbaf4d790cd39b6bb5230263a5da4e6",
+      "from": "github:Brightspace/awards-leaderboard-ui#semver:1.0.15",
       "dev": true,
       "requires": {
         "@brightspace-ui-labs/accordion": "^2",
-        "@brightspace-ui/core": "^1",
+        "@brightspace-ui/core": "1.58.0",
         "@polymer/polymer": "^3",
         "@webcomponents/webcomponentsjs": "^2",
         "d2l-fetch": "github:Brightspace/d2l-fetch#semver:^2",
-        "d2l-fetch-auth": "^1.2.0",
+        "d2l-fetch-auth": "^1.5.2",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -4484,23 +4484,42 @@
...
     "d2l-awards-leaderboard-ui": {
-      "version": "github:Brightspace/awards-leaderboard-ui#44a049d6465540d2cedd8f88757556cff1084506",
-      "from": "github:Brightspace/awards-leaderboard-ui#semver:^1",
+      "version": "github:Brightspace/awards-leaderboard-ui#5e9f6bfc5fbaf4d790cd39b6bb5230263a5da4e6",
+      "from": "github:Brightspace/awards-leaderboard-ui#semver:1.0.15",
       "dev": true,
       "requires": {
         "@brightspace-ui-labs/accordion": "^2",
-        "@brightspace-ui/core": "^1",
+        "@brightspace-ui/core": "1.58.0",
```
dlockhart, line 4492: This isn't going to do what you're expecting... the core version for `20.20.8` is `1.59.x` (currently `1.59.5`) and already contains a bunch of bug fixes for 20.20.8.

dlockhart: Essentially everything that's in BSI needs to use the same version of core -- awards can't use a different version. So if 1.59 of core broke something, it either needs to be removed/fixed from 1.59 and released as 1.59.6 or awards will need to be updated to resolve the issues.

cbelan-d2l: Oh no! Wasn't sure about that. Thanks for letting me know.
Yes, there's a bug in the list component that caused the leaderboard to break.
But @m6jones thinks that @zinabat may have a fix for it... ?

If you expand one of the accordion arrows, then another one by clicking the arrow, you receive this error:
```
focus.js:1 Uncaught TypeError: Cannot read property 'length' of null
    at u (focus.js:1)
    at Object.handleEvent (list.js:1)
    at O.handleEvent (lit-element.js:1)
    at HTMLSlotElement.O.__boundHandleEvent (lit-element.js:1)
list.js:1 Uncaught TypeError: Cannot read property 'getAttribute' of undefined
    at HTMLElement._setFocusInfo (list.js:1)
```

Another issue was the color of the row didn't highlight for the currently logged in user -- this I can deal with not making it into cert. But the ability to click multiple accordions is the issue I was more concerned with and wanted to switch back to older core for.

zinabat: This is a one-line fix in core, so I'll do that now.

cbelan-d2l:
Closing this off as it won't work!
Zina has the fix to list component

PR #2371, opened 2020-08-04
removing the widget demarcation class

dlockhart:
This feature flag has been removed and the CSS class is [always applied](https://search.d2l.dev/xref/lms/platformTools/homepage/definition/Web/Desktop/MasterPages/Default/HomepagePageFactory.cs?r=f326ca6a#76), so to simplify I'd like to just remove it.

### DIFF
--- a/sass/homepages/homepages.scss
+++ b/sass/homepages/homepages.scss
@@ -59,62 +59,59 @@ $viewport-width-sm: 768px;
 	}
 }
 
-.daylight-widget-demarcation {
+.d2l-homepage-pageheader {
+	margin: 0;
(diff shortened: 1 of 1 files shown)


sass/homepages/homepages.scss
```diff
@@ -59,62 +59,59 @@ $viewport-width-sm: 768px;
 	}
 }
 
-.daylight-widget-demarcation {
```
dlockhart, line 62: Everything here stayed the same, I just removed `daylight-widget-demarcation` and out-dented everything that was nested under it.

PR #2375, opened 2020-08-05
removing polymer behaviours expand-collapse

dlockhart:
This saves 2 requests and 2138 bytes from every page.


package.json
```diff
@@ -69,7 +69,6 @@
     "d2l-organizations": "BrightspaceHypermediaComponents/organizations#semver:^5",
     "d2l-outcomes-level-of-achievements": "Brightspace/outcomes-level-of-achievement-ui.git#semver:^2",
     "d2l-outcomes-user-progress": "Brightspace/d2l-outcomes-user-progress.git#semver:^1",
-    "d2l-polymer-behaviors": "Brightspace/d2l-polymer-behaviors-ui#semver:^2",
```
dlockhart, line 72: This doesn't get it out of our package-lock as there are still a lot of Polymer components referencing it.

PR #2390, opened 2020-08-11
expose a file on BSI that imports the intl-messageformat-parser

patel-kajal:
[US114902](https://rally1.rallydev.com/#/detail/userstory/378217884416)

Expose a file in BSI that imports the intl-messageformat-parser at runtime and then installs it on a global so that we can reference is in the monolith

We want to use this library to validate ICU Message input fields on the Manage Language page

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1209,9 +1209,9 @@
       }
     },
     "@brightspace-ui/core": {
-      "version": "1.70.0",
-      "resolved": "https://registry.npmjs.org/@brightspace-ui/core/-/core-1.70.0.tgz",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -4506,7 +4506,7 @@
       }
     },
     "d2l-consistent-evaluation": {
-      "version": "github:BrightspaceHypermediaComponents/consistent-evaluation#f88bd7d090be022a7f0d1945002601c2def0b898",
+      "version": "github:BrightspaceHypermediaComponents/consistent-evaluation#d743478c657ba95edb718580a5deddabe11b7106",
```
dlockhart, line 4509: @JamieTran @@jmurray-d2l FYI

PR #2407, opened 2020-08-24
Add Capture Central app to BSI

AdmiralCoco:


web-components/d2l-capture-central.js
```diff
@@ -0,0 +1 @@
+import 'd2l-capture-central/d2l-capture-central.js';
```
dlockhart, line 1: FYI, you don't need to create these files in `web-components` anymore -- you could just reference them directly from the `rollup-wc.config.js`.

PR #2420, opened 2020-08-27
Add client-side error logging client

AllanKerr:
# Changes
- Added a client-side error logging client.
    - **Batching:** This client will batch messages and send them every time the batch reaches the max size of 500 or a 5 second period with no errors has occurred.
    - **Unload**: When navigating away from a page any error messages that have been batched but not sent will be sent using the `navigator.sendBeacon` API in unload.
    - **Endpoint Provisioning:** The logging endpoint is cached in the browser HTTP cache and will only be refreshed if a `410` status code is received or the cache entry has expired.

### DIFF
--- a/web-components/bsi-unbundled.js
+++ b/web-components/bsi-unbundled.js
@@ -77,6 +77,57 @@ window.D2L.Telemetry = {
 	}
 };
 
+window.D2L.Logging.CreateClient = function() {
+	const batchSize = 500;
+	const batchTime = 5000;
+	const client = {
+		Log: function(messages) {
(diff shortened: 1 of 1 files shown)


web-components/bsi-unbundled.js
```diff
@@ -77,6 +77,57 @@ window.D2L.Telemetry = {
 	}
 };
 
+window.D2L.Logging.CreateClient = function() {
```
AllanKerr, line 80: We may want to move this out of `bsi-unbundled` and into its own file that gets imported.

PR #2421, opened 2020-08-27
bumped version of brightspace-ui/core to 1.72.7

adkulas:
bumped version of `@brightspace-ui/core`


astadelbauer:
Oops, sorry, got you to update to master's version instead of cert's 😅

PR #2423, opened 2020-08-27
Fix missing d2l-status-indicator in the LMS

svanherk:
The LMS is using the `d2l-status-indicator` but we don't explicitly include it.  My guess was that it used to be included "for free" with something else in the bundle, and that got moved out.  We'll want to unbundle this and have the LMS load the module, but just trying to fix the problem for now - there's probably some cleanup that can happen in the LMS to get the legacy component using the MVC one.

### DIFF
--- a/web-components/bsi-unbundled.js
+++ b/web-components/bsi-unbundled.js
@@ -40,6 +40,8 @@ import '@brightspace-ui/core/components/menu/menu-item.js';
 // EditNavbar, MobileLinksArea, HomepageManageMenu, PageActionsMenu, Menu (MVC), ContextMenu (MVC),
 // MediaPlayer, contextMenu (legacy), ActionButtons (legacy)
 import '@brightspace-ui/core/components/menu/menu.js';
+// StatusIndicator (both legacy and MVC)
+import '@brightspace-ui/core/components/status-indicator/status-indicator.js';
 
 window.D2L = window.D2L || {};
 


web-components/bsi-unbundled.js
```diff
@@ -40,6 +40,8 @@ import '@brightspace-ui/core/components/menu/menu-item.js';
 // EditNavbar, MobileLinksArea, HomepageManageMenu, PageActionsMenu, Menu (MVC), ContextMenu (MVC),
 // MediaPlayer, contextMenu (legacy), ActionButtons (legacy)
 import '@brightspace-ui/core/components/menu/menu.js';
+// StatusIndicator (both legacy and MVC)
```
dbatiste, line 43: Are you sure this shouldn't have gone into [rollup-wc.config](https://github.com/Brightspace/brightspace-integration/blob/master/rollup/rollup-wc.config.js)?

svanherk: It should eventually, but I _believe_ that's where it goes after it's been properly unbundled in the LMS (`codegen` file created, `BSILoader` called, etc).  This is putting it in the "to unbundle to do list", to keep the fix small and fast.

dlockhart: Thanks for finding/fixing this. I can handle the "proper" unbundling this week.

svanherk: Great thanks!  From a quick look it seemed legacy and MVC were basically copies of each other, so there might be an easy amount of cleanup there to reduce code duplication

PR #2439, opened 2020-09-03
Update leaderboard version

sgrinwis-d2l:
This is a fix for DE40387. 

We have a new version of leaderboard that drops the unneeded leftover FRA auth bits


svanherk (approved):
Nice!

PR #2441, opened 2020-09-03
delete collections that are not activity from serge-mapping

patel-kajal:
[DE40300](https://rally1.rallydev.com/#/110294864140d/custom/110851118712?detail=%2Fdefect%2F421999746960)

[PR](https://github.com/Brightspace/lms/pull/1009) where the terms for these collections are deleted in the database
[PR](https://github.com/Brightspace/lms/pull/990) to delete old activity lang terms from the database


JordanCorner (approved):
This looks good. Make sure this is merged and the BSI change is in the lms before merging the migration script so we don't recreate them immediately after deleting them.

patel-kajal:
bsi 20.20.10-36

PR #2480, opened 2020-09-15
Initial BSI update workflow

svanherk:
Uses https://github.com/Brightspace/lms/blob/master/.github/workflows/bsi-sync.yml as inspiration.  Still need to setup secrets, and for now the job runs on demand, not on a schedule.  I'm also not confident in my syntax, but I believe I need to merge this to master before I can test it out.

### DIFF
--- a/.github/workflows/bsi-update.yml
+++ b/.github/workflows/bsi-update.yml
@@ -6,7 +6,7 @@ on: workflow_dispatch
 
 jobs:
   build:
-    # Must use this because we are public
+    # self-hosted doesn't work for public repos (security), so need to use ubuntu
     runs-on: ubuntu-latest
 
     steps:
@@ -23,7 +23,9 @@ jobs:
       
       - name: Check For Changes
         id: check-changes
-        run: echo "::set-output name=changes-exist::$(git diff --exit-code package-lock.json)
+        run: |
+          echo "::set-output name=changes-exist::$(git diff --exit-code package-lock.json)"
+          echo "::set-output name=datetime::$(date +'%Y-%m-%d.%H')"
 
       - name: Create PR
         id: create-pr
@@ -31,11 +33,12 @@ jobs:
         uses: Brightspace/third-party-actions@actions/github-script
         with:
           script: |
+            const dateTime = 
             const createPrResult = await github.pulls.create({
               owner: 'Brightspace',
               repo: 'brightspace-integration',
-              title: 'Updating BSI Dependencies - ${{ date +'%Y-%m-%d %H:%M:%S' }}',
-              head: 'refs/heads/ghworkflow/bsi_update_${{ date +'%Y-%m-%d-%H-%M-%S' }}',
+              title: 'Updating BSI Dependencies - ${{ steps.check-changes.outputs.datetime }}',
+              head: 'refs/heads/ghworkflow/bsi_update_${{ steps.check-changes.outputs.datetime }}',
               base: 'refs/heads/master'
             });
             github.issues.addLabels({
--- a/.github/workflows/bsi-update.yml
+++ b/.github/workflows/bsi-update.yml
@@ -1,4 +1,5 @@
-# This workflow runs "npm i" to see if any dependencies have updated.  If so, it opens a PR and approves it.
+# This workflow runs "npm i" to see if any BSI dependencies have updated.
+# If so, it commits the package-lock changes and opens a PR if one does not already exist.
 
 name: BSI Update
 
@@ -10,6 +11,11 @@ jobs:
     # self-hosted doesn't work for public repos (security), so need to use ubuntu
     runs-on: ubuntu-latest
 
+    timeout-minutes: 20
+
+    env:
+      BRANCH_NAME: ghworkflow/bsi_auto_update
+
     steps:
       - name: Checkout Repository
         uses: Brightspace/third-party-actions@actions/checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/bsi-update.yml
```diff
@@ -0,0 +1,68 @@
...
+          node-version: '12.x'
+
+      - name: Fresh Dependency Install
+        run: npm install
+      
+      - name: Check For Changes
+        id: check-changes
+        run: |
+          echo "::set-output name=changes-exist::$(git diff --exit-code package-lock.json)"
+          echo "::set-output name=datetime::$(date +'%Y-%m-%d.%H')"
```
svanherk, line 28: I assume this will be in UTC, and I haven't found an easy way to convert this other than using some third-party github actions. We mostly need these to be different, not accurate, but I could see the offset being a bit annoying

.github/workflows/bsi-update.yml
```diff
@@ -0,0 +1,68 @@
...
+              repo: 'brightspace-integration',
+              title: 'Updating BSI Dependencies - ${{ steps.check-changes.outputs.datetime }}',
+              head: 'refs/heads/ghworkflow/bsi_update_${{ steps.check-changes.outputs.datetime }}',
+              base: 'refs/heads/master'
+            });
+            github.issues.addLabels({
+              owner: 'Brightspace',
+              repo: 'brightspace-integration',
+              issue_number: createPrResult.data.number,
+              labels: ['auto-merge-bsi']
```
wongvincent, line 48: just fyi, the `auto-merge bsi` workflow in the LMS repo looks for shit `auto-merge-bsi` label. Are you planning to add a similar CRON job to the BSI repo?

svanherk: Yeah that's what I was thinking, which would be almost identical to the one in the LMS.  I can change the tag name to avoid confusion, it just matched exactly what I wanted it to do lol

wongvincent (approved):

.github/workflows/bsi-update.yml
wongvincent: yeah, would be good to use a tag name that fits what you're doing

wongvincent (approved):

.github/workflows/bsi-update.yml
```diff
@@ -0,0 +1,69 @@
...
+  workflow_dispatch: # manual trigger for testing
+
+jobs:
+  build:
+    # self-hosted doesn't work for public repos (security), so need to use ubuntu
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
```
j3parker, line 15:
```suggestion
        uses: Brightspace/third-party-actions@actions/checkout
```

This isn't enforced yet but will be "soon". [Check the docs for the rationale](https://github.com/Brightspace/third-party-actions#why).

.github/workflows/bsi-update.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Use Node.js
+        uses: actions/setup-node@v1
```
j3parker, line 18:
```suggestion
        uses: Brightspace/third-party-actions@actions/setup-node
```

.github/workflows/bsi-update.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          node-version: '12.x'
+
+      - name: Fresh Dependency Install
+        run: npm install
+      
+      - name: Check For Changes
+        id: check-changes
+        run: |
+          echo "::set-output name=changes-exist::$(git diff --exit-code package-lock.json)"
```
j3parker, line 28:
```suggestion
          echo "::set-output name=changes-exist::$(git diff -s --exit-code package-lock.json && echo false || echo true)"
```

I think? Like `--exit-code` alone outputs the diff and sets the exit code to 1 if its non-empty... `-s` makes it not output the diff, but to get something usable I think you need to do the `&&`/`||` stuff.

If 0/1 work here (I don't remember) then this would work too:

```suggestion
          echo "::set-output name=changes-exist::$(git diff -s --exit-code package-lock.json ; echo $?)"
```

.github/workflows/bsi-update.yml
j3parker: What I've found works the best is actually the opposite approach...
Like maybe always use a consistent branch name and:

* checkout the branch if it exists, otherwise create it and check it out (so maybe `git checkout foo || git checkout -b foo`)
* do the update step (`npm install`)
* if there is a diff, `git reset --hard origin/master` and do `npm install` again
* make a commit
* _force_ push
* create the PR if it doesn't exist

That way if the PR sits around stale (e.g. CI was failing) and it runs again you'll just be changing the already open PR instead of doing a new one.

IIUC you're hoping to auto-merge these, and maybe CI failing long enough for this to matter is unlikely. It's up to you.

(Aside: this is why dependabot is nice -- it does an even more thorough job of handling this stuff.)

j3parker: Don't take this as gospel, like decide what makes sense for you.

.github/workflows/bsi-update.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Approve PR
+        id: approve-pr
+        if: steps.check-changes.outputs.changes-exist
+        env:
+          PULL_ID: ${{ steps.create-pr.outputs.result }}
+        uses: Brightspace/third-party-actions@actions/github-script
+        with:
+          script: |
+            github.pulls.createReview({
```
j3parker, line 63: The review will be created by the shared "github-actions" user, and a user can't approve their own PRs.

You could add another users token to this repo as a secret to approve the PR, but be careful because this is a public repo and we don't want someone to be able to fork and PR and steal that token. By default forks don't run actions/get secrets, but I'm a bit nervous because people like to change these settings (I think @khorwood 's team is interested in monitoring for stuff like that though.)

Maybe just add a human to the review to start?

j3parker: Oh I see, you have added a secret for this. OK.

j3parker (approved):

.github/workflows/bsi-update.yml
svanherk: Makes total sense

.github/workflows/bsi-update.yml
svanherk: Yeah, this part was a guess until I could actually test things out.  I'll try one of these and see how it goes

.github/workflows/bsi-update.yml
dlockhart: I like the idea of always reusing the same branch name and just constantly updating it or re-creating it.

dlockhart (approved):

.github/workflows/bsi-update.yml
svanherk: Yeah me too. will update do that

wongvincent: That idea would work well for the BSI bumps as well.. right now, if there are 2 bumps back to back, one of the PRs will fail to merge due to a merge conflict. hmmm.

PR #2488, opened 2020-09-17
Auto BSI - Set upstream

svanherk:
Getting closer...

### DIFF
--- a/.github/workflows/bsi-update.yml
+++ b/.github/workflows/bsi-update.yml
@@ -44,7 +44,7 @@ jobs:
           git config user.email github-actions@github.com
           git add package-lock.json
           git commit -m 'BSI Auto Update Dependencies'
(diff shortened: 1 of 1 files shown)


.github/workflows/bsi-update.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           git config user.email github-actions@github.com
           git add package-lock.json
           git commit -m 'BSI Auto Update Dependencies'
-          git push --force
+          git push --force --set-upstream origin $BRANCH_NAME
```
j3parker, line 47:
```suggestion
          git push --force origin $BRANCH_NAME
```

`--set-upstream` is fine, it does the same as this but also configures the branch to automatically track `origin` for "next time", but in these builds there is no next time (like next time will run on a fresh VM) so its a bit simpler to just specify it as a one-off

j3parker (approved):

PR #2490, opened 2020-09-17
US119926 Add ads scheduler

JeremyWinkler:


package.json
```diff
@@ -84,6 +84,7 @@
     "d2l-table": "BrightspaceUI/table#semver:^2",
     "d2l-teacher-course-creation": "Brightspace/custom-teacher-course-creation#semver:^1",
     "d2l-users": "BrightspaceHypermediaComponents/users#semver:^2",
+    "d2l-ads-scheduler": "Brightspace/custom-ads-scheduler#semver:^1",
```
lweber-d2l, line 87: From https://semver.org: `Major version zero (0.y.z) is for initial development.`
Should we be using 0.y.z until we are closer to release?

JeremyWinkler: I already made a 1.0.0 tag in the scheduler repo. Not sure if decrementing it would cause issues. For future projects, I'll make them start at 0.

lweber-d2l: Sounds good - thanks.

lweber-d2l (approved):

PR #2493, opened 2020-09-17
use @brightspace-ui/core 1.79.4

AKobets:


AKobets:
actually there is already another PR that inlcludes core bump https://github.com/Brightspace/brightspace-integration/pull/2492

dlockhart:
I'd ignore #2492 for now as I think that's just @svanherk experimenting with automating all this :)

svanherk:
Yeah that won't be merged for a little bit so I'd go ahead with this one, but it's coming along! 🤞

AKobets:
BSI updated with ui core release 1.80.0

PR #2495, opened 2020-09-17
Update oslo readme

pureooze:
Updating the docs to use markdown and to mention the issue with the special characters. I want to link to them from the OSLO controller so people have more context on why we are going to parse `@` when retrieving from the DB in the code I added (in a different PR in the lms).

To view the markdown: https://github.com/Brightspace/brightspace-integration/blob/ushamim/usa/update-oslo-readme/oslo/README.md

### DIFF
--- a/oslo/README.md
+++ b/oslo/README.md
@@ -17,6 +17,8 @@ Produces LMS definition/translation XMLs and an Oslo manifest file for the Serge
 configuration files specified in ".serge-mapping.json". Should be run every
 build, as output is synced to the monolith with the BSI version bump.
 
+Note: The LMS `LANG_OBJECTS` table is not able to store certain special characters like `@`. It will do a conversion to a unicode representation like `\u0040`. For this reason if there is a special character in the package/collection/term name support needs to be added to the OSLO controller in the LMS to parse that character.
+
 **Output:**
 
 	build/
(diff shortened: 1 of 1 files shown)


oslo/README.md
```diff
@@ -1,26 +1,25 @@
...
-generate-monolith-xml
----------------------
+## generate-monolith-xml
 
 Produces LMS definition/translation XMLs and an Oslo manifest file for the Serge
 configuration files specified in ".serge-mapping.json". Should be run every
 build, as output is synced to the monolith with the BSI version bump.
 
-Output:
+Note: The LMS `LANG_OBJECTS` table is not able to store certain special characters like `@`. It will do a conversion to a unicode representation like `\u0040`. For this reason if there is a special character in the package/collection/term name support needs to be added to the OSLO controller in the LMS to parse that character.
```
astadelbauer, line 20: @kieranderson  random thought... do we actually need to be encoding anything other than periods?  Seems weird now that we change @ ?  Possibly Patrick quickly wrote an "encode all the things" function not knowing it would affect things down the line?

astadelbauer (approved):

oslo/README.md
kieranderson: @astadelbauer Possible. Initial intention I believe was to be able to use whatever setup is being used in each of the components. What is actually used by each is probably a small scope of chars.

patel-kajal (approved):

PR #2496, opened 2020-09-17
Add class for disabling scroll on the page

ianlaidlaw:
When using LX, design has decided that we don't want to have "disabled" scroll bars on the side of the page. However, when using d2l-body, which we need to do for other behavior and style reasons, `overflow-y: scroll` is added. This pr just adds another class which we can append in our controller when creating LX to unset that property.

Followup pr to that controller will come later once/if this gets merged.

scroll

### DIFF
--- a/sass/page.scss
+++ b/sass/page.scss
@@ -77,3 +77,7 @@ body.d2l-body-unresolved {
 	position: fixed;
 	top: 10px;
 }
+
+.d2l-disable-overflow {
+	overflow-y: unset;
+}


dlockhart:
Could you not set this using JavaScript from your app instead? i.e. `document.body.style.overflowY = 'unset';`

dlockhart:
Also it seems weird to just apply this to LX and not make a decision system-wide.

ianlaidlaw:
> Could you not set this using JavaScript from your app instead? i.e. document.body.style.overflowY = 'unset';

I can give that a try. That does feel a little gross to me, reaching out of LX to remove a style like that - but it would probably work. With this CSS class then at least the behavior is documented.

> Also it seems weird to just apply this to LX and not make a decision system-wide.

Agreed, that's probably a conversation for the designers to have. But I guess since LX is designed to be an "immersive" experience they want to remove all these extraneous elements.

pbrown-d2l:
> > Could you not set this using JavaScript from your app instead? i.e. document.body.style.overflowY = 'unset';
> 
> I can give that a try. That does feel a little gross to me, reaching out of LX to remove a style like that - but it would probably work. With this CSS class then at least the behavior is documented.

Apparently, this is how Dialogs do it. When they open, they disable scrolling. We can probably do the same from within LX

dlockhart:
The dialog code calls into the backdrop stuff here:
https://github.com/BrightspaceUI/core/blob/master/components/backdrop/backdrop.js#L173

ianlaidlaw:
> > > Could you not set this using JavaScript from your app instead? i.e. document.body.style.overflowY = 'unset';
> > 
> > 
> > I can give that a try. That does feel a little gross to me, reaching out of LX to remove a style like that - but it would probably work. With this CSS class then at least the behavior is documented.
> 
> Apparently, this is how Dialogs do it. When they open, they disable scrolling. We can probably do the same from within LX

Sure, I'll do that for now. I wonder if since it looks like we have multiple places where we don't want the scroll bar we can add some solution here on `page.scss` so we don't have to manually override from within the app or dialog. A non-scrolling full screen view might be getting common enough at this point?

dlockhart:
Agreed. Like if you're going for a truly "immersive" experience you'll want more than just disabling scrolling. If you look at the backdrop code it has stuff that makes the document "behind" hidden to screen readers, which is likely something you want as well.

Ideally this definitely isn't something you need to do manually, and would just come for free with whatever component(s) you're using to get the experience in the first place -- be it a dialog, full-screen dialog or other kind of overlay.

PR #2497, opened 2020-09-17
Remove -s from the git diff command

svanherk:
It seems to think there is a diff every time even when nothing has changed since the last run on that branch, so I want to see what it sees.  I think this will be helpful for any future debugging, so probably going to keep it this way!

### DIFF
--- a/.github/workflows/bsi-update.yml
+++ b/.github/workflows/bsi-update.yml
@@ -36,7 +36,7 @@ jobs:
       
       - name: Check For Changes
         id: check-changes
-        run: echo "::set-output name=changes-exist::$(git diff -s --exit-code package-lock.json && echo false || echo true)"
+        run: echo "::set-output name=changes-exist::$(git diff --exit-code package-lock.json && echo false || echo true)"
 
       - name: Create Commit
         if: steps.check-changes.outputs.changes-exist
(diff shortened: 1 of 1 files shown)


.github/workflows/bsi-update.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       
       - name: Check For Changes
         id: check-changes
-        run: echo "::set-output name=changes-exist::$(git diff -s --exit-code package-lock.json && echo false || echo true)"
+        run: echo "::set-output name=changes-exist::$(git diff --exit-code package-lock.json && echo false || echo true)"
```
j3parker, line 39: Ok, so the trouble with this is that `$(...)` takes the standard output from the subsell and barfs it into this echo command.

Here's something from another repo: https://github.com/Brightspace/github-actions-infrastructure/blob/master/.github/workflows/update-base-amis.yml#L23

```
git diff --exit-code || echo "::set-output name=has_changes::true"
```

(you don't need `--ignore-submodules` in this repo)

It looks like variables are default falsey if they haven't been set.

j3parker: The line above doesn't have `-s` but it's ok because we're not using it inside `$( ... )`

svanherk: Perfect, thank you!  This looks a lot better and makes more sense

PR #2537, opened 2020-09-30
US121057 OSLO > Fail BSI build when there are invalid term names

pureooze:
This implements [US121057](https://rally1.rallydev.com/#/388869855380d/dashboard?detail=%2Fuserstory%2F439112983600&fdp=true) and causes the BSI build to fail if there are any characters in a term name that contain a forbidden character (for now that is only `.` but we may add more later).

### DIFF
--- a/oslo/Serge.js
+++ b/oslo/Serge.js
@@ -20,6 +20,8 @@ const ParseJsNameRegExp = /["']/g;
 const ParseJsEscapeRegExp = /\\(['"])/g;
 const ParseJsEscapeReplacer = (_, ch) => ch;
 
+let invalidTermFound = false;
+
 class Rewrite {
 
 	constructor(rewrite) {
@@ -158,20 +160,27 @@ class Serge {
 
 	_parse(text) {
 
+		let result;
 		switch (this._parserPlugin) {
 
 			case 'parse_json':
-				return this._parseJson(text);
-
+				result = this._parseJson(text);
+				break;
 			case 'parse_d2l_fra':
-				return this._parseFra(text);
-
+				result = this._parseFra(text);
+				break;
 			case 'parse_js':
-				return this._parseJs(text);
-
+				result = this._parseJs(text);
+				break;
 			default:
 				throw new Error(`Unsupported parser plugin for "${this._name}" ("${this._parserPlugin}")`);
 		}
+
+		if (invalidTermFound) {
+			throw 'OSLO error: Forbidden characters used in LangObject name';
+		}
+
+		return result;
 	}
 
 	_parseJson(text) {
@@ -179,7 +188,6 @@ class Serge {
 		const source = JSON.parse(text);
 		const entries = Object.entries(source);
 		const collection = new LangCollection(this._name);
-		let invalidTermFound = false;
 
 		for (const [name, defaultValue] of entries) {
 
@@ -193,10 +201,6 @@ class Serge {
 			}
 		}
 
-		if (invalidTermFound) {
-			throw 'OSLO error: Forbidden characters used in LangObject name';
-		}
-
 		return collection;
 	}
 
@@ -205,7 +209,6 @@ class Serge {
 		const source = JSON.parse(text);
 		const entries = Object.entries(source);
 		const collection = new LangCollection(this._name);
-		let invalidTermFound = false;
 
 		for (const entry of entries) {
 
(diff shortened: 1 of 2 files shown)


pureooze:
Well it definitely works lol

oslo/Serge.js
```diff
@@ -179,11 +179,22 @@ class Serge {
...
 
-			const object = new LangObject(name, defaultValue, EmptyString);
-			collection.addObject(object);
+			const validName = Util.validLangObjectName(name, this._name);
+
+			if (validName) {
+				const object = new LangObject(name, defaultValue, EmptyString);
+				collection.addObject(object);
+			} else {
+				invalidTermFound = true;
```
kieranderson, line 192: Can exit once we have at least one invalid term

oslo/Util.js
```diff
@@ -41,8 +43,20 @@ function backSlash(value) {
 	return value.replace(ForwardSlashRegExp, BackSlashChar);
 }
 
+function validLangObjectName(objectName, collectionName) {
+	const nameContainsForbiddenCharacter = ForbiddenCharacters.some(char => objectName.includes(char));
+
+	if (nameContainsForbiddenCharacter && !ForbiddenCharacterAllowedList.includes(collectionName)) {
```
kieranderson, line 49: What is the performance of adding this extra check?

See here for comparison times (In the comments)
https://github.com/Brightspace/brightspace-integration/pull/2011

oslo/Util.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+const ForbiddenCharacterAllowedList = ['d2l-activities\\activityEditor'];
```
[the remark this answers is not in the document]
pureooze: Oh I see the story only mentions activities under the "Need an exemption" category but attackments is listed under acceptance criteria. Will add!

oslo/Serge.js
pureooze: Oh we dont want to print them all out? Thought it might be more helpful to call them all out before failing.

pureooze: I was thinking it would be similar to how like eslint displays all the lint errors before failing.

oslo/Util.js
pureooze: **branch**
```sh
$ time npm run build:lang

> brightspace-integration@ build:lang C:\D2L\offstack\brightspace-integration
> node ./oslo/generate-monolith-xml


real    0m1.137s
user    0m0.030s
sys     0m0.182s
```

**master**
```sh
$ time npm run build:lang

> brightspace-integration@ build:lang C:\D2L\offstack\brightspace-integration
> node ./oslo/generate-monolith-xml


real    0m1.068s
user    0m0.075s
sys     0m0.153s

```

oslo/Util.js
```diff
@@ -1,5 +1,10 @@
 'use strict';
 
+const ForbiddenCharacterAllowedList = [
+	'd2l-activities\\activityEditor',
+	'@d2l\\d2l-attachment\\d2l-attachment'
+];
+const ForbiddenCharacters = ['.'];
 const EscapeRegExp = /([^a-zA-Z0-9\\:_-])/g;
```
adkulas, line 8:
```suggestion
const EscapeRegExp = /([^a-zA-Z0-9\\:_-@])/g;
```
Do we want to add `@` to this?

oslo/Util.js
adkulas: is this just one run?

pureooze: Although I suppose I could make a regex if I really had to....

adkulas: The reason I bring it up is because we already have one and are using it for the parse right? Is the goal here just to cause `.` to error?  Or to basically not have to parse anything and have all characters be valid?

(I will read the ticket lol)

adkulas: Yeah your approach makes sense given the context.

So overall we will have:
- a set of characters that are totally valid
- a set of characters that are valid but will be parsed to unicode
- a set of characters that are illegal and will throw an error (aka only `.` right now)

kieranderson: Keep in mind Patrick's original numbers were against a larger set of terms, since generate-serge-mapping was run.

But we can cross this perf bridge if/when we get to it.

pureooze: Oh thats true actually, I didnt think about that. But I ran it against master too so 🤷

Lgodden (approved):

kieranderson (approved):

PR #2540, opened 2020-10-02
Added tests for serge parser to assert forbidden character cases

pureooze:
Follow up from https://github.com/Brightspace/brightspace-integration/pull/2537 to add tests


pureooze:
Code owners works 😃

test/Serge.test.js
```diff
@@ -11,26 +12,24 @@ describe('Serge parsing', () => {
...
-
-		it('parses complicated js', () => {
-
-			const collection = serge._parseJs(js);
+		const termNames = ['txtTerm\\u0020One', 'txtTermTwo'];
+
+		it('parses valid js', () => {
+			const js = `
+				import something from 'something-else';
+
```
adkulas, line 20: First time looking at these tests, I don't know if its just me, but these tests are very hard to read 🤷 

Just one observation: why is this [test](https://github.com/Brightspace/brightspace-integration/blob/7cae644d9418f14fc74626716ab44ba5947c4bf7/test/Serge.test.js#L13) indented?

pureooze: What do you find hard to read? The syntax, they way they are written or something else?

`describe` blocks split the output from mocha into indented categories so you can group test results together using them. So here its creating a section for the `js` parser and I will add separate sections for the `json` and `fra` parsers.

adkulas: So is the described on line 13 "nested" within the describe on line 5?

pureooze: Yeah so in the output its like this: https://travis-ci.com/github/Brightspace/brightspace-integration/builds/187871044#L333-L349

adkulas: ah shoot I didn't see that the `}` was closing the object not the describe. 
so it looked like each describe was separate but somehow the only thing that was nesting it was the indent. Makes sense now

pureooze: Yeah I have always hated that code diffs dont let you interact like that (and they dont syntax highlight)

adkulas: as for the readability thing. Can we run this through a code formatter?

pureooze: They already run through eslint which forces certain formatting rules. What parts are you finding have bad readability?

adkulas: had to create PR into your branch #2543

test/Serge.test.js
```diff
@@ -11,26 +12,24 @@ describe('Serge parsing', () => {
...
+			const js = `
+				import something from 'something-else';
+				
+				// comment to be noisy
+				
+				const someConstant = 42;
+				
+				export default {
+					'txtTerm One'  :"Value \\"'\\''\\" 1"  ,    // txtTermOne description
+					txtTermTwo  :  '"\\'\\"Value: 2\\":,,\\'"'//txtTermTwo description
```
patel-kajal, line 27: does `txtTermTwo` need brackets around it?

test/Serge.test.js
```diff
@@ -40,22 +39,40 @@ txtTermThree  :  '"\\'\\"Value: 3\\":,,\\'"'//txtTermThree description
...
+				
+				const someConstant = 42;
+				
+				export default {
+					'termWith${char}InIt' : "Value"
+				};
+			`;
+
+				serge._parserPlugin = 'parse_js';
+				expect(() => serge._parse(js)).to.throw('OSLO error: Forbidden characters used in LangObject name');
```
patel-kajal, line 74: should we just add a comment here for which characters are forbidden?

test/Serge.test.js
```diff
@@ -40,22 +39,40 @@ txtTermThree  :  '"\\'\\"Value: 3\\":,,\\'"'//txtTermThree description
...
+				import something from 'something-else';
+				
+				// comment to be noisy
+				
+				const someConstant = 42;
+				
+				export default {
+					'termWith${char}InIt' : "Value"
+				};
+			`;
```
patel-kajal, line 71: nit: could indent this string

pureooze, line 71:
```suggestion
				const js = `
					import something from 'something-else';
				
					// comment to be noisy
				
					const someConstant = 42;
				
					export default {
						'termWith${char}InIt' : "Value"
					};
				`;
```

test/Serge.test.js
patel-kajal: Doesn't really matter though, if someone debugging they can just go look into the Util.js file

pureooze: Hmm good point, right now we dont actually log what the forbidden characters are.
[Here is where we log which terms had an issue](https://github.com/Brightspace/brightspace-integration/blob/master/oslo/Util.js#L49-L58)
[And here is where the exception is thrown](https://github.com/Brightspace/brightspace-integration/blob/master/oslo/Serge.js#L179-L181)

I think it would be good to add the characters to the exception, at least then people have more information about what could be wrong 🤔

patel-kajal: Is it possible to log it here when the test fails, since we know which character was passed in

patel-kajal: Like if it doesn't throw throw the expected exception, could we add a message to the failed test to just point out which forbidden character wasn't caught

patel-kajal (approved):
LGTM 👍  , just minor comments

test/Serge.test.js
pureooze: Yeah that should be possible and I think it would help when debugging the tests. Kinda hard to tell which specific one failed without it unless you add console.log I guess but thats not nice. I'll look into adding that to the message, it shouldnt be too hard. 🤔

patel-kajal: Yeah that would be good, I think right now there is only one marked forbidden ('.') so it might not be a big deal, but it would be nice to have

test/Serge.test.js
```diff
@@ -40,22 +39,43 @@ txtTermThree  :  '"\\'\\"Value: 3\\":,,\\'"'//txtTermThree description
...
+				
+					export default {
+						'termWith${char}InIt' : "Value"
+					};
+				`;
+
+				serge._parserPlugin = 'parse_js';
+				expect(
+					() => serge._parse(js),
+					`Expected OSLO error when parsing character '${char}'`
```
pureooze, line 76: @patel-kajal how about this?

JordanCorner (approved):

test/Serge.test.js
patel-kajal: This looks good!

PR #2541, opened 2020-10-02
Merge OSLO docs into the OSLO directory in BSI

pureooze:
This PR preserves the commit history from [the docs repo](https://github.com/Brightspace/oslo-docs) (which we can delete after this merges)


adkulas (approved):
We should clean up the commit history if this will end up in master

PR #2543, opened 2020-10-02
Update Serge.test.js

adkulas:
Took this from https://beautifier.io/

### DIFF
--- a/test/Serge.test.js
+++ b/test/Serge.test.js
@@ -4,18 +4,18 @@ import { ForbiddenCharacters } from '../oslo/Util.js';
 
 describe('Serge parsing', () => {
 
-	const serge = new Serge('/root', 'test', '/root/test', {
-		name: 'test',
-		source_dir: 'src',
-		output_file_path: 'out'
-	});
+    const serge = new Serge('/root', 'test', '/root/test', {
+        name: 'test',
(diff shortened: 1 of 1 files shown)


adkulas:
No idea if eslint doesn't like this. I just did it in the browser

test/Serge.test.js
```diff
@@ -4,18 +4,18 @@ import { ForbiddenCharacters } from '../oslo/Util.js';
 
 describe('Serge parsing', () => {
 
-	const serge = new Serge('/root', 'test', '/root/test', {
```
pureooze, line 7: I dont think this changed anything other than whitespace.

adkulas: I think you're right

adkulas: but the spacing does seem off

adkulas: some places are 4 space indent some are 2

pureooze: What do you mean by "off"? Its  tabs so should be the same size 🤷

pureooze: Hmm I dont see that 😕 
image

adkulas: yeah I checked out the branch and its all tabs 🤷

PR #2549, opened 2020-10-05
Add Discover as new BSI dependency

BhamelinD2L:
Discover-FRA is being migrated into working as a BSI component. 
Therefore it'll need to be in BSI!

All old and unnecessary dependencies were removed from discover before this PR. The remaining paper dialog and dompurify will be much more easily removed post-migration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4870,23 +4870,23 @@
       }
     },
     "d2l-discovery-fra": {
-      "version": "github:Brightspace/discovery-fra#7ccab0c1ccc14c3ecb4cd48a1ef0427c90227240",
+      "version": "github:Brightspace/discovery-fra#c260fac18aa89f92db5fa02b258aae3d618835c3",
       "from": "github:Brightspace/discovery-fra#semver:^3",
       "dev": true,
       "requires": {
         "@brightspace-ui-labs/pagination": "0.0.8",
-        "@brightspace-ui/core": "^1.83.1",
+        "@brightspace-ui/core": "^1",
         "@polymer/app-route": "^3.0.0-pre.15",
         "@polymer/iron-pages": "^3.0.0-pre.15",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -4846,6 +4869,62 @@
...
+        "d2l-typography": "github:BrightspaceUI/typography#semver:^7",
+        "dompurify": "^2.0.7",
+        "fastdom": "^1.0.8",
+        "promise-polyfill": "8.1.0",
+        "siren-parser": "^8.2.0",
+        "siren-sdk": "github:BrightspaceHypermediaComponents/siren-sdk#semver:^1",
+        "url-polyfill": "^1.1.10",
+        "whatwg-fetch": "^3.4.1"
+      },
+      "dependencies": {
```
svanherk, line 4903: Yay to less FRAs!  We'll need to get these specific dependency versions gone to satisfy CI, which hopefully won't cause require any changes to the actual app code.  In your `package.json` you'll want to update to something like:
```
"@polymer/polymer": "^3",
...
"d2l-fetch-auth": "^1",
...
"d2l-hypermedia-constants": "^6",
```

You'll still get the latest `d2l-fetch-auth` and `d2l-hypermedia-constants`, it's just changing it to a consistent format that doesn't require a specific callout.  You will be getting a newer version of Polymer, but it was locked years ago so I expect the old issue is gone and you'll want to be on the latest anyways.

BhamelinD2L: Thanks for this - I've got everything updated without any issues cropping up in the repo. I think it's all set, but let me know if there's anything else you need taken care of!

svanherk (approved):
Looks good to me!  I believe you are now able to skip the `web-components/d2l-discovery-app.js` file and just point directly to `./node_modules/d2l-discovery-fra/src/discovery-app.js',` in the rollup file, but what you have should also still work!

PR #2550, opened 2020-10-05
Added quiz editor component

ronan-f:


wongvincent (approved):

web-components/d2l-activity-editor.js
```diff
@@ -1,2 +1,3 @@
 import 'd2l-activities/components/d2l-activity-editor/d2l-activity-assignment-editor/d2l-activity-assignment-editor.js';
+import 'd2l-activities/components/d2l-activity-editor/d2l-activity-quiz-editor/d2l-activity-quiz-editor.js';
```
dlockhart, line 2: Should these be separate imports? Mainly: do they both always appear together, or are there scenarios where only one or the other are needed?

ronan-f: @dlockhart I would say there are scenarios when only one or the other is needed. For example, if an instructor is creating a quiz then we probably don't want to import the assignment editor component as well.

In that case should they have their own files with separate imports?

wongvincent: Hmm.. I would think only one or the other would be needed at any one time. Either an assignment or a quiz editor would be on the page and not both.

Are you suggesting they be in separate files in `/web-components`?

dlockhart: Yeah sounds like they should be separate. Now that things are unbundled, you actually don't need anything here in `/web-components`. You can just create entries in `rollup-wc.config.js` that point directly into `node_modules`, similar [to this example](https://github.com/Brightspace/brightspace-integration/blob/master/rollup/rollup-wc.config.js#L77).

dlockhart: If you want a non-breaking way to do this:
1. Add individual entries to `rollup-wc.config.js` for each of these two files
2. Wait for BSI to deploy
3. In the LMS, switch the `ActivityEditorImport.codegen` file to point to the new location of the editor. Aside: there are actually 2 copies of that same codegen file, which isn't great... ideally there's a single definition for the asset and it gets loaded from a single view's renderer and it's the view itself that gets used in multiple places.
4. Once that merges, you can remove this file and the entry for it in `rollup-wc.config.js`.

PR #2569, opened 2020-10-07
Fix navigation group focus

dlockhart:
Noticed that only on legacy pages that navigation groups get an outline when focused:

Screen Shot 2020-10-07 at 3 47 46 PM

MVC pages aren't impacted because the `reset.css` stylesheet removes all outlines. Not sure when this was introduced... it may have also been a recent Chrome change. 🤷

### DIFF
--- a/rollup/rollup-wc.config.js
+++ b/rollup/rollup-wc.config.js
@@ -16,6 +16,7 @@ const componentFiles = [
 	'./node_modules/@brightspace-ui/core/components/button/button-subtle.js',
 	'./node_modules/@brightspace-ui/core/components/button/floating-buttons.js',
 	'./node_modules/@brightspace-ui/core/components/dialog/dialog-fullscreen.js',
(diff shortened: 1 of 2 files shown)


rollup/rollup-wc.config.js
```diff
@@ -16,6 +16,7 @@ const componentFiles = [
 	'./node_modules/@brightspace-ui/core/components/button/button-subtle.js',
 	'./node_modules/@brightspace-ui/core/components/button/floating-buttons.js',
 	'./node_modules/@brightspace-ui/core/components/dialog/dialog-fullscreen.js',
+	'./node_modules/@brightspace-ui/core/components/dropdown/dropdown.js',
```
dlockhart, line 19: Oops, didn't meant to include this here... but I _do_ need it haha so I guess it can stay!

dbatiste (approved):

PR #2632, opened 2020-10-23
US121650 - Add auto approval

svanherk:


.github/workflows/bsi-update.yml
```diff
@@ -87,3 +87,19 @@ jobs:
...
+        with:
+          script: |
+            github.pulls.createReview({
+              owner: 'Brightspace',
+              repo: 'brightspace-integration',
+              pull_number: process.env.PULL_REQUEST_ID,
+              event: 'APPROVE'
+            });
+            console.log('PR approved')
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
svanherk, line 105: Based on the LMS version I _think_ this will need to be a different token so the same token isn't opening and approving the request, but we an try this one first

j3parker (approved):

wongvincent (approved):

svanherk:
Yeah, looks like that's not allowed: https://github.com/Brightspace/brightspace-integration/actions/runs/324684521

PR #2635, opened 2020-10-23
US121650 - Fix Approval step if check

svanherk:
I figured `undefined` would be counted as a failure for the step's `if`, but it is trying to run the approval with an `undefined` PR id.


wongvincent (approved):
Good to know.

wongvincent (commented):
If you return before the log, then it won't log `console.log('PR already exists')` ?

j3parker (approved):

svanherk:
> If it returns before the log, then it won't log console.log('PR already exists') ?

🤦 Haha thanks!

PR #2657, opened 2020-10-27
Change to d2l-hc-activity-editor and move to appFiles.

m6jones:
Moved activity editor to appFiles. I am assuming this means it will only be loaded for the page that uses it?


dlockhart (approved):

m6jones:
Tested with a built copy of the BSI on my test site. Assignments works!

PR #2684, opened 2020-10-29
BSI Auto Update Dependencies

johngwilkinson:


package-lock.json
```diff
@@ -4929,7 +4929,7 @@
       }
     },
     "d2l-engagement-dashboard": {
-      "version": "github:Brightspace/insights-engagement-dashboard#af549c2c10650dfa96bdbd93886ba0fb43363a9b",
+      "version": "github:Brightspace/insights-engagement-dashboard#164ff771e9838875aa6c78814efdb3b3cb17b451",
```
bemailloux, line 4932: Nick's discussion fix is not going in this release, right? (Since it doesn't include discussions?)

johngwilkinson: Right. Other than a perf fix I put in a while back, I haven't heard that anything else besides translations is going in during cert.

bemailloux: Something is weird with the releases - the commit hash in this change is 1.0.2, but there's also a 1.0.2.1 which includes Nick's changes. So this PR looks correct but I don't know why the other release exists. Is that for dev / demo / allspark purposes?

bemailloux (approved):

package-lock.json
johngwilkinson: That's not intentional, as far as I know. I made 1.0.2.1 because I forgot to bump the version in package.json. But I seem to have messed up the tagging, maybe. I'll go chase that down.

johngwilkinson: This version should be good.

PR #2838, opened 2020-11-24
Bump discover-fra for routing update and remove labs facet filter sort

BhamelinD2L:
This PR bumps to include new features in the BSI version of Discover, including routing via lit-element-router. Also removes the labs `facet-filter-sort` dependency, leaving only the core version of this dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1138,20 +1138,6 @@
         "lit-element": "^2"
       }
     },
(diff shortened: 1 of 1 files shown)


svanherk:
Looks like the auto update has you covered 🙂  : https://github.com/Brightspace/brightspace-integration/pull/2839/files

PR #2879, opened 2020-11-30
US122120: Add file dropper dependency for Content Store - Add Content

MarquisLP:


package.json
```diff
@@ -95,6 +95,7 @@
     "eslint-config-brightspace": "^0.9",
     "eslint-plugin-html": "^6",
     "eslint-plugin-sort-class-members": "^1",
+    "file-drop-element": "^1.0.1",
```
svanherk, line 98: Do you need to use this directly in the monolith, or just inside the `d2l-content-store-add-content` component?  If just in that component, you should be able to just add it there and the automated PRs will include it for free.

package-lock.json
```diff
@@ -4832,7 +4832,8 @@
       "requires": {
         "@brightspace-ui/core": "^1",
         "@webcomponents/webcomponentsjs": "^2",
-        "lit-element": "^2"
+        "lit-element": "^2",
+        "file-drop-element": "^1.0.1"
```
svanherk, line 4836: Although, I'm not seeing this actually added in that component's `package.json`: https://github.com/Brightspace/content-store-add-content/blob/master/package.json#L39-L43
Which means the automated BSI PRs would just remove this the next time they run

MarquisLP:
Thanks @svanherk and @dlockhart for informing me about how the automated PRs work with new subdependencies. `file-drop-element` will indeed only be used within `d2l-content-store-add-content`. Since [the component's next version update](https://github.com/Brightspace/content-store-add-content/pull/12) will already include it, we can close this PR.

PR #2907, opened 2020-12-03
US118253: add engagement dashboard to oslo

johngwilkinson:
As I understand it, the BSI build will read this file, follow it to our .serge.json, then package our langterms; then the LMS build will get that package and include it in the language manager.

FYI @anhill-D2L @MykolaGalian @NicholasHallman @rohitvinnakota


bemailloux (approved):
Looks good! (according to docs: https://docs.dev.d2l/index.php/Oslo)

johngwilkinson:
The first build failed on the old dots in langterms; I've now updated with the BSI master, which should include the renamed langterms.

Vitalii-Misechko (approved):

PR #2908, opened 2020-12-03
Add static TinyMCE plugins/styles to BSI for the HTML editor component

grant-cleary:
TinyMCE only provides their free offerings in their NPM packages. Since the HTML editor component relies on plugins and skins from their Enterprise subscription offering (which we need to download directly), we need to manually add these to the editor repo. Unfortunately, because TinyMCE needs to use these files directly, we can't import them and need to add them to rollup as static files. ☹️

### DIFF
--- a/rollup/rollup-wc.config.js
+++ b/rollup/rollup-wc.config.js
@@ -117,6 +117,7 @@ const appFiles = [
 // NOTE: Ideally these should all be dynamically imported by apps.
 //       Please don't add new entries to this list.
 const staticFiles = [
+	'node_modules/@brightspace-ui/htmleditor/tinymce/**',
 	'node_modules/@d2l/d2l-attachment/locales/*.json',
 	'node_modules/@d2l/d2l-attachment/icons/*.svg',
 	'node_modules/d2l-activities/components/d2l-activity-collection-editor/lang/*.js',


rollup/rollup-wc.config.js
```diff
@@ -117,6 +117,7 @@ const appFiles = [
 // NOTE: Ideally these should all be dynamically imported by apps.
 //       Please don't add new entries to this list.
 const staticFiles = [
+	'node_modules/@brightspace-ui/htmleditor/tinymce/**',
```
grant-cleary, line 120: We definitely need at least some subset of every directory that lives in here. I can try to be more specific if we feel that would be necessary (e.g. finding plugins that don't rely on non-JS resources and trying to import those), but long term I feel like this is less fragile. I'm not super familiar with rollup though, so if I'm wrong, let me know. 😄

dlockhart: This seems good and more future-proof. Storage on S3 is cheap. :)

dlockhart (approved):

rollup/rollup-wc.config.js
dbatiste: Agreed.  I think including this with the wild-card is the way to go.

dbatiste (approved):

rollup/rollup-wc.config.js
svanherk: Question... are they cool with us putting enterprise files in a public repo?

grant-cleary: They've given us permission to do so in the past. The enterprise files currently live on our CDN, which (to my understanding) is fine by them. I have no idea whether our legal relationship with TinyMCE has changed since then though.

svanherk: Huh, just seems like an odd setup, but I guess if it's all client-side there's no real "hiding" it anyways

grant-cleary: Nor do I have a good understanding of whether files on a CDN would be different to them vs. having them live in the editor repo...

dbatiste: It's technically possible to copy them from their CDN also, so there shouldn't be an issue about putting them in a public place.  However, perhaps we should be removing the LICENSE file from the repo - that might be on shakey ground.

PR #2943, opened 2020-12-09
Add d2l-quick-eval-widget

AlexBedley:
This PR adds the `d2l-quick-eval-widget` application to BSI. It's already contained in `activities` so no need to change `package.json`


amauryperez (approved):

dlockhart:
@amauryperez would you be able to switch this so that it doesn't use the file in `web-components` and just references the assets directly from `node_modules` from `rollup-wc.config.js`? There are several examples of places doing this already in there. We're trying to move away from needing the `web-components` directory at all.

AlexBedley:
Yep no problem!

PR #2945, opened 2020-12-09
Dont use web-components, reference node_modules directly, this is the way

AlexBedley:
As per https://github.com/Brightspace/brightspace-integration/pull/2943#issuecomment-741983637


dlockhart (approved):
Thanks!

PR #3147, opened 2021-01-21
Update htmleditor component to v1

grant-cleary:


grant-cleary:
Ick, forgot to follow the steps for updating the package-lock. Will do that now.

svanherk:
You'll also need to make sure every repo in BSI that's using the editor is also at your new version before you can bump it.  Which I think it's only `activities`?  But basically, your PR will need at minimum the bump in `package.json` that you have, and an `activities` and `html-editor` bump in `package-lock.json`

grant-cleary:
> You'll also need to make sure every repo in BSI that's using the editor is also at your new version before you can bump it. Which I think it's only `activities`? But basically, your PR will need at minimum the bump in `package.json` that you have, and an `activities` and `html-editor` bump in `package-lock.json`

Yeah, I'm currently bumping the version in the activities editor, so I'm not going to bother updating this PR until that's done. :)

grant-cleary:
The package-lock regeneration also updated a couple unrelated dependencies. Should I be overly concerned (like should I be manually removing those changes)?

svanherk (approved):

PR #3154, opened 2021-01-21
bump version

stevenao:


mdulat:
Why are we making these changes in `package.json`?

stevenao:
> Why are we making these changes in `package.json`?

didn't realize that `npm` was changing the file around. I have reverted back to be safe. thanks.

mdulat (approved):

mdulat:
Looks like you don't need this any more as the bot already did it https://github.com/Brightspace/brightspace-integration/pull/3155

PR #3179, opened 2021-01-26
[Cert Fix 20.21.2] fix learning path defects

AKobets:
fixed: DE42228, DE420984, DE42126, DE42282


AKobets:
Just discovered bug introduced with this PR. Will close

PR #3188, opened 2021-01-27
remove Firefox ESR from our browserlist

dlockhart:
We dropped support for the Extended Support Release of Firefox with the January release. Removing it from here will cause old-Firefox vendor prefixes from appearing in our CSS and stop Babel from doing special JS transpiling specifically for Firefox ESR.


svanherk (approved):
🎉

dbatiste (approved):

PR #3208, opened 2021-01-29
insights-engagement-dashboard 20.21.2 with translations

bemailloux:
Also updates pagination because of a translation hiccup (https://github.com/BrightspaceUILabs/pagination/pull/27)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1212,9 +1212,9 @@
       }
     },
     "@brightspace-ui-labs/pagination": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/@brightspace-ui-labs/pagination/-/pagination-1.0.1.tgz",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1212,9 +1212,9 @@
       }
     },
     "@brightspace-ui-labs/pagination": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/@brightspace-ui-labs/pagination/-/pagination-1.0.1.tgz",
-      "integrity": "sha512-AUHiA1q06iTTbtCZDU6ST1clf/8dDUb5/tRYftiDtRprs+fSHDdYJ64gHfRDA4oqckbMbNVivGxbEYmr0SKc7A==",
+      "version": "1.0.2",
```
johngwilkinson, line 1215: is this compatible with everything else in BSI?

johngwilkinson (approved):

package-lock.json
bemailloux: I don't know. I think the build job will fail if it's not.

anhill-D2L (approved):

package-lock.json
bemailloux: Took a quick look around package.json. All repos that use this use the `^1` version so they should all be compatible.

PR #3211, opened 2021-01-29
[Cert Fix 20.21.2] translations for hmc foundation-components

AKobets:
Tested with local bsi


zinabat (approved):
Tested with local BSI and several languages

PR #3269, opened 2021-02-08
add ability to ignore dev packages when deduping front-end dependencies

dlockhart:
This is the script that goes through our `package-lock` file and makes sure that all things that are deemed "front-end" aren't including multiple versions of different dependencies. Something is "front-end" if it has Polymer or Lit as a dependency.

The change I'm making here is to add a list of packages that are ignored. So in this case `polymer-cli` includes `polymer` and it's deemed "front-end"... but it's really not so we want to ignore it.

This will let us add `lodash` as a dependency in core and ensure that everyone else that uses it is using the same version.


GeorgePlukov (approved):
LGTM

svanherk:
I'm a bit confused... why would someone need `polymer-cli` as a `dependency` and not just a `devDependency`?

dlockhart:
`package-lock.json` is BSI is what's being traversed, and it just has a flat list of `dependencies`. Each one has a `dev: true` (or false), but I'm not sure if I can rely on that to determine if something is truly dev or not.

svanherk:
> package-lock.json is BSI is what's being traversed, and it just has a flat list of dependencies. Each one has a dev: true (or false), but I'm not sure if I can rely on that to determine if something is truly dev or not.

Huh, I always thought `package-lock.json` would ignore `devDependencies` of the things we were pulling in.  Looks like `polymer-cli` is in the `package-lock.json` because `d2l-activity-exemptions` has it as a main dependency, which _seems_ like a mistake.  But then I don't know what the `dev` is for, which I also don't think we should trust.  It seems to like to disappear and come back a lot.

Regardless, code change looks fine!  I just wonder if we could instead cleanup https://github.com/Brightspace/d2l-activity-exemptions/blob/master/package.json#L21 for `polymer-cli` specifically

dlockhart:
Right I see. Yeah in that case I think we'd also need to move all the D2L things in BSI's own `package.json` from dev -> prod dependencies. I can try giving that a try...

test-package-lock.js
```diff
@@ -1,43 +1,41 @@
...
 	depth++;
 	for (const key in json) {
-		if (depth > 1) {
-			const isPolymer = json[key].requires !== undefined && json[key].requires['@polymer/polymer'] !== undefined;
-			const isLit = json[key].requires !== undefined && (json[key].requires['lit-element'] !== undefined || json[key].requires['lit-html']);
-			if (isPolymer || isLit || packages.indexOf(key) > -1) {
-				console.error(`Duplicate dependency detected "${key}" in "${parentKey}". All front-end dependencies must be at root level of "package-lock.json" to avoid duplicate registrations. Check that the version ranges in "package.json" do not contain anything beyond the major version.`);
-				process.exitCode = 1;
+		const isDev = json[key].dev !== undefined;
+		if (!isDev) {
```
dlockhart, line 20: Based on @svanherk's input I've changed the approach here. By properly putting BSI's dependencies in its `package.json` in dev or non-dev, we can use the `"dev": true` flag when scanning these. If we see a non-dev dependency anywhere except the root, that means there are at least 2 copies of it and we can fail.

This removes the need to look for Lit/Polymer things, which is pretty sweet!

This uncovered 1 issue with `isarray`, but it's not easily fixable (and not new) so I'm suppressing it with a warning for now.

package.json
```diff
@@ -89,18 +75,33 @@
...
     "eslint-config-brightspace": "^0.9",
     "eslint-plugin-html": "^6",
     "eslint-plugin-sort-class-members": "^1",
     "frau-publisher": "^2",
     "glob": "^7",
-    "intl-messageformat-parser": "^3",
     "mocha": "^6",
     "moment-timezone": "^0.5",
     "node-sass": "^4",
-    "parse-srt": "^1.0.0-alpha",
```
dlockhart, line 103: This was added by the media-player by mistake -- it's listed as a dependency in that repo.

svanherk (approved):

PR #3350, opened 2021-02-21
Add support for dynamic imports

bearfriend:
[This plugin](https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#how-it-works) allows for the use of dynamic `import` statements by replacing variables with glob patterns to find and add files to the build. It then creates a `switch` to handle each case.

### DIFF
--- a/rollup/rollup-wc.config.js
+++ b/rollup/rollup-wc.config.js
@@ -1,6 +1,7 @@
 import { createBasicConfig } from '@open-wc/building-rollup';
 import copy from 'rollup-plugin-copy';
 import replace from '@rollup/plugin-replace';
+import dynamicImportVars from '@rollup/plugin-dynamic-import-vars';
 import merge from 'deepmerge';
 
 const componentFiles = [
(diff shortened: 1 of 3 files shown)


rollup/rollup-wc.config.js
```diff
@@ -157,6 +158,9 @@ export default merge(config, {
 		replace({
 			define: 'defineNoYouDont', /* prevents UMD time bomb as fastdom will try to call define() on UMD FRA pages */
 			include: ['node_modules/fastdom/fastdom.js', 'node_modules/focus-visible/dist/focus-visible.js']
+		}),
+		dynamicImportVars({
+			exclude: 'node_modules/d2l-html-editor/d2l-html-editor.js',
```
dlockhart, line 163: Why did we need to exclude the HTML editor, is it doing something wrong?

bearfriend: It's doing `import(tinymceBaseUrl+'/tinymce.min.js')`, but since that is preexisting I assumed getting what it needs into the build is already handled. Didn't look into it beyond that.

dlockhart: Ok. That's the legacy HTML editor anyway, so hopefully the new one "just works". @dbatiste FYI.

dlockhart (approved):

PR #3358, opened 2021-02-22
adding d2l-card

nleipoldd2l:


dlockhart (approved):
Cool. We can add `card-footer-link.js` and all the other helper components later when & if we end up needing them.

PR #3437, opened 2021-03-05
Adding a sample foundations component to rollup

zinabat:
## Context
[US123769](https://rally1.rallydev.com/#/357252275780d/custom/367300408400?detail=%2Fuserstory%2F480099321508&fdp=true&fdp=true?fdp=true)
Adding a sample component for use with a legacy page.
This component will be replaced with the real one once the MVC and Legacy adapters have been created and are functional.


zinabat:
Need to fix localization first in the component repo.

BhamelinD2L (approved):

PR #3472, opened 2021-03-10
Add ability to change hostname/port

rschick:


dlockhart (commented):
I think you can do this without introducing environment variables? Like with an NPM script you can pass extra arguments to the underlying script using `--`.

So for `es-dev-server` you could go:

```shell
npm start -- --port=1234 --hostname=foo
```

rschick:
I couldn't get it to work that way, but `$npm_config_` works.. (TIL!)

package.json
```diff
@@ -11,7 +11,7 @@
     "publish:autotag": "node ./cli/autotag.js",
     "publish:cdn": "frau-publisher -m lib -t bsi -f \"./build/**/*\" --versionvar TRAVIS_TAG",
     "start": "npm run start:dev",
-    "start:dev": "concurrently --kill-others \"es-dev-server --config es-dev-server-dev.config.js\" \"rollup -w -c ./rollup/rollup.config.js\"",
+    "start:dev": "concurrently --kill-others \"es-dev-server --config es-dev-server-dev.config.js --hostname $npm_config_hostname --port $npm_config_port\" \"rollup -w -c ./rollup/rollup.config.js\"",
```
dlockhart, line 14: Running this without passing anything changes the port to 8000 for me.

rschick:
Good catch, I tested it and didn't notice the port was wrong. Updated with a cleaner approach.

dlockhart (approved):
👍

PR #3511, opened 2021-03-16
Adding cy-gb, fr-fr, fr-on and updating zh-cn for oslo

JordanCorner:
US126015

- fr-on needs to be added for upcoming fr-on work
- cy-gb, fr-fr exist but haven't been part of oslo until now
- zh-cn should just be zh-cn, not moved to 2 chars (this doesn't actually change anything functionally, but brings us in line with how the fallbacks should be working)

### DIFF
--- a/oslo/generate-monolith-xml
+++ b/oslo/generate-monolith-xml
@@ -11,18 +11,21 @@ const Serge = require('./Serge');
 
 const Languages = [
 	new Lang('ar-sa', 'ar', 'ar-SA'),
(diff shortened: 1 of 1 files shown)


kieranderson:
If no components have this lang file, do we create an empty language xml or just skip it?

JordanCorner:
> If no components have this lang file, do we create an empty language xml or just skip it?

Create empty. Tested that locally and there is currently an fr-on package that is empty, which should be ok for now.

kieranderson (approved):

patel-kajal (approved):

PR #3515, opened 2021-03-16
handle scripts trying to access process.env.NODE_ENV, specifically mobx

dlockhart:
Certain client-side scripts have conditional code/logic keyed off `process.env.NODE_ENV`. Specifically React likes to do this, as does `mobx`.

`mobx` gets pretty cranky if it doesn't see "production" in there... specifically it's getting `undefined`, so it complains thus:
> [mobx] you are running a minified build, but 'process.env.NODE_ENV' was not set to 'production' in your bundler. This results in an unnecessarily large and slow bundle

Looking for `process.env.NODE_ENV` is so common that the [replace Rollup plugin](https://github.com/rollup/plugins/tree/master/packages/replace#usage) uses this as its main example for using it. So this is the approach I've taken to address it.

While I was in here, I noticed warnings about the `preventAssignment` option soon defaulting to `true` and we should start using that as soon as possible, so I've enabled it. More on that [and why it's a good idea](https://github.com/rollup/plugins/tree/master/packages/replace#preventassignment).

### DIFF
--- a/rollup/rollup-wc.config.js
+++ b/rollup/rollup-wc.config.js
@@ -165,7 +165,12 @@ export default merge(config, {
 		}),
 		replace({
 			define: 'defineNoYouDont', /* prevents UMD time bomb as fastdom will try to call define() on UMD FRA pages */
-			include: ['node_modules/fastdom/fastdom.js', 'node_modules/focus-visible/dist/focus-visible.js']
+			include: ['node_modules/fastdom/fastdom.js', 'node_modules/focus-visible/dist/focus-visible.js'],
+			preventAssignment: true
+		}),
+		replace({
+			'process.env.NODE_ENV': JSON.stringify('production'),
+			preventAssignment: true
 		}),
 		dynamicImportVars({
 			exclude: 'node_modules/d2l-html-editor/d2l-html-editor.js',


bemailloux:
I'm not very familiar with Rollup. Does this go through all the code and replace `process.env.NODE_ENV` with "production"? And does this have any effect on the BSI bundle in terms of size or perf?

dlockhart:
> Does this go through all the code and replace process.env.NODE_ENV with "production"?

Correct. So if the code used to say `if (process.env.NODE_ENV === 'production') {` it'll now be `if ('production' === 'production') {`.

> does this have any effect on the BSI bundle in terms of size or perf?

Bundle size, I wouldn't expect so since it's just replacing 20 characters with 12. Performance, possibly within mobx if it was doing things different based on the value of that "environment variable".

bemailloux:
Is there a difference between this approach and actually setting `NODE_ENV = production` in the build job?

dlockhart:
Yes, that's what I tried initially and it did nothing. I think it's just because all our stuff does is rely on ES6 module imports to load stuff... there's no special "react build" or anything that's going to do anything magical like this for us.

bemailloux (approved):

svanherk (approved):

PR #3530, opened 2021-03-18
upgrading table to 2.4.4 to pick up fix for sticky headers

dlockhart:
This adds the [changes from this PR](https://github.com/BrightspaceUI/table/pull/227) to fix an issue with column overlap in sticky headers in the cert stage of `20.21.4`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5615,7 +5615,7 @@
       }
     },
     "d2l-table": {
-      "version": "github:BrightspaceUI/table#bd6e3de0ccf3f855ad22d4fef1e2e85352b93415",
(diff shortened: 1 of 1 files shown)


svanherk (approved):
Noel's new action got me thinking... I wonder if it would be worth making this trigger an auto PR in the LMS cert branch like it does for master.  I assume it was never added originally for simplicity/time sake, not because it's technically impossible.

It probably wasn't deemed worth it because cert fixes shouldn't happen a lot, but I see at least 15 usually.  And maybe we lock it down to cert, not hotfix

PR #3575, opened 2021-03-24
removing offset now that -10 is the default

dlockhart:
As per [this change](https://github.com/BrightspaceUI/table/pull/233), `-10px` is now the default so no need for this code.


svanherk (approved):
🎉

PR #3674, opened 2021-04-09
DE43118 - Update BSI to use hotfixed version of the HTML editor component

grant-cleary:
I... Don't entirely know if I've done this right, so if I haven't please feel free to yell at me. The editor needed to be hotfixed to include a fix to ISF (https://github.com/BrightspaceUI/htmleditor/pull/155).

This also contains two unrelated commits (1.4 and 1.5 of the editor respectively) that have the net result of introducing a new API. This shouldn't impact existing apps, but would let us backport an LMS-side fix that would need to use this new API if it turns out this is necessary. I would rather just bundle the changes together and get away with one BSI release as opposed to having to do another one later if it turns out we need to backport the API changes after all. If others disagree I'm open to changing this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1393,9 +1393,9 @@
       }
     },
     "@brightspace-ui/htmleditor": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/@brightspace-ui/htmleditor/-/htmleditor-1.3.1.tgz",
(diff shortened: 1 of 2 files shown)


svanherk:
PR looks good from a backporting POV 👍 

I'll let Dave speak to the second half about what should or shouldn't be backported.  Link for convenience: https://github.com/BrightspaceUI/htmleditor/compare/v1.3.1..v1.5.2

dbatiste (approved):

PR #3728, opened 2021-04-19
Bump htmleditor version in cert (contains % height fix)

grant-cleary:
Didn't quite make branching on this HTML editor change last week. For reference, this is what we're backporting: https://github.com/BrightspaceUI/htmleditor/compare/v1.7.6...v1.7.7

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1414,9 +1414,9 @@
       }
     },
     "@brightspace-ui/htmleditor": {
-      "version": "1.7.6",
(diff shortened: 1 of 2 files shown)


grant-cleary:
This is branched off of v20.21.5-215, which _should_ still be the last 20.21.5 BSI release (as per https://github.com/Brightspace/lms/pull/9404, the tag list, and the NQC commit reports).

dbatiste (approved):

PR #3897, opened 2021-05-14
US126758 - Add custom early progress report app

jorge-ramirez-arredondo:
Part of https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F519546414124&fdp=true?fdp=true

### DIFF
--- a/rollup/rollup-wc.config.js
+++ b/rollup/rollup-wc.config.js
@@ -96,6 +96,7 @@ const appFiles = [
 	'./node_modules/@brightspace-hmc/foundation-components/features/lti/d2l-lti-activity.js',
 	'./node_modules/d2l-activities/components/d2l-work-to-do/d2l-work-to-do.js',
 	'./node_modules/d2l-engagement-dashboard/engagement-dashboard.js',
+	'./node_modules/d2l-custom-early-progress-report/custom-early-progress-report.js',
 	'./web-components/d2l-activity-alignments.js',
 	'./web-components/d2l-activity-collections.js',
 	'./web-components/d2l-activity-editor.js',
--- a/.serge-mapping.json
+++ b/.serge-mapping.json
@@ -3,5 +3,6 @@
   "d2l-activities": "d2l-activities/activities.serge.json",
   "d2l-engagement-dashboard": "d2l-engagement-dashboard/insights-engagement-dashboard.serge.json",
   "@brightspace-hmc/foundation-components": "@brightspace-hmc/foundation-components/foundation-components.serge.json",
-  "d2l-cpd": "d2l-cpd/continuous-professional-development.serge.json"
+  "d2l-cpd": "d2l-cpd/continuous-professional-development.serge.json",
+  "d2l-custom-early-progress-report": "d2l-custom-early-progress-report/custom-early-progress-report.serge.json"
 }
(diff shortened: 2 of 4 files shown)


imcdonald-d2l:
Do we have serge setup in early progress tool? If so it looks like we might need to do this: https://github.com/Brightspace/brightspace-integration/blob/master/docs/web-components.md#lang-terms

jorge-ramirez-arredondo:
> Do we have serge setup in early progress tool? If so it looks like we might need to do this: https://github.com/Brightspace/brightspace-integration/blob/master/docs/web-components.md#lang-terms

Ah, yeah. I missed that. Just added it now.

imcdonald-d2l:
We shouldn't merge this until `custom-early-progress-report.js` is merged on your separate PR to be safe. That way we aren't referencing something that doesn't exist yet.

imcdonald-d2l (approved):

jorge-ramirez-arredondo:
> We shouldn't merge this until `custom-early-progress-report.js` is merged on your separate PR to be safe. That way we aren't referencing something that doesn't exist yet.

Do you mean in the custom-early-progress-report repo? There should be a `custom-early-progress-report.js` file in there already (just a sample component that is created with the template).

rollup/rollup-wc.config.js
```diff
@@ -96,6 +96,7 @@ const appFiles = [
 	'./node_modules/@brightspace-hmc/foundation-components/features/lti/d2l-lti-activity.js',
 	'./node_modules/d2l-activities/components/d2l-work-to-do/d2l-work-to-do.js',
 	'./node_modules/d2l-engagement-dashboard/engagement-dashboard.js',
+	'./node_modules/d2l-custom-early-progress-report/custom-early-progress-report.js',
```
imcdonald-d2l, line 99: In this PR you are adding `custom-early-progress-report.js` to `src/components/custom-early-progress-report.js`, we should adjust this so it references the proper file. 
https://github.com/Brightspace/custom-early-progress-report/pull/3/files#diff-aeed9368f0cb092cdbd223ba0aa48fc3a19e689b102eb50aa7942a404a80776b

imcdonald-d2l:
> > We shouldn't merge this until `custom-early-progress-report.js` is merged on your separate PR to be safe. That way we aren't referencing something that doesn't exist yet.
> 
> Do you mean in the custom-early-progress-report repo? There should be a `custom-early-progress-report.js` file in there already (just a sample component that is created with the template).

Yes, we should wait on [this PR](https://github.com/Brightspace/custom-early-progress-report/pull/3). See previous comment, no point creating two PRs in BSI if we can just merge one after your other work in early progress repo is merged.

rollup/rollup-wc.config.js
jorge-ramirez-arredondo: I didn't remove the file at the root though, I left it there as an entry point so that it could follow the pattern set by BSI:
https://github.com/Brightspace/custom-early-progress-report/pull/3/files#diff-49a3426eebcf59ecf1e368f39701b9f45b0febdd81e139fee0d8735fb2646511R1

I think doing it this way makes it so we don't have to worry about changes within the src folder messing up the BSI integration.

Now that I say that though, I guess the BSI entry point file could be renamed to better signal that.

PR #3953, opened 2021-05-26
Using Lit table

dlockhart:
These changes go alongside the ones [in d2l-table](https://github.com/BrightspaceUI/table/pull/243) and [in the monolith](https://github.com/Brightspace/lms/pull/10893).

This switches all BSI things over to use the new Lit-based tables and scroll-wrapper.


web-components/d2l-table.js
```diff
@@ -1,15 +1,8 @@
...
-	</style>
-</custom-style>`;
-
-document.head.appendChild($_documentContainer.content);
+if (document.getElementById('d2l-table-style') === null) {
+	const style = document.createElement('style');
+	style.id = 'd2l-table-style';
+	style.appendChild(document.createTextNode(tableStyles.cssText));
+	document.head.appendChild(style);
+}
```
dlockhart, line 8: This mirrors the approach used in `d2l-table` to get Lit's table styles into the document. If you're in Lit this isn't necessary since you'd import the table styles into your Lit component. But the monolith can't wrap all usages of `<d2l-table-wrapper><table class="d2l-table">` in a Lit element, so bringing these styles globally into the document is needed.

web-components/d2l-table.js
```diff
@@ -1,15 +1,8 @@
...
+import { tableStyles } from '@brightspace-ui/core/components/table/table-wrapper.js';
 
-$_documentContainer.innerHTML = `<custom-style>
-	<style is="custom-style" include="d2l-table-style">
-		.d2l-grid-mvc > * > tr > td {
-			vertical-align: top;
-		}
-		.d2l-grid-mvc > * > tr > th {
-			vertical-align: middle;
-		}
```
dlockhart, line 11: I looked into these overrides. `th` cells have `middle` vertical-align anyway, and the first rule wasn't actually working to override anything as far back as I can test. Maybe one day early in Daylight these did something, but no more. Figured it was safer to just remove them than bring back old behaviour that we haven't seen in a while.

package.json
```diff
@@ -73,7 +73,6 @@
     "d2l-scroll-spy": "BrightspaceUI/scroll-spy#semver:^1",
     "d2l-sequences": "BrightspaceHypermediaComponents/sequences#semver:^2",
     "d2l-simple-overlay": "Brightspace/simple-overlay#semver:^3",
-    "d2l-table": "BrightspaceUI/table#semver:^2",
```
dlockhart, line 76: This unfortunately won't impact the lock file until all other uses of `d2l-table` are removed.

svanherk (approved):

PR #4094, opened 2021-06-16
US128385 [AQ][Filter] Apply tree filter selection API

Vitalii-Misechko:
[US128385](https://rally1.rallydev.com/#/23525809118d/iterationstatus?detail=%2Fuserstory%2F605488757484&sharedViewId=328666453148&view=dc9e7043-1cf6-486d-a4aa-3a212976e789&fdp=true?fdp=true): [AQ][Filter] Apply tree filter selection API

This PR adds `@brightspace-ui-labs/lit-router` and `@brightspace-ui-labs/ou-filter` to BSI in order to be available in Assesment Quality Dashboard

* Functional Testing 
  * [x] Test cases
    * Verify that tree filter is available on Draco LMS
* Security, devops, perf, UX and docs: N/A
  
FYI @anhill-D2L @hyehayes @johngwilkinson @NicholasHallman @bemailloux @JikaiLong @SEsseltine

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,9 @@
   },
   "homepage": "https://github.com/Brightspace/brightspace-integration",
   "dependencies": {
+    "@brightspace-ui-labs/lit-router": "^1",
     "@brightspace-ui-labs/media-player": "^1",
+    "@brightspace-ui-labs/ou-filter": "^1",
     "@brightspace-ui-labs/pagination": "^1",
     "@brightspace-ui-labs/user-feedback": "^1",
     "@brightspace-ui-labs/user-profile-card": "^0.5.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1345,6 +1345,18 @@
         "lit-element": "^2"
       }
     },
+    "@brightspace-ui-labs/lit-router": {
+      "version": "1.0.7",
+      "resolved": "https://registry.npmjs.org/@brightspace-ui-labs/lit-router/-/lit-router-1.0.7.tgz",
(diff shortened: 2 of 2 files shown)


bemailloux (approved):

package.json
```diff
@@ -29,7 +29,9 @@
   },
   "homepage": "https://github.com/Brightspace/brightspace-integration",
   "dependencies": {
+    "@brightspace-ui-labs/lit-router": "^1",
```
dlockhart, line 32: Is `ou-filter` using the router? If so, it should be a dependency of `ou-filter`, not BSI.

Vitalii-Misechko: No, it is not. [Assessment Quality Dashboard ](https://github.com/Brightspace/insights-assessment-quality-dashboard)uses both router and `ou-filter`

Vitalii-Misechko:
> And is the `ou-filter` being used in the LMS directly, outside of `d2l-assessment-quality-dashboard`? If not, adding them to `d2l-assessment-quality-dashboard` as dependencies will automatically pull them into BSI for you

That is exactly what I'm trying to achieve (to use components inside `d2l-assessment-quality-dashboard`). I ran local BSI with linked d2l-assessment-quality-dashboard and used that local hosted BSI on siteloard site. When I navigated to  Assessment Quality Dashboard local BSI server showed an error that it can't find `ou-filter` and `lit-router`. 

From your comment, it sounds like it should work without these changes. I'll make another attempt. Thanks!

package.json
dlockhart: Ok, then no need to list it here in BSI's dependency list as it'll get pulled in by that project as one of its dependencies.

Sometimes when you're doing local development and there's a new dependency it doesn't always get picked up by `npm link`, so that could be the issue.

Vitalii-Misechko: That makes sense then. Thanks!

PR #4109, opened 2021-06-17
Certfix image-banner-overlay and my-courses

annabelsegalld2l:
These changes are for a certfix to allow auto-loading of images, after the behaviour was broken by changing the wrappers from simple-overlay to d2l-dialog-fullscreen.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5196,7 +5196,6 @@
       "from": "github:Brightspace/d2l-image-banner-overlay#semver:^4",
       "requires": {
         "@brightspace-ui/core": "^1",
-        "@polymer/iron-scroll-threshold": "^3.0.0-pre.18",
         "@polymer/polymer": "^3.0.0",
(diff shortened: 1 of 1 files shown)


svanherk (approved):
Looks good!

svanherk:
Although, I wonder if it'll be mad these changes aren't listed: https://github.com/Brightspace/d2l-image-banner-overlay/pull/117/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

svanherk:
I don't _think_ so because the observer is already included and the iron-threshold can't be removed, so I'd expect the actual building to be fine.  But I did just kick off the auto update if you want to see what it comes up with an copy that

annabelsegalld2l:
I added those dependencies to package.json to match the auto-update diff!

svanherk (approved):

PR #4180, opened 2021-06-25
Suppress MVC framework validation tooltip in LMS (use the editor component's tooltip instead)

grant-cleary:
We're building validation into the editor component, so we would prefer to suppress the MVC framework's tooltip and rely on the component's form validation instead.

### DIFF
--- a/sass/html-editor/html-editor.scss
+++ b/sass/html-editor/html-editor.scss
@@ -177,6 +177,6 @@
 	}
 }
 
-d2l-htmleditor ~ .vui-validation-bubble {
+d2l-htmleditor + .vui-validation-bubble {
 	display: none;
 }
--- a/sass/html-editor/html-editor.scss
+++ b/sass/html-editor/html-editor.scss
@@ -176,3 +176,7 @@
 		background-color: $d2l-color-sylvite;
 	}
 }
+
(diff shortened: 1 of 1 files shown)


dbatiste (approved):

sass/html-editor/html-editor.scss
```diff
@@ -176,3 +176,7 @@
 		background-color: $d2l-color-sylvite;
 	}
 }
+
+d2l-htmleditor ~ .vui-validation-bubble {
```
dbatiste, line 180: Are you sure you don't want the `+` combinator?

grant-cleary: Yeah, that's likely more specific. Each editor should be immediately followed by exactly one validation bubble (if one is present at all).

sass/html-editor/html-editor.scss
```diff
@@ -176,3 +176,7 @@
 		background-color: $d2l-color-sylvite;
 	}
 }
+
+d2l-htmleditor + .vui-validation-bubble {
+	display: none;
```
grant-cleary, line 181: Crud. Tabs vs. spaces in GitHub's UI.

grant-cleary: Oh, never mind, in this case I think this is actually what we want.

PR #4286, opened 2021-07-14
Fix UMD timebomb caused by the focus-visible polyfill

svanherk:
Prod builds are fine, but dev was mixing the fix.


dlockhart (approved):
Thanks!

PR #4292, opened 2021-07-15
DE44358: bump core, mycourses, image-banner in 20.20.7

annabelsegalld2l:
The changes are already in 20.20.8. Backporting them to 20.20.7 for an upcoming hotfix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1445,9 +1445,9 @@
       }
     },
     "@brightspace-ui/core": {
-      "version": "1.139.4",
-      "resolved": "https://registry.npmjs.org/@brightspace-ui/core/-/core-1.139.4.tgz",
-      "integrity": "sha512-QruuErSfmTeimleLeuSNmh6CzbN7ZK1xsSDI5rb1kzW+NZ6LYvJJuvdyeoB/86PJnNwT848hCyDl4eEoiyd9KQ==",
+      "version": "1.139.5",
+      "resolved": "https://registry.npmjs.org/@brightspace-ui/core/-/core-1.139.5.tgz",
+      "integrity": "sha512-F9RWkR+JNFUMR5LiJp5EsWAtNesUJcNFNsJE5vns5EXyTs8aK4k5XV3GHiWtY410eVnWhJ6hxgDGgLlsqJ1urg==",
       "requires": {
         "@brightspace-ui/intl": "^3",
         "@formatjs/intl-pluralrules": "^1",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -5192,7 +5192,7 @@
       }
     },
     "d2l-image-banner-overlay": {
-      "version": "github:Brightspace/d2l-image-banner-overlay#ceb3663c037070479b8bf458f8978e51761d6f5a",
+      "version": "github:Brightspace/d2l-image-banner-overlay#190335cbfcc43252979cb14fe01c5415cbcd3776",
```
annabelsegalld2l, line 5195: I'm a bit concerned about this change, as a release did not get generated for that PR: https://github.com/Brightspace/d2l-image-banner-overlay/issues/123
The build has passed though, so I'm not sure if a manual release is even needed.

svanherk: Hmm yeah, that repo has the problem where the next change after branching was a patch release and not a minor one, so `semantic-release` wants you to make the `4.3.x` off the last patch version for `4.3` - https://github.com/Brightspace/d2l-image-banner-overlay/releases/tag/v4.3.3.  We try to make minor release in `core` after branching to avoid this.

We're always a little torn here - we want to limit changes in a hotfix, but all patches _should_ be able to go in.  There's a ton of translation changes there though, that probably should've been a minor release if we didn't want them backported.

svanherk: What you have should work, it's just a little confusing to come back to later.  So what I'd probably do, because this repo isn't publishing to `npm` and we _can_ do this... is make a branch called `20.21.7` off of the `ceb3663c037070479b8bf458f8978e51761d6f5a` commit.  Then apply your changes there (this might be the same as just renaming the `4.3.x` branch, if that's the commit you originally branched off of).  Then you could make a manual release off that new branch called `v4.3.1-20.21.7-hotfix-1` or whatever, and if _another_ hotfix has to go in someone could just commit to that new branch and make another manual release.

That might be overkill but helps with tracking, and is only an option because that repo uses the `semver:^4` stuff, and doesn't pull from `npm`.  I can help if what I said above makes no sense... thoughts?

annabelsegalld2l: Could we do another PR with a manual version downgrade on that branch (like 4.2.0) and generate a patch release 4.2.1? Just an idea

annabelsegalld2l: Ah, didn't see your comment. I'll try that first.

annabelsegalld2l: Now updated image-banner to point at https://github.com/Brightspace/d2l-image-banner-overlay/commit/07cd7e3f061ea3b5f4e70bdcb9622102d91c4788. Should be the same results, but it is better for tracking purposes.

svanherk (approved):

dlockhart (approved):

PR #4293, opened 2021-07-15
Add copy-role web component

svanherk:


rollup/rollup-wc.config.js
```diff
@@ -89,18 +89,19 @@ const componentFiles = [
 	'./web-components/mathjax.js'
 ];
 const appFiles = [
```
svanherk, line 91: Just trying to alphabetize below.  The lower half is not either, but baby steps

dlockhart (approved):

PR #4304, opened 2021-07-16
updating browserslist to be more inclusive

dlockhart:
This now [mirrors what the LMS will do](https://github.com/Brightspace/lms/pull/12444/files).

I've gone with supporting the browsers that aren't "unsupported", which should help out those using versions in the "legacy" range today that weren't falling into the "last 2 versions" bucket.

It'll also let us update the Edge line in a month or so to be `>= 79` without causing other breakages.


svanherk (approved):
Nice!

PR #4308, opened 2021-07-19
using shared config

dlockhart:
These configs are shareable and since we now need them in at least 2 places, it makes sense to centralize this.

Shared config is here:
https://github.com/BrightspaceUI/browserslist-config


svanherk (approved):
Nice!

PR #4403, opened 2021-07-29
fix: empty commit to make sure DE44936 gets merged

mbardakovD2L:
See https://github.com/BrightspaceHypermediaComponents/foundation-components/commit/9a42dc800ada4cdd332147c75aaa75cb1b52ddc7, https://rally1.rallydev.com/#/357251704080ud/iterationstatus?detail=%2Fdefect%2F603983938391


bearfriend (approved):

svanherk:
An empty commit will just create a copy of the last release with no changes, but I think the change you want made it in here? https://github.com/Brightspace/brightspace-integration/pull/4404/files

bearfriend:
Whoops! Yes, this was supposed to be an empty commit to `foundation-components`, but it looks like it has since been released anyway so we're good to go.

PR #4413, opened 2021-07-30
adding a GitHub action workflow for CI to monitor how it goes

dlockhart:
The plan here is to just have this run in parallel with the main Travis CI job and monitor it over time to see how stable it is. If things look good, we'll have more confidence that switching over to GHA for the rest of the jobs makes sense.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
+name: CI
+on:
+  push:
+    branches:
+      - master
+jobs:
+  build:
+    name: Build
+    timeout-minutes: 15
```
dlockhart, line 9: This is a guess... most Travis runs take ~10 minutes.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
+name: CI
+on:
+  push:
+    branches:
+      - master
```
dlockhart, line 5: I could also run it on CI, but figured this would be less disruptive and would accomplish the same thing.

svanherk: Ah good idea, I misunderstood what you were saying in the retro 👍

svanherk (approved):

PR #4416, opened 2021-07-30
Add input-duration component to rollup

jorge-ramirez-arredondo:
[US127919](https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F603444721280&fdp=true?fdp=true)

https://github.com/BrightspaceUILabs/input-duration

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   },
   "homepage": "https://github.com/Brightspace/brightspace-integration",
   "dependencies": {
+    "@brightspace-ui-labs/input-duration": "^1.1.1",
     "@brightspace-ui-labs/media-player": "^1",
     "@brightspace-ui-labs/pagination": "^1",
     "@brightspace-ui-labs/status-icon": "^0.0.3",
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "node-sass": "^4",
     "rally": "^2",
     "rimraf": "^3",
-    "rollup": "2.48.0",
+    "rollup": "^2",
     "rollup-plugin-copy": "^3",
     "rollup-plugin-postcss": "^3",
     "rollup-plugin-uglify": "^6",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
   },
   "homepage": "https://github.com/Brightspace/brightspace-integration",
   "dependencies": {
+    "@brightspace-ui-labs/input-duration": "^1.1.1",
```
dlockhart, line 32: We typically don't bother specifying the minor/patch release just to keep things clean. Then the lock file is where the nitty gritty is tracked anyway.

```suggestion
    "@brightspace-ui-labs/input-duration": "^1",
```

dlockhart (approved):
Looks good. I kicked the build as it failed on a pesky memory problem we've been seeing more and more.

dlockhart (approved):

dlockhart:
I was able to get this going by increasing the `--max-old-space-size` flag via `NODE_OPTIONS` environment variable in Travis CI. Disaster averted for now I think.

jorge-ramirez-arredondo:
> I was able to get this going by increasing the `--max-old-space-size` flag via `NODE_OPTIONS` environment variable in Travis CI. Disaster averted for now I think.

Is this going to continue to cause problems if more and more packages are added to BSI?

dlockhart:
> Is this going to continue to cause problems if more and more packages are added to BSI?

I don't _think_ so. The way I understand that flag is that it's telling the garbage collection to get more aggressive as it reaches that value. So if we've already reached it here with this PR and this flag solved it, then in theory adding more and more shouldn't matter. But who knows!

Stacey and I will be starting a project shortly that will try to find better ways to serve up our new component-based apps than BSI. It's problematic for several reasons, this memory thing just being the latest.

PR #4461, opened 2021-08-06
switching from es-dev-server to web-dev-server

dlockhart:
This switches BSI over from using the deprecated `es-dev-server` to its replacement, [@web/dev-server](https://modern-web.dev/docs/dev-server/overview/).

It's only used on local dev machines -- so the risk here to prod is low -- but I've tested it in both "dev" and "prod" build modes.

Unfortunately a log of the configuration has changed, been plugin-ified, removed or renamed. Pretty I've got everything switched over though. While I was at it, I opted into the ES module syntax for the config files, which resulted in some `.eslintrc.json` updates since the default at the root changed from `script` to `module`.


web-dev-server-dev.config.mjs
```diff
@@ -16,7 +13,6 @@ module.exports = {
 			return next();
 		}
 	],
-	nodeResolve: true,
```
dlockhart, line 19: This changed to a `--node-resolve` CLI flag.

web-dev-server-prod.config.mjs
```diff
@@ -0,0 +1,11 @@
+import cors from '@koa/cors';
+
+export default {
+	hostname: 'localhost',
+	port: 8080,
+	esbuildTarget: 'esnext',
```
dlockhart, line 6: This is the equivalent to "none" from before. We don't want the dev server doing any transpilation since Rollup will have already done it for the prod builds. It's just a glorified web server at this point.

oslo/.eslintrc.json
```diff
@@ -5,7 +5,8 @@
     "node": true
   },
   "parserOptions": {
-    "ecmaVersion": 9
+    "ecmaVersion": 9,
```
bearfriend, line 8: Can we change this to `2018` for consistency? Or even upgrade it?

dlockhart: I'll try and remove it which will let it just inherit.

dlockhart: Done.

bearfriend (approved):

pureooze (approved):

web-dev-server-dev.config.mjs
```diff
@@ -1,13 +1,10 @@
 
-'use strict';
+import cors from '@koa/cors';
 
-const cors = require('@koa/cors');
-
-module.exports = {
-	compatibility: 'auto',
-	dedupe: true,
```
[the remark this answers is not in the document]
dlockhart: Right, I looked into this. I think their documentation is wrong, because it treats it as a `Boolean`, but if you look [at the plugin docs](https://github.com/rollup/plugins/tree/master/packages/node-resolve#dedupe) it's actually an array a package module names.

I tried passing in `true` and it exploded. The [old version](https://github.com/open-wc/es-dev-server#code-transformation) supports both a Boolean or an array... so maybe it used to be somehow magical? Reading through what this is actually supposed to do though, I don't think we ever needed this. We already have checks to make sure that there aren't duplicated nested libraries.

It also seems to let you dedupe `@namespace/package` and `package` but I can't think of places we do this either.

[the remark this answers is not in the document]
svanherk: Yeah I was trying to figure out what it was doing before too and got lost.  If it did ever work then it was doing something magical to create an array of everything for us. If we don't need it then that's great!

svanherk (approved):

web-dev-server-dev.config.mjs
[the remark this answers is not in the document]
dlockhart: Haha well clicking around I didn't notice any issues... but something to watch out for as I couldn't test all-the-things.

PR #4489, opened 2021-08-11
Add back deduping for local dev BSI runs

devpow112:
This re-adds deduping by getting a list of all package names and adding them to the list of dedupable packages. Not having this causes some complications with, at least my teams, local workflows. This does add time to the start-up of the dev server but it still was up and running before the build completed. I did this is vanilla nodejs as to not add new development dependencies.

```
[1] rollup v2.56.2
[1] bundles ./js/bsi.js → build\bsi.js...
[1] created build\bsi.js in 1.2s
[1] bundles ./sass/bsi.scss → build\bsi.css...
[1] (!) The emitted file "bsi.css" overwrites a previously emitted file of the same name.
[1] created build\bsi.css in 1.9s
[1] bundles ./sass/datagrid/datagrid.scss → build\datagrid.css...
[1] (!) The emitted file "datagrid.css" overwrites a previously emitted file of the same name.
[1] created build\datagrid.css in 181ms
[1] bundles ./sass/homepages/homepages.scss → build\homepages.css...
[0] Web Dev Server started...
[0]
[0]   Root dir: C:\D2L\brightspace-integration
[0]   Local:    http://localhost:8080/
[0]   Network:  http://10.151.124.224:8080/
[0]
[1] (!) The emitted file "homepages.css" overwrites a previously emitted file of the same name.
[1] created build\homepages.css in 222ms
```


dlockhart:
I'm trying to find the code in `es-dev-server` that must have been doing this previously... so far haven't.

devpow112:
> I'm trying to find the code in `es-dev-server` that must have been doing this previously... so far haven't.

Yeah I tried to find it as well and came up dry which is why I just went this route. If you find it I'd love to see what they did as I'm sure it's better than this (or at least I'd hope it was).

svanherk:
So seems like if you pass in `true` instead of an array of packages, `es-dev-server` will set `dedupe` to a function that decides on the fly if a package should be deduped: https://github.com/open-wc/es-dev-server/blob/98dd47464c97c5ab5fc6ef86723b3f529bc8265c/src/config.ts#L299-L300
Which `node-resolve` allows: https://github.com/rollup/rollup-plugin-node-resolve/pull/225/files
And it looks like the function is "if it's a package (doesn't start with `.` or `/`), dedupe it it" ?

dlockhart:
> So unless I'm misunderstanding... could we just do this?

Haha yep, sorry Devon and I were chatting in Slack and came to the same conclusion!

devpow112:
> > Haha yep, sorry Devon and I were chatting in Slack and came to the same conclusion!
> 
> Haha 🎉 . We used to dedupe in dev and prod, do we want it in both?

This doesn't seem to be needed for `prod` builds. I'll switch to the simpler method and add some comments around our finding for the next person that comes along going "what is this".

dlockhart (approved):
Thanks for finding/fixing!

PR #4516, opened 2021-08-13
polyfilling Object.entries and Object.values

dlockhart:
While doing a lot of old browser testing for Lit 2.0, I ran into problems with uses of `Object.entries` and `Object.values` -- even in our Lit 1.0 stuff.

This adds a dependency on `core-js`, which Babel already uses under the covers (the older version though), and then loads these two polyfills. With Lit 2.0, there'll be a few more required which is why I have this in a separate file.


package-lock.json
```diff
@@ -3759,6 +3759,12 @@
         "regenerator-runtime": "^0.11.0"
       },
       "dependencies": {
+        "core-js": {
+          "version": "2.6.12",
+          "resolved": "https://registry.npmjs.org/core-js/-/core-js-2.6.12.tgz",
+          "integrity": "sha512-Kb2wC0fvsWfQrgk8HU5lW6U/Lcs8+9aaYcy4ZFc6DDlo4nZ7n70dEgE5rtR0oG6ufKDUnrwfWL1mXR5ljDatrQ==",
+          "dev": true
+        },
```
dlockhart, line 3767: This ends up getting nested, which should be OK since this is a built-time thing. I tried using the older `2.x` version but it doesn't have `array.flat` which I'll need in Lit 2.0-land.

svanherk (approved):
Seems good!

PR #4711, opened 2021-09-14
Convert ifrau to an ESM module and surface it in BSI for the LMS to consume

grant-cleary:
We converted ifrau to an ESM module in 0.38.0 and we want to surface that to the LMS via BSI now.


grant-cleary:
I'm reasonably sure I'm expecting this to fail CI, but I would like to confirm that. ifrau is a transitive dependency for a few other dependencies within BSI.

grant-cleary:
Huh, this didn't fail CI... I think I must not understand how transitive dependencies are handled in BSI.

web-components/ifrau.js
```diff
@@ -0,0 +1,3 @@
+import { Host } from 'ifrau';
+
+window.ifrauhost = Host;
```
dlockhart, line 3: So this will need to be loaded using an `IBsiAsset` in the iFRA apploader code, but that's good because then it only gets included on that page. I guess `/web-components/` is no longer a great name for these things haha.

package-lock.json
```diff
@@ -7959,15 +7959,9 @@
...
-      "integrity": "sha512-Q/f8hubYmBc1Fhb7g4h6SOM1rXyboVVZJ2cs1HLLCn1f6EnHYN7KMX5J2hTaEL1qrC34p6Jad9aa9CU9CN7YcA==",
-      "requires": {
-        "iframe-resizer": "^3.5.0",
-        "inherits": "^2.0.1",
-        "lie": "^3.0.2",
-        "uuid": "^3.1.0"
-      }
+      "version": "0.38.0",
+      "resolved": "https://registry.npmjs.org/ifrau/-/ifrau-0.38.0.tgz",
+      "integrity": "sha512-s/JojuX4MpH/S4Mk/sscIWR9xzJH4o3XoSmw5KFFzt02qyj4/gyxOKE1uxJJ4Dl5rQq6OhZngr1s0ywT9iQNuA=="
```
dlockhart, line 7964: I'm a little surprised `frau-jwt` didn't get a nested version or something weird. Hmph.

dlockhart (approved):

web-components/ifrau.js
grant-cleary: Yeah, that's the workflow I figured we would want (and what I have mocked up in my local LMS). If there's a different way this should be done, that's fine too. The name caught me off guard a first, but the LMS has zanier naming than this. 😆

package-lock.json
grant-cleary: Yeah, I'm a bit nervous this will break something. I kind of assumed CI would catch anything that went wonky here... I guess if it fails on build I'll be ready to revert this quickly.

grant-cleary: I'm going to regenerate the package-lock to make sure I didn't miss anything. My connection isn't super stable so I can't be 100% sure this isn't broken. I imagine life is going to be terrible if I need to update frau-jwt, but... So be it.

grant-cleary: Yup, that's exactly what happened. Sigh. I guess my npm install crapped out and I didn't notice.

package-lock.json
```diff
@@ -12047,9 +12027,10 @@
       }
     },
     "uuid": {
-      "version": "3.4.0",
-      "resolved": "https://registry.npmjs.org/uuid/-/uuid-3.4.0.tgz",
-      "integrity": "sha512-HjSDRw6gZE5JMggctHBcjVak08+KEVhSIiDzFnT9S9aegmp85S/bReBVTb4QTFaRNptJ9kuYaNhnbNEOkbKb/A=="
+      "version": "3.3.2",
```
grant-cleary, line 12030: Hmm, how did this get downgraded? 😕

grant-cleary: Oh, I see. aws-sdk has a fixed version (3.3.2) that was previously nested as a dependency. The only other consumer of this was ifrau, which allowed anything ^3. So now that ifrau doesn't use this, we lock to the dependency aws-sdk is asking for instead.

dlockhart (approved):

PR #4724, opened 2021-09-15
adding colors to the main import

dlockhart:
Emily noticed that in certain quizzing pages our colour CSS custom properties weren't working.

The reason is that we don't explicitly include our colours as part of BSI's main import -- we're just relying on there being a web component somewhere on the page that imports them for us. While that's almost always true (the navbar ensures this), there are still some pages with no navbar and no other web components on them that need the colours.


bearfriend (approved):

dbatiste (approved):

danyakarras:
Thank you!! We'll be able to update so many hex values :D

PR #5076, opened 2021-11-02
Add `d2l-rubric-structure-editor` and `d2l-rubric-title` to build

devpow112:
These are the 4 top level components that rubrics advertises for use.


devpow112:
Will most likely conflict with #5077.

imcdonald-d2l (approved):

PR #5186, opened 2021-11-18
preparations to rename master branch to main

dlockhart:
We'd like to rename BSI's `master` branch to `main`, and these changes are needed before that can happen. There's also the [sync job in the LMS](https://github.com/Brightspace/lms/blob/master/.github/workflows/bsi-sync.yml), but since BSI communicates with it, it doesn't appear to require any updates.


.github/workflows/bsi-update.yml
```diff
@@ -30,6 +30,5 @@ jobs:
           AUTO_MERGE_TOKEN: ${{ secrets.BSI_AUTO_APPROVE_GITHUB_TOKEN }}
           BRANCH_NAME: ghworkflow/bsi_auto_update
           COMMIT_MESSAGE: 'BSI Auto Update Dependencies'
-          DEFAULT_BRANCH: master
```
dlockhart, line 33: `main` is the default value for this parameter

mdulat (approved):
look good!

dlockhart:
Interesting CI isn't going to run in Travis because of the branch name change haha

svanherk (approved):

svanherk:
> Interesting CI isn't going to run in Travis because of the branch name change haha

Probably a good thing so a release isn't triggered and all the other stuff runs and breaks?

PR #5262, opened 2021-11-29
US133990 - Add course-selector to BSI

jguann:
Following this doc: [https://desire2learn.atlassian.net/wiki/spaces/CUST/pages/2525790298/How+to+Create+a+Web+Component+LMS+Tool#Update-BSI-to-Serve-your-New-Tool](https://desire2learn.atlassian.net/wiki/spaces/CUST/pages/2525790298/How+to+Create+a+Web+Component+LMS+Tool#Update-BSI-to-Serve-your-New-Tool)


dlockhart:
@jguann you'll need to update the `package-lock.json` file as well whenever you add/update a dependency. It'll update automatically with a `npm install` -- just revert any additional updates to it other than this new dependency.

dlockhart:
@jguann looks like your new component is using Lit 2. Unfortunately we're not quite ready to make the jump over to the new Lit stuff, and we can't run multiple versions of Lit in parallel. So for the time being I think you'll need to downgrade to Lit 1 (`lit-element` 2 and `lit-html` 1).

package.json
```diff
@@ -54,6 +54,7 @@
     "d2l-consistent-evaluation": "BrightspaceHypermediaComponents/consistent-evaluation#semver:^1",
     "d2l-copy-assignment-to-other-courses": "Brightspace/copy-assignment-to-other-courses#semver:^1",
     "d2l-copy-role": "Brightspace/copy-role#semver:^1",
+    "d2l-course-selector": "BrightspaceUILabs/course-selector#semver:^1.0.3",
```
dlockhart, line 57: You can leave this as `semver:^1` since the `^` will automatically pick up any non-major version updates.

Once you're ready, you can also update course-selector to publish to NPM and at that point you could import it here using just `@brightspace-ui-labs/course-selector` (like flashcards or status-icon above). I can help with that if needed.

jguann: Got it. Thanks for your reviews!

rollup/rollup-wc.config.js
```diff
@@ -67,6 +67,7 @@ const componentFiles = [
 	'./node_modules/@brightspace-ui-labs/role-selector/role-selector.js',
 	'./node_modules/@brightspace-ui-labs/status-icon/status-icon.js',
 	'./node_modules/d2l-activities/components/d2l-subtitle/d2l-subtitle.js',
+	'./node_modules/d2l-course-selector/course-selector.js',
```
Syeda-Azim, line 70: Should be `./node_modules/@brightspace-ui-labs/course-selector/course-selector.js` as package name is `@brightspace-ui-labs/course-selector`
https://github.com/BrightspaceUILabs/course-selector/blob/main/package.json#L2

jguann: Oh, is this expected to be different from the path it's in for my local `node_modules`? My local has it in the path I specified.

dlockhart: I think you want the entry in BSI's `package.json` to match the name in its `package.json` -- that's where it installs things in `node_modules`.

So it should be:
```
"@brightspace-ui-labs/course-selector": "BrightspaceUILabs/course-selector#semver:^1",
```

dlockhart (approved):
LGTM!

Syeda-Azim (approved):

PR #5289, opened 2021-12-02
Bump user-profile-card to use latest version.

annabelsegalld2l:
I will also update consistent-eval here: https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/784.


svanherk:
@dlockhart This is the PR I mentioned in standup.  Would you have expected the CI script to fail because BSI and CE are using different user profile versions here?  Is it because it's an npm package? 😕

dlockhart:
So when I make this change locally, remove `package-lock.json` and `node_modules` and then re-install, I don't get these same changes -- I get a nested user-profile-card inside CE, as I would have expected. It's that nested entry that the `test-package-lock.mjs` script looks for and fails the build.

@annabelsegalld2l is it possible you reverted out the diff to CE part thinking it was a change you didn't want here?

svanherk:
>  I don't get these same changes -- I get a nested user-profile-card inside CE, as I would have expected

Interesting, in that case I would've expected `npm ci` to fail once it realized it needed a different version that wasn't called out explicitly for CE, but I guess it's not as smart as I thought it was

annabelsegalld2l:
> So when I make this change locally, remove `package-lock.json` and `node_modules` and then re-install, I don't get these same changes -- I get a nested user-profile-card inside CE, as I would have expected. It's that nested entry that the `test-package-lock.mjs` script looks for and fails the build.
> 
> @annabelsegalld2l is it possible you reverted out the diff to CE part thinking it was a change you didn't want here?

Yep, looks like I forgot to stage that change. The build should? fail now until consistent-eval's update is merged.

svanherk (approved):

PR #5323, opened 2021-12-08
Widen range explicitly

devpow112:
Currently was picking up latest in `package-lock.json` this just makes it more clear in the `package.json`.


dlockhart (approved):
👍

PR #5380, opened 2021-12-15
converting BSI to ESM-by-default

dlockhart:
While I had ESM on the brain converting all the visual-diff stuff, it seemed to make sense to convert the rest of BSI to be ESM-by default.

Most of this is the Oslo stuff, so @pureooze including you here.


package.json
```diff
@@ -105,7 +105,7 @@
     "@web/dev-server-esbuild": "^0.2",
     "autoprefixer": "^9",
     "chai": "^4",
-    "chalk": "^2",
+    "chalk": "^5",
```
dlockhart, line 108: Chalk 5 is ESM-only, and is the main motivator for pushing us to just use ESM now that it's available everywhere.

pureooze (approved):
Its amazing 🥲

PR #5382, opened 2021-12-15
upgrading dependencies

dlockhart:
Dependabot would get us these updates but it's broken with GitHub NPM references. Looks like it's been a while since we looked at these...


dbatiste (approved):
Wow yeah, lots of big upgrades.  :)

PR #5394, opened 2021-12-16
reverting rollup upgrade

dlockhart:
Something in here broke some of the static files HTML editor is using. Will investigate after this is fixed and narrow it down.


grant-cleary (approved):

grant-cleary:
Thanks @dlockhart! :)

PR #5435, opened 2021-12-22
Update `@brightspace-ui-labs/status-icon`

devpow112:
Noticed it was still picking up version `0.0.3`.


dlockhart (approved):
Cool -- looks like there weren't [actually any changes to the components](https://github.com/BrightspaceUILabs/status-icon/compare/v0.0.3...v0.1.0) so this seems safe.

devpow112:
> Cool -- looks like there weren't [actually any changes to the components](https://github.com/BrightspaceUILabs/status-icon/compare/v0.0.3...v0.1.0) so this seems safe.

Yeah it was just the work of switching to module by default.
https://github.com/BrightspaceUILabs/status-icon/pull/14

PR #5491, opened 2022-01-05
Us132915/adoption dashboard

johngwilkinson:
Testing: added LMS code to kook up, ran local BSI, and visited the new page


package-lock.json
```diff
@@ -4702,6 +4702,36 @@
...
+        "d2l-fetch": "github:Brightspace/d2l-fetch#semver:2",
+        "d2l-fetch-auth": "^1",
+        "export-from-json": "^1.4.0",
+        "highcharts": "^9.2.2",
+        "lit-element": "^2",
+        "lit-html": "^1.3.0",
+        "mobx": "^5.15.7"
+      },
+      "dependencies": {
+        "d2l-fetch": {
```
dlockhart, line 4729: The nested entry here is why CI is failing. Not totally sure why it doesn't like the `github:` bit, but you could try using: `"d2l-fetch": "Brightspace/d2l-fetch.git#semver:^2"` like the root BSI project does.

johngwilkinson: Thanks. I notice the assessment dashboard, which has basically the same dependencies, doesn't have this block at all, so I'll try removing it.

johngwilkinson: Good thought, but compare https://github.com/Brightspace/insights-assessment-quality-dashboard/blob/master/package.json#L62 - no problems there

I think/hope this is just something weird my local npm decided to do.

dlockhart: I just re-applied these changes locally and things worked, although you were missing a `^` in the lock file entry, which I've fixed.

bemailloux (approved):

PR #5523, opened 2022-01-10
removing release pin now that Rally is fixed

dlockhart:
Tara has fixed the release info in Rally so that they overlap again, allowing us to remove this temporary pinning of the release version.


svanherk (approved):
Phew, glad you had a reminder for this 😝 Even after joking this would be my problem next month, I did not do so 🤦

PR #5576, opened 2022-01-18
Update `d2l-rubric` to `4.x`

devpow112:
This is the lit version of rubrics with the LD flag for switching back to the polymer version if needed.


package.json
```diff
@@ -79,7 +79,8 @@
     "d2l-program-outcomes-picker": "Brightspace/program-outcomes-picker#semver:^3",
     "d2l-quiz-components": "Brightspace/quiz-components#semver:^1",
     "d2l-resize-aware": "BrightspaceUI/resize-aware.git#semver:^1",
-    "d2l-rubric": "Brightspace/d2l-rubric#semver:^3",
+    "d2l-rubric": "github:Brightspace/d2l-rubric#semver:^4",
+    "d2l-rubric-polymer": "github:Brightspace/d2l-rubric#semver:^3",
```
devpow112, line 83: This is needed for now since we are feature flagging stuff.

package.json
```diff
@@ -59,6 +58,7 @@
     "d2l-copy-role": "Brightspace/copy-role#semver:^1",
     "d2l-cpd": "Brightspace/continuous-professional-development#semver:^3",
     "d2l-custom-early-progress-report": "Brightspace/custom-early-progress-report#semver:^1",
+    "d2l-custom-my-courses-in-other-orgs": "github:Brightspace/custom-my-courses-in-other-orgs.git#semver:^1",
```
devpow112, line 61: Just some npm sorting when running npm install/update

imcdonald-d2l: Looks like `github:` was appended with the sort as well. Is that intented?

devpow112: Yeah it was done automatically by npm. It seems to want `github:` as a prefix. It helps it determine which git endpoint to use.

imcdonald-d2l (approved):

PR #5580, opened 2022-01-18
US134773: Add practice question prototype into BSI

ianmclean2011:


package-lock.json
```diff
@@ -5329,6 +5339,33 @@
         "siren-parser": "^8.0.0"
       }
     },
+    "d2l-practice-questions": {
+      "version": "github:Brightspace/team11_pq_prototype#77ffabcb3b875b18e09dae11461d917c1ef11291",
+      "from": "github:Brightspace/team11_pq_prototype#semver:^0",
+      "requires": {
+        "@brightspace-ui/core": "^1",
+        "lit-element": "^3"
```
dlockhart, line 5347: Unfortunately we're not ready to go to Lit 2 just yet (see Annabel's blog post from today), so for now this will need to downgrade to `lit-element` version 2.

mdgbayly (dismissed):

package-lock.json
mdgbayly: oops - good eye - did not notice that.

ianmclean2011: @dlockhart ah ok ... I think I received a dependabot PR update about that and just sort of blindly accepted it :D ... I'll update that and update the PR ... thanks!

PR #5582, opened 2022-01-18
Update `concurrently` dependency

devpow112:
New version works for our use case without changes
https://github.com/open-cli-tools/concurrently/releases/tag/v7.0.0


dlockhart (approved):
Nice, thanks.

PR #5587, opened 2022-01-19
Remove htmleditor static files from BSI build

grant-cleary:
Now that BSI has updated to use v1.40.0 of the html-editor component, we no longer need these static files as part of the build (the editor doesn't need them anymore). 🎉


dlockhart (approved):
🎉

PR #5630, opened 2022-01-26
Set npm range limit until we can fix the build at version 7

svanherk:
Now when someone runs `npm i` with npm 7 or greater, it will fail.  This is to avoid them suffering through a 1.5 hour BSI build, until we can fix the problems we're seeing with npm 7.

I went with an `npm` range instead of `node` (since the real problem is `npm`), but let me know if you think we should force them on Node 14.


devpow112 (approved):

dlockhart (approved):
Thanks Stacey.

annabelsegalld2l (approved):
I tested with Node version 17 and 16, both of which fail on `npm install` as expected. Node 14 installs correctly. Just a note that it will take a few minutes to actually fail for the unaccepted versions, it's not instantaneous.

PR #5640, opened 2022-01-27
Update dirname to use fileURLToPath in oslo

kieranderson:
To work with windows


oslo/generate-monolith-xml.js
```diff
@@ -3,8 +3,9 @@ import { Lang } from './Lang.js';
 import { LangPackage } from './LangPackage.js';
 import path from 'path';
 import { Serge } from './Serge.js';
+import { fileURLToPath } from 'url';
 
-const dirname = path.dirname(new URL(import.meta.url).pathname);
```
pjdon, line 7: what was the issue with the old method?

svanherk:
We ran into this in `core` as well: https://github.com/BrightspaceUI/core/pull/2004
Does https://github.com/Brightspace/brightspace-integration/blob/main/oslo/generate-serge-mapping.js#L5 need to be updated as well, or is that never run locally?

pureooze (approved):

oslo/generate-monolith-xml.js
pureooze:
```javascript
import { dirname, join, resolve } from "path";
import path from 'path';
import { fileURLToPath, pathToFileURL } from "url";

console.log(path.dirname(new URL(import.meta.url).pathname))
console.log(dirname(fileURLToPath(import.meta.url)))

$ node app.js
/C:/D2L/offstack/teamusa-scratch/code-search
C:\D2L\offstack\teamusa-scratch\code-search
```

oslo/generate-monolith-xml.js
```diff
@@ -3,8 +3,9 @@ import { Lang } from './Lang.js';
 import { LangPackage } from './LangPackage.js';
 import path from 'path';
```
pureooze, line 4: nit: could make this specific, from what I can tell this file only uses `dirname` and `join`

dlockhart (approved):
Thanks for fixing this up. We had a [similar issue in core](https://github.com/BrightspaceUI/core/pull/2004) and fixed it the same way.

dlockhart:
BSI just switched to use GitHub Actions, so if you want CI to pass you'll need to rebase from `main`.

kieranderson:
> We ran into this in `core` as well: [BrightspaceUI/core#2004](https://github.com/BrightspaceUI/core/pull/2004) Does https://github.com/Brightspace/brightspace-integration/blob/main/oslo/generate-serge-mapping.js#L5 need to be updated as well, or is that never run locally?

I run it very rarely on my local, but I imagine most people aren't doing that. I've updated it though, thanks for pointing it out.

PR #5642, opened 2022-01-28
Replacing Travis CI with GHAs

dlockhart:
First attempt at replacing Travis CI with GitHub Actions. Actually it's like attempt number 39282 but those all happened in my fork lol.

We already had a `CI` action in here that we were using just to monitor and see if it was flaky. I'd forgotten about it, so I guess it was working. I've updated it to run on any PR and added linting and unit tests.

The main change here is obviously the "Release" action. I'd originally split this into two steps similar to before -- a "tag" and then "deploy", but because you can't trigger an action from a release that was created automatically (without using separate tokens), it seemed easier (and quicker) to just put everything in one action. This matches how other repos work anyway.


.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          persist-credentials: false
+      - uses: Brightspace/third-party-actions@actions/setup-node
+        with:
+          node-version: '14'
+          cache: 'npm'
+      - name: Install dependencies
+        run: |
+          npm ci
+          npm install frau-publisher --no-save
```
dlockhart, line 24: This allowed me to move it out of BSI's dev dependencies.

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Get Release
+        id: get-release
+        run: |
+          if [ "${{ steps.get-branch.outputs.branch }}" == "main" ]; then
+            echo "::set-output name=release::$(echo ${{ steps.get-lms-version.outputs.LMS_VERSION }})"
+          else
+            echo "::set-output name=release::$(echo ${{ steps.get-branch.outputs.branch }})"
+          fi
+        shell: bash
+      - name: Increment Release
```
dlockhart, line 49: I've kept the release version scheme the same so as not to introduce too many crazy changes, but in theory I think we could switch this to use Scy's automatic release incrementing and branch creation stuff too. But that's a larger change.

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+                if (version > maxVersion) {
+                  maxVersion = version;
+                }
+              }
+            }
+
+            console.log(`Maximum existing build for release "${release}" is "${maxVersion}".`);
+            return `v${release}-${++maxVersion}`;
+      - name: Create Release/Tag
+        uses: actions/github-script@v5
```
dlockhart, line 81: These `github-script` actions are sweet. So much easier than bash or separate JS files.

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag_name: ${{ steps.increment-release.outputs.result }},
+              name: ${{ steps.increment-release.outputs.result }},
+              target_commitish: context.sha
+            });
+      - name: Publish to CDN
+        run: |
+          echo "Publishing to CDN as version: ${{ steps.increment-release.outputs.result }}"
+          # npx frau-publisher -m lib -t bsi -f \"./build/**/*\" -v ${{ steps.increment-release.outputs.result }}
```
dlockhart, line 95: This is commented out right now as we'll need to switch the token stuff over from Travis to GHA. The only thing I can't test is whether this will publish into `lib/bsi/v20.22.2-123` (with the `v`) or without it like we want.

.github/workflows/ci.yml
```diff
@@ -1,11 +1,8 @@
...
-  push:
-    branches:
-      - main
+on: pull_request
 jobs:
-  build:
-    name: Build
+  test:
+    name: Test
     timeout-minutes: 15
```
dlockhart, line 6: Will adjust timeout once I know roughly how long this takes.

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Brightspace/third-party-actions@actions/checkout
+        with:
+          persist-credentials: false
+      - uses: Brightspace/third-party-actions@actions/setup-node
+        with:
+          node-version: '14'
+          cache: 'npm'
```
omsmith, line 20: Could consider [node-version-file](https://github.com/Brightspace/third-party-actions/blob/actions/setup-node/docs/advanced-usage.md#node-version-file) to not have extra places to forget updating.

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+          npm install frau-publisher --no-save
+      - name: Build
+        run: |
+          npm run build
+          npm run build:lang
+        env:
+          NODE_OPTIONS: '--max-old-space-size=6656'
+      - name: Get Branch Name
+        id: get-branch
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
```
omsmith, line 33: This is a different syntax than I'm used to. Is it written down somewhere?
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

Also, the subshell looks unnecessary.

```suggestion
        run: echo "::set-output name=branch::${GITHUB_REF#refs/heads/}"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+          RALLY_API_KEY: ${{ secrets.RALLY_API_KEY }}
+      - name: Get Release
+        id: get-release
+        run: |
+          if [ "${{ steps.get-branch.outputs.branch }}" == "main" ]; then
+            echo "::set-output name=release::$(echo ${{ steps.get-lms-version.outputs.LMS_VERSION }})"
+          else
+            echo "::set-output name=release::$(echo ${{ steps.get-branch.outputs.branch }})"
+          fi
+        shell: bash
```
omsmith, line 48: Prefer passing through environment variables rather than substituting directly into the script to avoid injection possibilities.

```suggestion
      - name: Get Release
        id: get-release
        run: |
          if [ "${BRANCH}" == "main" ]; then
            echo "::set-output name=release::${LMS_VERSION}"
          else
            echo "::set-output name=release::${BRANCH}"
          fi
        shell: bash
        env:
          BRANCH: ${{ steps.get-branch.outputs.branch }}
          LMS_VERSION: ${{ steps.get-lms-version.outputs.LMS_VERSION }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+              if (match) {
+                const version = parseInt(match[1]);
+                if (version > maxVersion) {
+                  maxVersion = version;
+                }
+              }
+            }
+
+            console.log(`Maximum existing build for release "${release}" is "${maxVersion}".`);
+            return `v${release}-${++maxVersion}`;
```
omsmith, line 79:
```suggestion
        id: increment-release
        uses: actions/github-script@v5
        with:
          script: |
            const options = github.rest.repos.listReleases.endpoint.merge({
              owner: context.repo.owner,
              repo: context.repo.repo,
              per_page: 100
            });
            const releases = await github.paginate(
              options,
              response => response.data.map(release => release.tag_name)
            );

            const release = process.env.RELEASE;
            const releaseRe = new RegExp(`^v${release.replace('.', '\\.')}-([0-9]+)$`);

            let maxVersion = 0;
            for (let i = 0; i < releases.length; i++) {
              const match = releaseRe.exec(releases[i]);
              if (match) {
                const version = parseInt(match[1]);
                if (version > maxVersion) {
                  maxVersion = version;
                }
              }
            }

            console.log(`Maximum existing build for release "${release}" is "${maxVersion}".`);
            return `v${release}-${++maxVersion}`;
        env:
          RELEASE: ${{ steps.get-release.outputs.release }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+          script: |
+            const options = github.rest.repos.listReleases.endpoint.merge({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              per_page: 100
+            });
+            const releases = await github.paginate(
+              options,
+              response => response.data.map(release => release.tag_name)
+            );
```
omsmith, line 62: This is pretty long and getting longer right?

> To stop paginating early, you can call the done() function passed as 2nd argument to the response map function. Note that you still have to return the value you want to map the response to, otherwise the last response will be mapped to undefined.

Are we able to assume earlier is newer / larger, just take the first one that matches our release, and stop paginating, or something like that?

.github/workflows/release.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          script: |
+            console.log(`Creating release ${{ steps.increment-release.outputs.result }} at SHA "${context.sha}"...`);
+            await github.rest.repos.createRelease({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              tag_name: ${{ steps.increment-release.outputs.result }},
+              name: ${{ steps.increment-release.outputs.result }},
+              target_commitish: context.sha
+            });
```
omsmith, line 91:
```suggestion
      - name: Create Release/Tag
        uses: actions/github-script@v5
        with:
          script: |
            console.log(`Creating release ${process.env.RELEASE_NAME} at SHA "${context.sha}"...`);
            await github.rest.repos.createRelease({
              owner: context.repo.owner,
              repo: context.repo.repo,
              tag_name: process.env.RELEASE_NAME,
              name: process.env.RELEASE_NAME,
              target_commitish: context.sha
            });
        env:
          RELEASE_NAME: ${{ steps.increment-release.outputs.result }}
```

.github/workflows/release.yml
svanherk: Yeah, now that we can have actions inside of actions, we'd probably be able to use this instead separate js scripts in our `actions` repo

svanherk: I think we should be using the version from https://github.com/Brightspace/third-party-actions though?

.github/workflows/release.yml
dlockhart: I got it from the accepted answer here:
https://stackoverflow.com/questions/58033366/how-to-get-the-current-branch-within-github-actions

.github/workflows/release.yml
dlockhart: Yep good call. I've updated it but our copy is 2 major versions back so the API has changed and stuff is breaking. Going to chat with @j3parker to see if we can get that updated or if staying back on v3 is intentional.

.github/workflows/bsi-update.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Use Node.js
         uses: Brightspace/third-party-actions@actions/setup-node
         with:
-          node-version: '14.x'
+          node-version-file: .nvmrc
 
       - name: Update package-lock.json
         uses: BrightspaceUI/actions/update-package-lock@main
```
svanherk, line 27: So I think the below will need to change to:
```
          APPROVAL_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          AUTO_MERGE_TOKEN: ${{ secrets.BSI_AUTO_APPROVE_GITHUB_TOKEN }}
          GITHUB_TOKEN: ${{ secrets.BSI_AUTO_APPROVE_GITHUB_TOKEN }}
```
Because `secrets.GITHUB_TOKEN` does not trigger additional workflows.  So we want to use our special token to open the PR (triggering `ci./yml`), the regular token to approve it (since the same token can't both open and approve a PR), and then the special token to merge it (triggering `release.yml`).

README.md
```diff
@@ -128,7 +127,7 @@ To skip automatic tagging, include the text `[skip release]` in your merge commi
 
 ### Publishing to the CDN
 
-For each tag, all BSI assets (contents of the `build` directory) will be published to the Brightspace CDN by its [Travis CI job](https://travis-ci.com/github/Brightspace/brightspace-integration).
+For each tag, all BSI assets (contents of the `build` directory) will be published to the Brightspace CDN.
```
svanherk, line 130: This feels a little inaccurate... if I just create a random tag, that will no longer trigger a publish.  Maybe something like this?

```suggestion
For each merge to `main` or a release branch, a tag is created and all BSI assets (contents of the `build` directory) will be published to the Brightspace CDN.
```

svanherk: Or maybe this section and the section above should just be combined together and rewritten.  Basically, tagging is a thing we do now to help track/map commit to publish location (and for now, trigger the lms sync job).  But it's no longer a trigger itself

.github/workflows/release.yml
dlockhart: Yeah, agreed. I've updated this to stop fetching new releases once it sees the first match. The only time it'll be slow will be the first release of each month.

README.md
dlockhart: Agreed -- thanks!

.github/workflows/release.yml
dlockhart: I've tested this and the `v` indeed gets stripped out by `frau-publisher`. This is now uncommented and ready to go.

.github/workflows/release.yml
```diff
@@ -0,0 +1,104 @@
...
+              tag_name: process.env.RELEASE_NAME,
+              name: process.env.RELEASE_NAME,
+              target_commitish: context.sha
+            });
+        env:
+          RELEASE_NAME: ${{ steps.increment-release.outputs.result }}
+      - name: Publish to CDN
+        run: |
+          echo "Publishing to CDN as version: $RELEASE_NAME"
+          npx frau-publisher -m lib -t bsi-test -f \"./build/**/*\" -v $RELEASE_NAME
```
svanherk, line 99: So is the plan to:
- Turn off the Travis cred creation and setup the GitHub Action creation instead
- Merge this PR and see that this publishes as expected?
- Update this line to point to the real location?

I think the BSI sync job will trigger from the tag creation, but then the PR in the lms will fail because it won't be able to find the files.  But maybe that's fine?

dlockhart: Sorry no that `bsi-test` was a mistake left over from my testing in my fork -- I've updated it. But yeah... I think those steps minus the last one should work.

.github/workflows/release.yml
```diff
@@ -0,0 +1,104 @@
...
+                    break;
+                  }
+                }
+            });
+
+            console.log(`Maximum existing build for release "${release}" is "${maxVersion}".`);
+            return `v${release}-${++maxVersion}`;
+        env:
+          RELEASE: ${{ steps.get-release.outputs.release }}
+      - name: Create Release/Tag
```
svanherk, line 82: Thinking about it more, I think we also need our special token to create this release to trigger the `trigger-lms-sync.yml` workflow.  I assume we can pass `github-script` a different token?

dlockhart: Right, that's the same issue I noticed when I tried to initially split the "create release" task from the "deploy to CDN" task. Updated to use `BSI_AUTO_APPROVE_GITHUB_TOKEN`, which is a little misleading but it's already used in `bsi-update.yml` in a misleading way, so 🤷 ?

svanherk: Yeah, we could probably rename it lol.  And really, couldn't we just use the org-wide `D2L_GITHUB_TOKEN`?  Does this `brightspace-bot` token have special permissions?

dlockhart: It just has whatever permissions we give `brightspace-bot` in this repo... which is currently `Write`.

svanherk (approved):

PR #5660, opened 2022-02-01
fixing lms-sync workflow trigger

dlockhart:
The release step was still using the built-in GitHub token, which was causing the `trigger-lms-sync` workflow to NOT be triggered. This should have it use the `brightspace-bot` account to create the release.


.github/workflows/release.yml
```diff
@@ -12,8 +12,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: Brightspace/third-party-actions@actions/checkout
-        with:
-          persist-credentials: false
```
dlockhart, line 16: Don't think this was ever required.

.github/workflows/release.yml
```diff
@@ -82,6 +80,7 @@ jobs:
       - name: Create Release/Tag
         uses: Brightspace/third-party-actions@actions/github-script
         with:
+          github-token: ${{ secrets.D2L_GITHUB_TOKEN }}
```
dlockhart, line 83: From: https://github.com/actions/github-script#using-a-separate-github-token

svanherk (approved):

PR #5661, opened 2022-02-01
Send ci and release failures to slack

svanherk:
If the previous job fails, we send a message to a webhook workflow in Slack that I'll setup in the `build-triage-bsi` channel.  Will looks like similar to this (the first one is a slightly older version):
image
image


dlockhart (approved):
Looks amazing!

PR #5665, opened 2022-02-01
Backport GitHub Actions to 20.22.2

dlockhart:
I regenerated the `package-lock.json` file and then hand-staged only the changes (mostly removals) that appeared in the `main` pull request.

This seems pretty safe, but if we're worried about picking up something we don't want, we could always NOT remove the dependencies (e.g. Rally, frau-publisher, etc.) in these branches.


svanherk (approved):
Up to you!  Leaving the dependencies for the backports sounds fine to me, but if you feel confident with the cherry-picking I'm good with that too.  None of the removals in `package-lock.json` stand out to me as particularly worrisome

dlockhart:
Ok I'll leave it in this one... but it took like 45 minutes to hand cherry-pick them so if this doesn't apply nicely to `20.22.1` then I might opt to leave them in that branch.

PR #5677, opened 2022-02-03
Use native focus-visible for menu-item component

grant-cleary:
Contains these changes: https://github.com/BrightspaceUI/core/compare/v1.215.0...v1.215.1


package.json
```diff
@@ -37,7 +37,7 @@
     "@brightspace-ui-labs/status-icon": "^0.0.3",
     "@brightspace-ui-labs/user-feedback": "^1",
     "@brightspace-ui-labs/user-profile-card": "^1",
-    "@brightspace-ui/core": "^1.114.0",
+    "@brightspace-ui/core": "1.215.1",
```
dbatiste, line 40: Did you intend to peg to this version?

dbatiste: I see, this is for you hotfix.

dbatiste (approved):

package.json
grant-cleary: That's right. Up until a very short moment ago, there was an additional change in 1.215.2 that I didn't want to come along for the ride. That has now been reverted so in theory I could just switch this to ~1.215, but I don't really have a preference either way.

PR #5683, opened 2022-02-04
DE47116 bump version of brightspaceui-core

tylergee:
For input-number component bug
https://rally1.rallydev.com/#/?detail=/defect/625004712675&fdp=true


package.json
```diff
@@ -37,7 +37,7 @@
     "@brightspace-ui-labs/status-icon": "^0.1",
     "@brightspace-ui-labs/user-feedback": "^1",
     "@brightspace-ui-labs/user-profile-card": "^1",
-    "@brightspace-ui/core": "^1",
+    "@brightspace-ui/core": "1.223.4",
```
dlockhart, line 40: Technically this change here isn't required -- just the lock file change -- but it doesn't hurt anything.

dlockhart (approved):
Here's the full diff:
https://github.com/BrightspaceUI/core/compare/v1.223.1...v1.223.4

@grant-cleary FYI as I know you were waiting for this change too.

package.json
tylergee: Good to know, thanks

PR #5753, opened 2022-02-14
fixing publish step to not pull everything new

dlockhart:
I did the `main` version first to see which dependencies this added, since I need to somehow (likely manually) cherry-pick these and nothing else back into 20.22.3 and 20.22.2.


devpow112 (approved):

svanherk (approved):

svanherk:
Or - can we `npm install frau-publisher --no-save` after the build is done? Not sure if we want that to be the final solution, but might be safer for the backports than manually cherry-picking all these dependencies?

dlockhart:
I like that better. Lol I just spent 15 minutes manually backporting but it's still safer. Will update.

dlockhart:
Ok done -- that'll be much easier to apply to the other branches.

